PR #231, opened 2014-10-30
handling closed_content issues with view api

ethn:
I believe this is now ready to merge.


lib/card/format.rb
```diff
@@ -295,11 +296,16 @@ def process_content_object content=nil, opts={}
...
+          :too_deep
+        when @@perms[view] == :none                     # permission skipping specified in view definition
+          view
+        when args.delete(:skip_permissions)             # permission skipping specified in args
+          view
+        when !card.known? && !tagged(view, :unknown_ok) # handle unknown cards (where view not exempt)
+          view_for_unknown view, args  
+        else                                            # run explicit permission checks
+          permitted_view view, args    
         end
```
ethn: Do you like this better, Gerry?

(you can probably track down your previous comment via the Communication tab)

GerryG: Yes, that's what I had in mind, only, why the extra space before the comments?

PR #263, opened 2014-12-01
Card gem or the Card/Deck Split

GerryG:
This seems to be proceeding past the discussion stage towards getting it ready to merge.

I've looked a little at how rails does the many projects in one repo and tried to do something similar here.  My thinking now is that if we get this split right, then everything needed for the platform is in the card gem and the wagn (soon to be decko) gem is what makes it case 1, a web application that uses the card engine.  Update: Ethan suggested he wanted to the card gem to be just the model part as far as possible (a little view stuff leaks in), so that means we now have three gems.  The original wagn-gem now mostly has the wagn command and rake tasks, plus some assets that probably shouldn't be there.  Then there is the card-gem, which gets referred to in configuration as Cardio.  It might be logical the call this gem cardio and include the Card class in it and avoid ':require=>false' needed so it doesn't load the Card class before AR is all loaded and connected.  Now we have added a decko-rails gem with the rails glue.  If we moved the rest of the Application stuff from wagn to decko, we would be close to having no run-time dependencies on the wagn gem (remaining assets without a good home excepted).

Question for all is how we want to connect to an external user model, do we always want to have a card with a card_id that is associated with the external model?  Is that just another instance of an external model bound to a deck via a standard way of doing that?  My plan was to add card_id to the external table (making it belong to a card and the card will have_one of the external model).

The part I haven't figured out/finalized is the way the top level objects should work.  Initially I thought (wrongly) that the Engine (Card::Engine < Rails::Engine) needed to be part of the Wagn::Application < Rails::Application object, but there just isn't a way to have one object fill both roles.  And this split implies that the application will always have very simple routes that mount the Card::Engine thing, and the same thing in the platform use, but you'd have all your application's initial routes.  We may also want to split out some of the legacy routes to the application and leave only the platform standard stuff in Card::Engine routes.

Update: I still don't think this stuff is the best it can be, but it works.  I think it should be possible for Wagn::Application, Decko::Engine and maybe Cardio::Railtie to share one config (and paths) object, but I gave up on it for now.  Now, there is glue stuff in Decko::Engine that pushes it's config object in to a Cardio class method where we can get it as Cardio.config (and will be Card.config too when I add that).


card/Gemfile
```diff
@@ -0,0 +1,6 @@
+source 'http://rubygems.org'
+
+gemspec
+
+#gem 'card', :path=>'.'
```
GerryG: I'm not sure that I meant to check this in, but I do wan't some things to work from inside the gem dir (like rake?).  Maybe this should be only one or both of the other two.

Opinions on that?

card/lib/card.rb
```diff
@@ -36,6 +36,7 @@ class Card < ActiveRecord::Base
   after_save :extend
   
   TRACKED_FIELDS = %w(name type_id db_content trash)
-  Loader.load_mods if count > 0
 
+  ActiveSupport.run_load_hooks(:card, self)
```
GerryG: I think this is pretty important to the load/configure api.  We hook onto this in Cardio to load mods, but devs can also if need be.  This would make it easy to have several of those for loading mods in different places.  That might argue for configuring mods differently.  Maybe we don't need paths entries and just have the load call take an argument with the mod location.

card/Gemfile
ethn: yes, we have a policy there: no Gemfiles in the repo

ethn: in general our approach with rake is increasingly that rake tasks should be performed from a dev deck

card/lib/card/location.rb
```diff
@@ -0,0 +1,45 @@
...
+  # Links ----------------------------------------------------------------------
+
+  def link_to_page( text, title=nil, options={})
+    title ||= text
+    url_options = {}
+    [:type, :view].each { |k| url_options[k] = options.delete(k) if options[k] }
+    url = page_path( title, url_options )
+    link_to text, url, options
+  end
+
```
GerryG: So, you are saying that this can just become a module in set/all/location.rb and get included in all cards?  Which mod, probably core ?

I should probably hold off in adding more changes to this branch until we finish the important changes and merge it, right?

card/lib/card/location.rb
```diff
@@ -0,0 +1,45 @@
...
+end
```
GerryG: It all seems to be working a lot better now.  The module Cardio was pretty important.

card/mod/01_core/set/all/permissions.rb
```diff
@@ -72,7 +72,7 @@ def deny_because why
 
 def permitted? action
 
-  if !Wagn.config.read_only
+  if !Cardio.config.read_only
```
ethn: I still think, API-wise, it would be nice if we could access this as `Card.config.read_only`.  I know there are some loading contexts where this won't work and Cardio is needed.  But can't we just add this to the card class somewhere:

````
def config
  Cardio.config
end
````
...and use that in mod contexts?

card/lib/card/location.rb
ethn: I would love to have location wrapped into the mod system.

It used to be included in controllers and mailers if not elsewhere.  I think you managed to remove it from mailers without problem.  And you separated out the controller stuff a bit, too, right?  (Can't see all that here). If so, I'd guess there are some calls like wagn_url that would need to be tweaked.

Right, I'd definitely like to be thinking MVP (minimal viable product) here so that we can get this stuff merged in.  I just don't know whether the location stuff is currently fully functional.

card/mod/01_core/set/all/permissions.rb
GerryG: Sure, that should be fine.  If there are places where Card.config ... doesn't work (and Card.paths and Card.root) then Cardio.x will still work.

card/lib/card/location.rb
GerryG: You understand that some of the stuff in Wagn::Location moved to Decko::Location, right?  Essentially the session stuff because that is bound to the controller environment.  I just explicitly require 'decko/location' in mods that use sessions.  Here we are only talking about the Card::Location methods.  We should discuss if you think we need this to be done differently before we can merge.  I think it is ok for some modules to have bindings to decko.  We would have to adapt those parts for a different web-context, but it probably isn't that hard to generalize this important touch point.  Could be as simple as being able to configure a replacement for Decko::Location via Cardio.

card/Gemfile
GerryG: I guess that is typical for gems, they can add tasks, but you don't usually run anything from the gem dir.

card/lib/card/location.rb
ethn: yeah, I mentioned it, but I couldn't see Decko::Location here (diff is too big) so wasn't sure how you were pulling off wagn_redirect.  I just pulled up the code in your branch and It looks like the controller is still including Card::Location, so we're not ready to go to mods.

Now that I've looked there, I'm a little surprised to see that wagn is depending upon decko-rails, but I think that's ok for now.

GerryG: I think the controller could reference the functions as Card.method.  If I'm not mistaken we are talking all class methods and Card will be fully loaded in the controller.

GerryG: Isn't it that the modules have to save locations and maybe clear them, where the controller needs it for redirects?

ethn: the history stuff is effectively a separate system; I doubt that separation will cause much problem.

the main overlap in the handling in controllers/formats is in all the wagn_url and wagn_path stuff.  the controller needs it for redirection/success stuff, and the formats need it for urls in content.

Yes, you could refer to the mods version from the controller, but from unless you're thinking of a bigger change than I know, it would have to be @card.format.method (not Card.method).

The nice thing about the mods approach is that you can then customize the handling of locations on a set basis. Can't say that's come up before, but it could prove useful at some point.

GerryG: Just looked a little at the methods in Card::Location.  Yes, they do reference the web context, but only through attributes of the config object and the host/port via Card::Env.  That config attribute is one that is set up by the Rails::Application setup, but we just have to match it for other contexts.  Note that these have been mostly renamed from wagn_ prefix to card_ prefix.

ethn: Yeah, I wasn't ever concerned about that.

GerryG: And I missed something ... there is one method used from the view context, link_to in the last method (link_to_path ?), and that is only used in a couple of places.  The others can all be class methods.

Now I'm not sure about why split off the session (history) methods, the session is just one more item in the view context.

Maybe they all go in Decko::Location (well not the 'class method' like ones) ?  I'll go with your direction on this.

ethn: did you not like the idea of making them instance methods in the mod?  that way they could be overridden by set (not true of class methods).

This does mean they'd need to be called from a card instance in the controller, but that's not really a problem, because the controller always has one.

GerryG:
I didn't want to add to an already huge PR, but this is some key stuff that we'll want soon after this is ready: https://github.com/GerryG/wagn/pull/41
This one is between the two branches, and is for reference  now.  If we want to combine them for the merging at some point we can just merge this second branch into card-gem.

ethn:
cool.  thanks for doing it that way.  Sorry I'm a bit backed up on PR's.

GerryG:
I wanted to ask you about the semaphore configuration.  It looks like it isn't using 'wagn rspec', which I might be more dependent on now.  Or, pointing to the gem directory is a little funny now.  So where do I see that config and can I change it?

GerryG:
You can now see this other branch (mostly) building on my semaphore config: https://semaphoreapp.com/gerrygleason/wagn/branches/generator

I got the build to work with the wagn command for rspec and cucumber.  It was a bit odd, the output wasn't showing up unless I added 2>&1 to the system command that ruby was running from command.rb.  It also seemed to loose the exit status, so I added some code for that too.  I could add a separate pull request for that if you want.

ethn:
I tried playing with this on my flight yesterday but couldn't get the data seeding (wagn seed -t) to work.  There was an issue with Card::Version.  Any chance you could have a look at that?

(Let me know if you have trouble reproducing it; I suspect it will show up when you run it and that it's just a loading situation not emulated in our tests)

GerryG:
That was working.  It had to be working before I tried to merge latest master because semaphore passed and it uses it in setup.

GerryG:
This commit: https://github.com/GerryG/wagn/commit/5a597345f4e16c40fae118a0ed0ab3030b227290

ethn:
Now that we have semaphore working on master again (there was an issue with the 1.9.8.0 version of the phantomjs gem), we can get 1.14.7 out the door, and then I'd like to get the card-gem stuff merged into master and start putting out 1.15 pre-releases.  (and we can stop having to do all these difficult merges; I'm really grateful to you for doing so many already!)

I know you had it working before and it got broken in the updates, but would you be willing to have another go at getting it to pass semaphore?

By the way, was it you who changed semaphore to start using the -c on wagn new?  I like that a lot.  We can probably delete the other gemfiles there now.

One other side note (not really related to this PR): I think rather than the gem_path arg  for wagn new, we should support environmental variables in core-dev decks. That way we can put WAGN_DEV_GEM_PATH in our bash profiles (and in the ENV variables on semaphore) and not give it another thought ;)

GerryG:
I didn't think I had the power to change the build on the merge events.  I have my own account and you can note some other changes there too (in particular, I use wagn to run tests, not rspec or cucumber commands directly).  I managed to get my branches to start working on the main build, but it may not run all the tests (not sure).

Note,  you may want to consider getting this branch and the generator branch in together.  Well, your call.  There is a bit more cleanup work on that branch, and I think I could use your help specing just what I need to do to finish.  I think all the current use cases are still functioning as expected, but I'm trying to support some of the new ones in the generators now.

GerryG:
I'm pretty close on the merge now.  And I think the changes you want for setup, etc. are great additions to the generator branch.  Comment on that PR and tell me what you like me to attempt to make work.

GerryG:
seed working now, but cucs still not.  Shouldn't be too bad.

GerryG:
Looks like I got the merge problems fixed.  All that moving around of files made it pretty difficult.  Missed changes in base.rb when it moved to basic_types.

xithan:
Thanks for figuring out how to use the wagn command on semaphore. I had been trying that when I set up semaphore. I added 2>&1 and it seems to work now.

GerryG:
I asked the semaphore guy who sent an email greeting and he never answered about what is happening to stderr in our wagn process or the process it spawns with system.  Still a mystery.  The only response what "I can echo to stderr and I see it", so who knows but we have a workaround.

ethn:
Does this work in the browser for you?  For me it immediate fails with a routing error
````
undefined method `cache' for #<Rails::Application::Configuration:0x007fc6fafb5518>
````

ethn:
(fixed that)

ethn:
OK, I'm going to close this PR but not yet merge into master.  Instead I'm going to start working on the new branched named "1_15".  I want to get this, the latest from master, the generator stuff, the following stuff, and the bootstrap stuff all into that branch and start field testing it a bit.  Then we'll merge the whole thing into master and release 1.15.  I hope this will go fairly quickly, but I feel like we should do this extra branch for a short period since there are SO many changes in the air.

GerryG:
That seems like a great idea as long as we manage it well.  The other PR (generator branch) has some things we will want eventually, but may need some more polish.  Let me know what change sets I should synchronize with to keep up.  I guess we will close out side branches when merged with this branch?  Let's maybe call it a beta branch.  I guess there is no reason not to use develop that way if we want.

ethn:
For the short term, just treat 1_15 as if it were the new master -- all PR's go there, all new branches come off of it, etc.

The generator branch should be much easier to digest/discuss with the new PR.

PR #280, opened 2014-12-29
fixed ref_spec.rb

xithan:
the last change didn't take into account that we introduced a new reference type "partial"


mod/01_core/set/all/references.rb
```diff
@@ -45,15 +45,32 @@ def update_references rendered_content = nil, refresh = false
...
+            # I = inclusion
+            # P = partial (i.e. the name is part of a compound name that is referenced by a link or inclusion)
+            # The partial type is needed to keep track of references of virtual cards. 
+            # For example a link [[A+*self]] won't make it to the reference table because A+*self is virtual and
+            # doesn't have an id but when A's name is changed we have to find and update that link.
+            ref_type = if name == referee_name
+                Card::Chunk::Link===chunk ? 'L' : 'I'
+              else
+                'P'
+              end
```
GerryG: Yes, exactly.  We have a number of ways it can be addressed now, and I understand why it isn't getting addressed in this change.  I think noticing when link addresses are edited will be something you will be interested in for multi-lingual as well.  There might be an event or two connected to these changes at some point.  Or an 'on' option (:on => :references_changed) ?

Should there be a ticket for it?

db/migrate_core_cards/20141208132416_partial_reference_type.rb
```diff
@@ -0,0 +1,9 @@
+# -*- encoding : utf-8 -*-
+
+class PartialReferenceType < Wagn::CoreMigration
+  def up
+    Card.all.each do |card|
+      card.update_references
```
GerryG: So, find_each just makes a simpler loop, one nest rather than two.  Maybe if you were doing an update_all, you would get the ids in groups and do an update_all on the query for that set of ids, but we (often) need to do something to each card so we want the objects (in groups).

mod/01_core/set/all/references.rb
ethn: Sure, a ticket would be good.

The general issue is that we don't have any events at all about chunk-level changes.  At present we just wipe all the references and start fresh.  Would certainly be good to improve that.

db/migrate_core_cards/20141208132416_partial_reference_type.rb
ethn: Right, find_each effectively handles the group loop for you.

ethn: fwiw, making this work in batches is currently the only thing I'm aware of that we need to get done before we merge this.  (though it would be good to get the tests passing; I'm going to work on that separately, since they're broken everywhere)

ethn:
One more comment on the batches thing: we probably need to reset the local cache after every thousand cards or so.  I realized this when doing a bulk import the other night.  Things were bogging down horribly, but then I added the cache clearing, started over, and everything stayed snappy.

Perhaps we should override Card.find_in_batches to add this?

lib/card.rb
```diff
@@ -38,4 +38,25 @@ class Card < ActiveRecord::Base
...
+
+  def self.find_each(options = {})
+    if block_given?
+      find_in_batches(options) do |records|
+        records.each { |record| yield record }
+      end
+    else
+      super(options)
+    end
+  end
```
ethn: find_in_batches looks great!  The above doesn't look like it changes anything about the default #find_each behavior.  Am I missing something?

lib/card/reference.rb
```diff
@@ -45,25 +45,16 @@ def repair_missing_referees
...
-          Rails.logger.info "batch done.  \t\t#{count_updated} total updated"
+      count_updated = 0
+      
+      Card.find_in_batches(:batch_size=>100) do |batch|
+        Rails.logger.debug "Updating references for '#{batch.first.name}' to '#{batch.last.name}' ... "; $stdout.flush        
+        batch.each do |card|
+          count_updated += 1
+          card.update_references 
         end
+        Rails.logger.info "batch done.  \t\t#{count_updated} total updated"
```
ethn: ahhh.  much cleaner!

lib/card.rb
```diff
@@ -38,4 +38,25 @@ class Card < ActiveRecord::Base
...
+  def self.find_in_batches(options = {})
+    if block_given?
+      super(options) do |records|
+        yield(records)
+        Card.cache.reset_local
+      end
+    else
+      super(options)
+    end
+  end
```
ethn: I should probably test this on a deck with a few thousand cards...

ethn:
no major obstacles to merging; I'd just like to understand whether there really is a modification to find_each.

lib/card.rb
xithan: It doesn't change anything but if I don't overwrite find_each then it calls the original find_in_batches instead of Card.find_in_batches

GerryG: And this is a copy of find_each in AR?

Seems like it should call the right find unless the code has an explicit class, right?  I mean, if this exact code is running un the AR::Base class, shouldn't self be the Card class when then executes, and the find_in_batches here on line 44 should be Card.find_in_batches.

Or I' missing something.  Presumably, this didn't work as expected until you did this, right?

ethn: Weird. I wouldn't think that would be true, and last night I could swear I
tried it without it and it worked.

I'll check again.

On Sat, Jan 10, 2015 at 10:00 AM, xithan <notifications@github.com> wrote:

> In lib/card.rb
> <https://github.com/wagn/wagn/pull/280#discussion-diff-22760249>:
>
> > @@ -38,4 +38,25 @@ class Card < ActiveRecord::Base
> >    TRACKED_FIELDS = %w(name type_id db_content trash)
> >    Loader.load_mods if count > 0
> >
> > +
> > +  def self.find_each(options = {})
> > +    if block_given?
> > +      find_in_batches(options) do |records|
> > +        records.each { |record| yield record }
> > +      end
> > +    else
> > +      super(options)
> > +    end
> > +  end
>
> It doesn't change anything but if I don't overwrite find_each then it
> calls the original find_in_batches instead of Card.find_in_batches
>
> —
> <https://github.com/wagn/wagn/pull/280/files#r22760249>.
>



-- 
Ethan McCutchen
One of the Wagneers, Wagn.org

Wagn. How pioneers roll.

s: ethan.mccutchen
t: @intogreater

xithan: It's defined in a separate module Batches. Could that be a reason? Without this it didn't work when I tested it.

GerryG: Hmm, I wouldn't think so.  Being a class method in a module means it will be attached to the class with extend (when included), but when it runs the method, what is self?  It should be Card (the class).  I can't really see why this could be happening.

ethn: fwiw, in the master branch it's actually moved out of card.rb and into a ClassMethods module somewhere (collection.rb, I think), so it was a slightly different loading situation when I saw it pass.

ethn: figured it out.  the References.repair_all was calling find_in_batches directly, so it didn't matter when I commented out find_each.

The reason Philipp was seeing it fail was that it was actually run on a relation object, not on the record object. 

Will push out my changes in a new branch.

PR #281, opened 2014-12-31
Link type

GerryG:
Updated and now targeted to master


lib/wagn/commands.rb
```diff
@@ -100,7 +100,7 @@ def find_spec_file filename, base_dir
       opts[:files] = find_spec_file( file, "#{Wagn.gem_root}" )
     end
     parser.on('-m', '--mod MODNAME', 'Run all specs for a mod') do |file|
-      opts[:files] = "mod/#{file}"
+      opts[:files] = "#{Wagn.gem_root}/mod/#{file}"
```
ethn: this is wrong; it's intended for deck mods.

lib/card/format.rb
```diff
@@ -495,7 +495,7 @@ def build_link href, text=nil
       opts[:class] = case href.to_s
         when /^https?:/                      ; 'external-link'
         when /^mailto:/                      ; 'email-link'
-        when /^([a-zA-Z][\-+.a-zA-Z\d]*):/   ; $1 + '-link'
+        when /^([a-zA-Z][\-+\.a-zA-Z\d]*):/  ; $1 + '-link'
```
ethn: interesting.  I don't really know if that's ever used.

mod/05_standard/set/type/uri.rb
```diff
@@ -0,0 +1,11 @@
+format do
+  include Phrase::Format
+
+  view :core do |args|
+    build_link _render_raw(args), render_title(args) || card.name
```
ethn: interestingly, there is no :title view in the base format class, so it looks like this would break there.  I suppose we should add that view (just make it do the name).

But I think you're right that in HtmlFormat the :title view should not return nil.

mod/05_standard/set/type/uri.rb
```diff
@@ -0,0 +1,11 @@
...
+  end
+
+  view :uri do |args|
+    build_link _render_raw(args)
```
ethn: aliasing to :core could make a mess.  If you expect a uri and get a core, you're not likely to be happy.  there is already a :url view for *all (in base.rb).  I suppose we could alias to that.

mod/05_standard/spec/set/type/uri_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    card = Card.create(:type=>'URI', :name=>'A URI card', :content=>'http://wagn.org/Home')
+    assert_view_select card.format.render('core', :title=>'My Title'), 'a[class="external-link"][href="http://wagn.org/Home"]' do
+      assert_select 'span[class="card-title"]', {:text => 'My Title' }
+    end
+  end
+
+  it "renders a uri view" do
+    card = Card.create(:type=>'URI', :name=>'A URI card', :content=>'http://wagn.org/Home')
+    assert_view_select card.format.render('uri'), 'a[class="external-link"]', {:text => 'http://wagn.org/Home'}
+  end
```
ethn: tests!!  wooot!!

mod/05_standard/spec/set/type/uri_spec.rb
```diff
@@ -0,0 +1,26 @@
+# -*- encoding : utf-8 -*-
+
+describe Card::Set::Type::Uri do
+  it "should have special editor" do
+    assert_view_select render_editor('Uri'), 'input[type="text"][class="card-content"]'
+  end
+
+  it "renders core view links" do
+    card = Card.create(:type=>'URI', :name=>'A URI card', :content=>'http://wagn.org/Home')
```
ethn: stick this in a before block?

db/migrate_core_cards/20141230204340_uri_codename.rb
```diff
@@ -0,0 +1,11 @@
+# -*- encoding : utf-8 -*-
+
+class UriCodename < Wagn::Migration
+  def up
+    contentedly do
+      puts "adding new codename card for URI"
+      codename = 'URI'
+      Card.create! :type_id=>Card::CardtypeID, :name=>codename, :codename=>codename.to_name.key
```
ethn: This will break if there is already a uri card.  Do you think that's likely enough that we should handle it?

mod/05_standard/set/type/uri.rb
ethn: hold on, the "uri" view yields an html link.  hmm, that seems confusing.

ethn: OK, here's what I would propose:
1. As implemented, we use URI for the cardtype name/codename
2. In views, we make uri and url synonymous.  I know that technically they're not, but for now the effort to make it so that "url" balks at non-url uri's seems a bit unnecessary.
3. Both of those views should actually return simple uris in pretty much every format. In other words, something like `http://wagn.org`, NOT `<a href="http://wagn.org">http://wagn.org</a>`
4. For that latter case, we can create a new view called "uri_link".  In many formats this will just be a simple uri, but in HtmlFormat it will have a hypertext link.
5. For all of the following views: uri, url, and uri_link, MOST cards will use the url of the card itself (based on the name), but URI cards will use the content url.

How does that sound?

mod/05_standard/set/type/uri.rb
GerryG: Will add those changes.  A default title view seems like a good idea.

mod/05_standard/set/type/uri.rb
GerryG: I think we just have the wrong name for the view.  The intention is to create a link whose link text is the link rather than the cardname/title.  Maybe this should be done with an inclusion options instead, then other types would just ignore it.

db/migrate_core_cards/20141230204340_uri_codename.rb
GerryG: Yes, maybe.  Shouldn't we handle that in a standard way in migrations?  Seems like you might even want configured options about how to handle name conflicts in the cardname and codename spaces.  We can do like it used to do a long time ago and change the name to URI1 and still use the codename for the original name.  A codename collision is going to be either an error, or re-running the same migration.

mod/05_standard/set/type/uri.rb
ethn: that's pretty close to my proposal, which is ultimately pretty simple.  Is there something there you don't like?

GerryG: I think I'm not understanding you.  You are talking about aliasing uri and url, and I'm saying we leave that alone as far as view names.  The raw view of the new type already returns the bare uri, so we don't need another one for that, just a way to make it a link with the link also as the text.

How about {{a uri card|title:_raw}} or something else.  Some coding in title to say 'use the card's raw content', which is what we want for this feature.

ethn: aliasing uri and url is a minor point.  we already have url and want to make it work; no big deal adding uri, imo, but I don't really care about that.

I do care strongly about NOT turning the nest (inclusion) syntax into a programming language (which would dramatically complicate building the nest interface, which is the centerpiece of our strategy for making wagneering approachable).  So I don't like the _raw idea.  (especially since underscore already has a meaning in name contexts and it doesn't have to deal with views).

Basically, I think we just need a uri_link view.  Isn't that easier?

db/migrate_core_cards/20141230204340_uri_codename.rb
ethn: I like the idea of handling this in a standard way; not sure it needs to be migration specific.  Maybe something like :autoname=>true ?

right, codename collisions aren't really a worry here, imo.

btw, this code is a little harder to read than it needs to be. the variable "codename" is used for the name version.  not a big deal, but since we're here...

mod/05_standard/set/type/uri.rb
GerryG: Ok, url_link makes more sense.  You don't think this view is at all related to the existing url view?

What should that view give on other types?  Seems illogical for it to be a link at all, which is why I think it should come back as core or raw view (core seems right, but I'm not sure it matters).  We just want to see something when this view is used on other types, otherwise debugging is harder.

ethn: I can't find this comment on github, but core seems like an awful idea.
Why would anyone ever want to see "core" of, say, an image card.

And why do you think they would never want a uri_link for a non-URI card?
 seems highly useful.

On Thu, Jan 1, 2015 at 1:04 PM, Gerry Gleason <notifications@github.com>
wrote:

> In mod/05_standard/set/type/uri.rb
> <https://github.com/wagn/wagn/pull/281#discussion-diff-22401972>:
>
> > @@ -0,0 +1,11 @@
> > +format do
> > +  include Phrase::Format
> > +
> > +  view :core do |args|
> > +    build_link _render_raw(args), render_title(args) || card.name
> > +  end
> > +
> > +  view :uri do |args|
> > +    build_link _render_raw(args)
>
> Ok, url_link makes more sense. You don't think this view is at all related
> to the existing url view?
>
> What should that view give on other types? Seems illogical for it to be a
> link at all, which is why I think it should come back as core or raw view
> (core seems right, but I'm not sure it matters). We just want to see
> something when this view is used on other types, otherwise debugging is
> harder.
>
> —
> <https://github.com/wagn/wagn/pull/281/files#r22401972>.
>



-- 
Ethan McCutchen
One of the Wagneers, Wagn.org

Wagn. How pioneers roll.

s: ethan.mccutchen
t: @intogreater

GerryG: It doesn't have any meaning for non-uri types.  If you think otherwise, tell me what it would do?  What would the link be targeted at?  This is what I'm saying.  The core view for most types is not a link, but it is for this type.  So, the url_link view is a link for this type because the target is the raw content of my card. exactly like the core view, it's just that I also want that content in the <a> element content.  There isn't really an equivalent to this for other types.  Core may not be right, maybe raw, but I don't see what other option would give something that makes sense.

ethn: Normal types:

url view (already exists): http://wagnsite.com/cardname
uri_link view (proposed): <a href="http://wagnsite.com/cardname">
http://wagnsite.com/cardname</a>

(use case = when you're trying to get other people to promote a link to
your card.  this happens)

~~~~~

URI type:

url view: (same as raw)
uri_link view: <a href="(raw)">(raw)</a>


~~~~~~~~

If you zoom out, you'll see that it definitely makes sense.  Effectively
what you're saying is that nobody is really interested in linking to a URI
card; what they want to do is link to the resource it links to.  So in
cases where you might normally provide a card url, provide the destination
url instead.


On Thu, Jan 1, 2015 at 4:44 PM, Gerry Gleason <notifications@github.com>
wrote:

> In mod/05_standard/set/type/uri.rb
> <https://github.com/wagn/wagn/pull/281#discussion-diff-22403114>:
>
> > @@ -0,0 +1,11 @@
> > +format do
> > +  include Phrase::Format
> > +
> > +  view :core do |args|
> > +    build_link _render_raw(args), render_title(args) || card.name
> > +  end
> > +
> > +  view :uri do |args|
> > +    build_link _render_raw(args)
>
> It doesn't have any meaning for non-uri types. If you think otherwise,
> tell me what it would do? What would the link be targeted at? This is what
> I'm saying. The core view for most types is not a link, but it is for this
> type. So, the url_link view is a link for this type because the target is
> the raw content of my card. exactly like the core view, it's just that I
> also want that content in the element content. There isn't really an
> equivalent to this for other types. Core may not be right, maybe raw, but I
> don't see what other option would give something that makes sense.
>
> —
> <https://github.com/wagn/wagn/pull/281/files#r22403114>.
>



-- 
Ethan McCutchen
One of the Wagneers, Wagn.org

Wagn. How pioneers roll.

s: ethan.mccutchen
t: @intogreater

GerryG: Sorry, being a little dense.  Oh, and I think you get just the url to the card in the base Format class.

ethn:
ok if I hold off on further comments until ATP?

(that's "all test passing" for the uninitiated; I think Gerry coined that one a while back)

GerryG:
Is there something wrong with the migration?  Some tests are failing because it didn't run.

db/migrate_core_cards/20141230204340_uri_codename.rb
GerryG: In my use of codename, this is the codename, just not in 'key' form.  It is also the initial name, but it may not be.   'autoname' doesn't seem the right name for this feature, maybe auto_new_name ?  Doesn't seem that related to this ticket in any case.

ethn:
Oh, the tests don't run the migration; they seed from the data.  I gather I need to regenerate the data to get the tests to pass?

db/migrate_core_cards/20141230204340_uri_codename.rb
ethn: the general solution isn't, true, but we've got to get a fix of some sort
in for this before we release it.  migration issues are a huge headache.

On Fri, Jan 2, 2015 at 8:01 PM, Gerry Gleason <notifications@github.com>
wrote:

> In db/migrate_core_cards/20141230204340_uri_codename.rb
> <https://github.com/wagn/wagn/pull/281#discussion-diff-22429000>:
>
> > @@ -0,0 +1,11 @@
> > +# -*- encoding : utf-8 -*-
> > +
> > +class UriCodename < Wagn::Migration
> > +  def up
> > +    contentedly do
> > +      puts "adding new codename card for URI"
> > +      codename = 'URI'
> > +      Card.create! :type_id=>Card::CardtypeID, :name=>codename, :codename=>codename.to_name.key
>
> In my use of codename, this is the codename, just not in 'key' form. It is
> also the initial name, but it may not be. 'autoname' doesn't seem the right
> name for this feature, maybe auto_new_name ? Doesn't seem that related to
> this ticket in any case.
>
> —
> <https://github.com/wagn/wagn/pull/281/files#r22429000>.
>



-- 
Ethan McCutchen
One of the Wagneers, Wagn.org

Wagn. How pioneers roll.

s: ethan.mccutchen
t: @intogreater

GerryG:
There is a new type card with codename, so yes.

But, I'm working on it a little more.  I think maybe we can resolve the location stuff here and merge that forward (carefully merging moved files).  I think that stuff belongs in the Format classes, that's where we are including it.  The session stuff goes in HtmlFormat and link_to_page can be replaced with build_link, which is really what we want.

GerryG:
The failing tests are all related to the missing type card

GerryG:
Cool, all green.

lib/wagn/commands.rb
GerryG: Fixed

db/migrate_core_cards/20141230204340_uri_codename.rb
GerryG: Added code for this.  Let me know if the option needs a different name.  Tested both failure conditions manually.

GerryG:
Added one more commit.  Moved Wagn::Location stuff to Format or HtmlFormat.  Can revert if you don't like, or finalize if you do.  This was more inspired by having thought about moving it on the split branch.

The changes to use build_link in more places are related to the new link type, but it does add some breadth to the scope.  I think it makes a good group of changes.

GerryG:
I think I found it.  The fixtures had some duplicate ids on card_acts.

GerryG:
This is the case where you just want to kick it again.  That one token test failed as it does sometimes.  Is there a way to make it consistent?

ethn:
I would love to make it consistent; I'm not sure this doesn't indicate a
real bug.  Every now and then people do seem to have token problems.

I kicked it.  Re project permissions, my best guess is that you have to
sign up for a semaphore account to get things going. Presumably the email
invitation only works if it's associated with an account?


On Thu, Jan 8, 2015 at 11:12 AM, Gerry Gleason <notifications@github.com>
wrote:

> This is the case where you just want to kick it again. That one token test
> failed as it does sometimes. Is there a way to make it consistent?
>
> —
> <https://github.com/wagn/wagn/pull/281#issuecomment-69222385>.
>



-- 
Ethan McCutchen
One of the Wagneers, Wagn.org

Wagn. How pioneers roll.

s: ethan.mccutchen
t: @intogreater

db/migrate_core_cards/20141230204340_uri_codename.rb
```diff
@@ -0,0 +1,17 @@
+# -*- encoding : utf-8 -*-
+
+class UriCodename < Wagn::Migration
+  def up
+    contentedly do
+      puts "adding new codename card for URI"
+      targetname = 'URI'
+      codename = targetname.to_name.key
+      if Card.exists? codename.to_sym
+        raise "Migration failed, codename #{codename} taken"
```
ethn: I think this pretty much would happen anyway ;)

mod/01_core/set/all/initialize.rb
```diff
@@ -1,13 +1,24 @@
...
+  def find_unused_name base_name
+    test_name = base_name
+    add = 1
+    while Card.exists?(test_name) do
+      test_name = "#{base_name}#{add}"
+      add +=1
+    end
+    test_name
+  end
 end
```
ethn: did you look at autoname?  It does pretty much this.  It might be worth improving that code to do this, but I'm not sure we need two separate methods.

In any event, we can't merge this as is because it reduces code coverage.  We don't have to unit test every bit of code, but we do need to make sure the code at least gets touched.

mod/04_settings/set/right/comment.rb
```diff
@@ -33,7 +33,7 @@
               %{
                 <div class="group-option">
                   #{ check_box_tag( "#{option.key}-perm-checkbox", option.name, checked, :class=>'perm-checkbox-button'  ) }
-                  <label>#{ link_to_page option.name, nil, :target=>'wagn_role' }</label>
+                  <label>#{ build_link option.name, nil, :target=>'wagn_role' }</label>
```
ethn: I suspect the "target" would break here.

mod/05_standard/set/self/account_links.rb
```diff
@@ -5,8 +5,9 @@
     #ENGLISH
     links = []
     if Auth.signed_in?
-      links << link_to_page( Auth.current.name, nil, :id=>'my-card-link' )
+      links << build_link( Auth.current.cardname, nil, :id=>'my-card-link' )
```
ethn: guessing wouldn't work

spec/lib/card/content_spec.rb
```diff
@@ -89,7 +89,7 @@
     "\n        More urls: ",
     "<a class=\"external-link\" href=\"http://wagn.com/a/path/to.html\">wagn.com/a/path/to.html</a>",
     "\n        ",
-    "<a class=\"external-link\" href=\"http://localhost:2020/path?cgi=foo&bar=baz\">http://localhost:2020/path?cgi=foo&bar=baz</a>",
+    "<a class=\"external-link\" href=\"http://localhost:2020/path?cgi=foo+bar=baz\">http://localhost:2020/path?cgi=foo+bar=baz</a>",
```
ethn: what does ?cgi=foo+bar=baz mean??

spec/lib/card/format_spec.rb
```diff
@@ -3,33 +3,80 @@
...
+      expect(cobj.to_s).to eq url_text5
+    end
+
+    it "should format html links" do
+      cobj = Card::Content.new url_text1, html_format
+      expect(cobj.to_s).to eq "with external free link <a class=\"external-link\" href=\"http://localhost:2020/path?cgi=foo&amp;bar=baz\">http://localhost:2020/path?cgi=foo&amp;bar=baz</a>"
+      cobj = Card::Content.new url_text2+url_text3+url_text4, html_format
+      expect(cobj.to_s).to eq url_text2+url_text3+url_text4
+      cobj = Card::Content.new url_text5, html_format
+      expect(cobj.to_s).to eq "external with port: <a class=\"external-link\" href=\"http://localhost:2020/path?cgi=foo+bar=baz\">http://localhost:2020/path?cgi=foo+bar=baz</a> after "
```
ethn: seems like the above have two different forms of wonky urls, no?

mod/05_standard/set/self/account_links.rb
ethn: ...or `link to`

ethn:
Hoping to get this into 1.14.5, which I may get out tomorrow night.  Any chance the remaining fixes will be in by then?

spec/lib/card/content_spec.rb
GerryG: I'm not totally positive what is correct for this.  It is entity encoding the & now, and it wasn't before.  I think this is related to using build_link vs. that other odd method (link_to_page ?)  I think it get's interpreted the same way encoded or not.

db/migrate_core_cards/20141230204340_uri_codename.rb
GerryG: Yeah, it should.  I guess that isn't really needed.

mod/01_core/set/all/initialize.rb
GerryG: autoname seems like the wrong name for this.  Does it do that same thing, and where it is?  Seemed to me that was an old feature that doesn't quite make sense now.  I did test it manually, and it wasn't really my idea to add this, I thought you were concerned about this particular name collision on existing sites.

mod/05_standard/set/self/account_links.rb
GerryG: I think this stuff works.  I don't recall as I comment, but I think this is just a matter of accepting options to be used as attributes of the link element.

spec/lib/card/format_spec.rb
GerryG: It's the & character again, and I don't think this breaks anything.

GerryG:
I'll do what I can.  I'm not totally clear on what you want fixed.  I don't think the & stuff is significant, and I actually think that is an important improvement to always us build_link as it responds appropriately to different formatters.  I can add a test for the feature for name duplicates.  It could be done without an option through new if you'd prefer (just use it in the migrations to find the name before calling create.

spec/lib/card/content_spec.rb
ethn: + and & have completely different meanings.  + is equivalent to a space in most urls.  These urls are clearly broken.

ethn:
The & stuff is clearly broken, and I pointed out several other cases where build_link has not been sufficiently modified to do what link_to_page did.  And find_unused_name is not good enough for the core api.  The functionality is closely related to the autoname functionality, so imo that's the right route to look in if we want a generalized solution for the migration issue, but if not it could be handled narrowly in the migration.

spec/lib/card/content_spec.rb
GerryG: The point isn't the correctness of the URLs, but the handling of &.  I don't think it is a problem that they are encoded, and it may be correct that they get encoded for this context.  In other words, they may have been more broken the way it was.  Can you show that this is functionally wrong?

mod/01_core/format/html_format.rb
```diff
@@ -216,8 +265,8 @@ def unsupported_view view
     end
 
     def final_link href, opts={}
-      text = opts[:text] || href
-      %{<a class="#{opts[:class]}" href="#{href}">#{text}</a>}
+      text = (opts.delete(:text) || href).dup
+      content_tag :a, text, opts.merge(:href=>href)
```
GerryG: Note that the options from build_link are passed here, and now they are all passed to content_tag as options.  They tests would be failing otherwise and they aren't.

This is also where the & get's html encoded.  We could try to do something here to prevent that, but I think that would actually be wrong.  It isn't clear to me why link_to doesn't similarly process the & characters, but the question still remains if the new behavior is broken in any functional way.

If you make the case that it is broken, I'd be happy to try to fix it, but I'm still not convinced it is.  I think it was an oversight with how it did work, and I think it is a bad thing to have more than one piece of code generating final links.

GerryG:
Use case to look at: international characters in a URL.  It will show up a &entity_name;, and therefore any real & is encoded as &amp;  Probably, most implementations just leave an & if it doesn't match an entity for other reasons, but this is an ambiguity.

GerryG:
To me, this is correct now, but I concede that I did combine two things which I could still split out at this point.  Separately, each is a fairly narrow change, so I don't think it is so bad.

Remember that the link processing that this applies to is part of the content chunk processing.  It would happen to free urls in the content, probably to the core view.  I think that means we would get html links in a plain text formatter.  I may attempt to add that test case now, predicting it will break in master and work with this change.

GerryG:
Well, added the spec and it doesn't break the way I expected.  The update does behave better in that when the link is the text, it leaves off the [link] part that is duplication.

mod/01_core/format/html_format.rb
ethn: My comments were not about encoding.  Encoding is great.  But you changed a bunch of tests from `http://localhost:2020/path?cgi=foo&bar=baz` to `http://localhost:2020/path?cgi=foo+bar=baz`. That's not an encoding change, it's just a screwed up url.  The parameters aren't even parseable.  I gather this is fixed now?

I think I was wrong about build_link needing to handle url_options; in general the assumption is that this is handled before we get there.  (I do like the idea of moving the text into the options hash, but that's not vital).  This definitely makes sense given the different contexts that we want to use this.

The reason I got confused was that link_to_page did handle url_options for :type and :view.   That was a bit of a hack, but I'm concerned that this could mean there's a hidden bug now.

spec/lib/card/format_spec.rb
ethn: Re the first link above, it's clearly good to encode the html in the "text" part of the link. In the href attribute, it seems odd to be altering a url, which creates a serious risk of double encoding.  Can't we just leave it unaltered there?

Re the second link, this is just a screwy CGI query, although in this case at least it starts off that way.

ethn:
re international characters, they're supposed to be CGI encoded (with %'s) in the context of an href attribute and  HTML encoded (with &'s) in the context of rendered html text.

mod/01_core/format/html_format.rb
ethn: re "if you make the case that it is broken".  My take is that the case when we change this stuff we have to make the case that we're NOT breaking things with a lot of field testing.   If this dirty work falls to me, then it might be a while before we get this merged in.  I've just got to be cautious, because if I merge it in and it screws up urls all over the place, I'm going to have to do all the clean up.

lib/wagn/location.rb
```diff
@@ -1,93 +0,0 @@
...
-    end
-  end
-
-
-  # Links ----------------------------------------------------------------------
-
-  def link_to_page( text, title=nil, options={})
-    title ||= text
-    url_options = {}
-    [:type, :view].each { |k| url_options[k] = options.delete(k) if options[k] }
```
ethn: ok, I looked more closely and can't find any place where :type and :view are passed explicitly or could even be hidden in the options hash somehow.  I think we're good on that.

mod/04_settings/set/right/comment.rb
ethn: I was wrong about these, sorry.

mod/01_core/format/html_format.rb
ethn: sorry, I followed that stackoverflow link, and I think that pretty well settles the last issue I had.  It looks like you addressed all my other concerns, so we're good now; I'll get this merged into 1.14.5.

I should have said this before, btw, but I really love that you consolidated these two.  I'm not yet totally in love with our link handling, but this was a BIG step in the right direction!!

GerryG: You'll note that I changed those back and compensated by changing the expected result to an encoded &.

Also added test case for text rendering of links which should not change a regular url.  If the recognized url is missing something (like http://, a bare host+path), then the url is different from the parsed text, so it does the [] thing.

lib/wagn/location.rb
GerryG: I think I looked for that, but thanks for double checking.

mod/04_settings/set/right/comment.rb
GerryG: Not surprising you didn't see it as it isn't that obvious.  I could have added that comment earlier.  All good.  Took me a bit to realize it was only handling class explicitly before.

mod/05_standard/set/self/account_links.rb
ethn: I still prefer the two arg approach, fwiw.  It's weird, imo, to have all the explicit nil calls.  No need to do that now...

GerryG: When we start requiring ruby 2+, we can address that with optional keyword arguments.  The nil is typically for the link text argument that is optional, but because the next positional argument is used it needs to be explicit.

ethn: Understood.  I was suggesting:
````
def build_link href, options
````
...but that if you had link text, you stick it in the options hash like:
````
build_link http://wagn.org, :text=>'the Wagn site'
````
Actually, I was playing with names other than build_link (link_to?), but you get the point.

GerryG: Ah, yes, that would be fine.  I could add that pretty easily.

Well, link_to is what the rails method that link_to_X was using before.  I think it may be used, maybe quite a bit.  Then 'format_link' may be more accurate than build_link.  We could bring back the other method name (link_to_p...) as it was but use this API (with suggested change with :text).  I'll add whatever you like best.  I'm even ok with build_link.

GerryG:
I don't really see why the build is failing.  The stuff that got merged in (fixed on my master branch, I can make it a separate PR) is just about running tests with the wagn command, it shouldn't really effect anything else.  On my semaphore account, they work.  I run the specs and cucs through the wagn command.  Since the ethan/wagn build uses the cucumber and rspec commands directly, these just merged changes should not have mattered.

PR #283, opened 2015-01-05
Rails 4

ethn:
This is an experimental branch to see how hard this upgrade is.


ethn:
I got the tests running (locally at least; I suspect we'll have to update the Gemfile to get things going on semaphore).  But about a quarter of them are still failing.

lib/card/action.rb
```diff
@@ -44,7 +44,8 @@ def delete_old_actions
   class Action < ActiveRecord::Base
     belongs_to :card
     belongs_to :act,  :foreign_key=>:card_act_id, :inverse_of=>:actions 
-    has_many   :changes, :foreign_key=>:card_action_id, :inverse_of=>:action, :dependent=>:delete_all
```
ethn: "changes" already means something to active record (Dirty), and rails started getting grumpy about that.

lib/card.rb
```diff
@@ -14,11 +14,9 @@ class Card < ActiveRecord::Base
...
   has_many :references_from, :class_name => :Reference, :foreign_key => :referee_id
   has_many :references_to,   :class_name => :Reference, :foreign_key => :referer_id
-  has_many :acts, :order => :id
-  has_many :actions, :order => :id, :conditions=>{:draft => [nil,false]}
-  has_many :drafts, :order=>:id, :conditions=>{:draft=>true}, :class_name=> :Action
-
-  cache_attributes 'name', 'type_id' # review - still worth it in Rails 3?
+  has_many :acts, -> { order :id }
+  has_many :actions, -> { where( :draft=>[nil,false]).order :id }
+  has_many :drafts, { :class_name=> :Action }, -> { where( :draft=>true ).order :id }
```
ethn: It's a new lambda shortcut as of ruby 1.9.1 (http://ruby.about.com/od/newinruby191/qt/LambdaOperator.htm)

The notion has been that the upgrade to Rails 4 would imply moving Wagn's minimal ruby requirement to 1.9.3 (with a preference for 2).

I wouldn't swear that it's impossible to support older rails versions, but it would be a significant challenge I don't see a ton of value there, do you?

lib/wagn/application.rb
```diff
@@ -9,9 +9,10 @@
   # Bundler.require(:default, :assets, Rails.env)
 end
 
-module ActiveSupport::BufferedLogger::Severity
-  WAGN = UNKNOWN + 1
-  
+module ActiveSupport::Logger::Severity
+#  WAGN = UNKNOWN + 1
+  WAGN = 6
```
ethn: yeah, though I haven't really played with it yet and I wouldn't be surprised if I just broke it.

Philipp put this together as a way of logging wagn-specific performance details.

(His solution is better than mine, but "UNKNOWN" was breaking for some reason.

lib/wagn/application.rb
```diff
@@ -106,10 +109,10 @@ def paths
         add_gem_path paths, 'gem-mod',             :with => 'mod'
         add_gem_path paths, 'gem-assets',          :with => 'public/assets'
 
-        paths['app/models'] = []
-        paths['app/mailers'] = []
-        paths['app/views'] = File.join( Wagn.gem_root, 'lib/card' )
-        paths['local-mod'] = approot_is_gemroot? ? [] : 'mod'
+#        paths['app/models'] = []
+#        paths['app/mailers'] = []
+#        paths['app/views'] = File.join( Wagn.gem_root, 'lib/card' )
```
ethn: you can't do []= on paths any more.

They actually restructured the path handling considerably.

lib/card.rb
GerryG: I figured out the -> thing, but then the question is whether this way of setting the relation (conditions and order, etc.) works with 3.2.x, or whether we'd need a version conditional here.

GerryG:
Since we can't just leap to 4.x (or can we?), don't we need to make changes that also don't break current Rails versions?  Now I can't remember what the overlap was like with the 2 -> 3 update.  Seems like it was a more all or nothing change, right?

lib/card.rb
ethn: it's already there (see gemspec)

ethn: we should add the ruby version directly, but the rails version specification will mean that ruby 1.9.3 is required.

ethn:
We are leaping to ~> 4.2.  This is actually *less* specific before when we were setting the point release (3.2.16).  see the gemspec.

lib/card.rb
GerryG: So we would have both a 3.2.x and a 4.2.x version of our gem?  I don't think we want to require rails 4 yet, and the platform use case means we'll want to be able to conform to the external project's version.

ethn: I don't know what you meant about BOTH 3.2.x and 4.2.x version.  The initial idea was to stop supporting 3.2.16 (which was the only version we supported before; not 3.2.15 or 3.2.17) and start using ~> 4.2 (which is different from 4.2.x because it includes 4.3 but not 4.1).

It took me a while to figure out what "platform use case" meant.  Could we refer to this as "wagn-as-rails-engine" or "decko-rails"?  If anything, the "platform use case" would suggest the old way, not the new (let's call the old way "wagn-as-platform" or maybe "decko-based").

At any rate, you're point is well taken.  Whereas in our primary use case (decko-based) we we're fine to specify the rails version, in our new use case (decko-rails) we need to support a decent range of versions or we're going to cause problems.

As for where to draw the line, it would be a lot simpler to support 4.x rather than 3.2.x and 4.x.  And since Rails 4 has been out for a year and a half now, we'd only need 3.2.x to support adding decko-rails to fairly old applications.  But presumably this is *your* use case, which makes it an important one.  Exactly what version are you using now?

Here's what I would propose: let's keep moving this branch forward with a focus on figuring out what has to happen to get things to work in the latest rails code (4.2).  We've already got it working in 3.2.16 (the current code).  Once we have those both worked out, we can better assess how hard it would be to go back as far as 3.2.x.  If we're just talking about a dozen or so "if" statements to keep 3.2 going, it's probably worth it, at least until 2.0.  I'm not at all excited about limiting ourselves to 3.2 functionality much past Decko 1.0.

I'm really itching to start requiring 1.9, too.  That was released many years ago now, so it feels ancient not to be moving forward on that!

ethn: "Can't we just change the spec to get 3.2.something + up to latest 3.x.y ?"

If you mean, not supporting 4.x, that's a non starter.  Wagn needs to stay fairly close to the latest rails so that new rails sites can use decko-rails.  The vast majority of installations in the past 1.5 years use rails 4.   I'm much more open to support 3.2.x than to NOT supporting 4.x.

At some point, of course, we will need to drop 3.x  But so long as we support it in the initial releases, *old* rails sites can use *old* decko-rails versions, and when they want to upgrade their decko they have to upgrade their rails, too.  That seems sensible enough, no?

ethn: ok, great, it sounds like we're on the same page.  I do think we should try to get to >3.2 or >3.2.x if there are problems with the early versions.  

This probably means we need a separate semaphore setup for testing whatever we decide is the lowest ruby version and the lowest rails version we support.

By the way, I think wagn is currently broken in ruby 2.2.0 (but working well in 2.1.5)

ethn: fyi, I will still probably start by getting it to work on 4.2 and then trying to connect the dots.  It would be my preference to try to keep this to one version for now.  Perhaps what we'll do is jump to requiring ~> 4 at Decko 1.0 but, as you suggest, continue supporting old 3.2.x versions in limited Decko 0.x releases?

ethn:
Like a lot of the other recent PRs, I'm closing this one and will make a new PR to 1.15 when ready.

-e

PR #284, opened 2015-01-06
advanced following

xithan:
a short explanation how this is supposed to work:

There is now a *follow setting and user-specific rules like 
- Basic+*type+*follow+Joe
- *all+*follow+Jim
- Home+*self+*follow+John

These rules are pointers and possible values are follow option cards like 
[[always]]             -> always notify me about changes
[[created by me]] -> only if card was created by me
[[edited by me]]   -> only if card was edited by me
[[never]]               -> don't notify me

These rules are only used for caching and fast access. To edit a follow rule you change your +*following card. For example if you add "Basic+*type+always" to "Joe+*following" that adds [[always]] to the rule card Basic+*type+*follow+Joe.


lib/card/query/ref_spec.rb
```diff
@@ -1,9 +1,14 @@
...
-      :link_to  => [ :out, 'L' ],   :linked_to_by   => [ :in, 'L' ],
-      :include  => [ :out, 'I' ],   :included_by    => [ :in, 'I' ]
+      # syntax:
+      # wql query key => [ direction, {reference_type} ]
+          # direction      = :out | :in
+          # reference_type =  'L' | 'I' | 'P' 
+
+      :refer_to => [ :out, 'L','I' ], :referred_to_by => [ :in, 'L','I' ],
+      :link_to  => [ :out, 'L' ],     :linked_to_by   => [ :in, 'L' ],
+      :include  => [ :out, 'I' ],     :included_by    => [ :in, 'I' ]
```
GerryG: It would simplify a few things to make the 'types' a string rather than an array of single characters.  Easy enough to iterate over the chars in a string (more than 1 in length.

lib/card/query/ref_spec.rb
```diff
@@ -16,9 +21,15 @@ def initialize key, val, parent
     end
 
     def to_sql *args
-      dir, type = REFERENCE_DEFINITIONS[ @key.to_sym ]
+      dir, *type = REFERENCE_DEFINITIONS[ @key.to_sym ]
       field1, field2 = REFERENCE_FIELDS[ dir ]
-      cond = type ? ["ref_type='#{type}'"] : []
+      cond = case type.size 
+        when 0 then [] 
+        when 1 then ["ref_type='#{type.first}'"]
```
GerryG: This would be type(.each_character).map{|letter|"ref_type='#{type.first}'"} when type.length < 2 and then the else case.  Not that case statements are bad, but aren't two cases better than three?

GerryG: I can see it either way, but to me this seems like having two limit cases when they really can fold into one.  Small issue in any case.

ethn:
I'm going to close this for now.  When you're ready to re-open, would you be willing to make the PR (and all other PR's not intended for immediate release) to the new wagn/1_15 branch?

PR #286, opened 2015-01-08
jasmine rails integration

chuenlok:
test with semaphore


lib/wagn/generators/wagn/templates/Rakefile
```diff
@@ -3,5 +3,6 @@
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
 require File.expand_path('../config/application', __FILE__)
+require 'jasmine'
```
ethn: are you sure this is needed here?  It's an odd place for this require...

lib/wagn/generators/wagn/templates/config/application.rb
```diff
@@ -16,6 +16,10 @@ class Application < Wagn::Application
   
     
     # Email configuration
+
+    if !Rails.env.production? && Object.const_defined?( :JasmineRails )
+        config.assets.enabled = true
+    end
```
ethn: We don't want this here in the templates.  Perhaps in the environment files?  Do we need it in development or only test?

lib/wagn/generators/wagn/wagn_generator.rb
```diff
@@ -27,12 +27,15 @@ class WagnGenerator < Rails::Generators::AppBase
   public_task :create_root
   
 ## should probably eventually use rails-like AppBuilder approach, but this is a first step.  
-  def dev_setup
-    @wagn_path = options['gem-path']
+  def dev_setup  
     if options['core-dev']
-      @wagn_path ||= ask("Enter the path to your local wagn gem installation: ")
+      @wagn_path = options['gem-path'] || ask("Enter the path to your local wagn gem installation: ")
```
ethn: I liked this outside of the core-dev block; it's a nice shortcut for using a locally modified gem even when not doing core dev per se.

lib/wagn/generators/wagn/templates/config/application.rb
ethn: (this is the only one important enough to slow down the merge)

PR #287, opened 2015-01-09
1 14 bugs

ethn:
This is to fix two ugly bugs that were recently reported.  One involved signup breakage on the CARB site and the other was WQL returning multiple copies of the same card on Wikirate.org.

### DIFF
--- a/lib/card/query/ref_spec.rb
+++ b/lib/card/query/ref_spec.rb
@@ -20,7 +20,7 @@ class Card::Query
       field1, field2 = REFERENCE_FIELDS[ dir ]
       cond = type ? ["ref_type='#{type}'"] : []
 
-      sql =  %[select #{field1} as ref_id from card_references]
+      sql =  %[select distinct #{field1} as ref_id from card_references]
       if @val == '_none'
         cond << "present = 0"
       else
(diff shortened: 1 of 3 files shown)


lib/card/query/ref_spec.rb
```diff
@@ -20,7 +20,7 @@ def to_sql *args
       field1, field2 = REFERENCE_FIELDS[ dir ]
       cond = type ? ["ref_type='#{type}'"] : []
 
-      sql =  %[select #{field1} as ref_id from card_references]
+      sql =  %[select distinct #{field1} as ref_id from card_references]
```
ethn: without this, multiple references to the same card can lead to WQL returning duplicate cards.

mod/05_email/set/type/email_template.rb
```diff
@@ -5,7 +6,7 @@ def clean_html?
 def deliver args={}
   begin
     mail = format.render_mail(args)
-    mail.deliver 
+    mail.deliver
```
ethn: without the change below, the "mail" variable can end up as a String and will fail here.

ethn: It's worth discussing whether it would be better to wrap the above in an Auth.as_bot block.  Issue is whether the mail view can render anything that we wouldn't want an unpermissioned user to see.

Currently tempted to make that change...

PR #288, opened 2015-01-10
putting the pieces together

ethn:


ethn:
did this so I could make sure tests pass before merging into master

PR #289, opened 2015-01-13
Generator

GerryG:
Mostly, I want to see this building in semaphore.  This is downstream of the card split, so go here to discuss: https://github.com/GerryG/wagn/pull/41   You can either merge the card split first, then this will be reduced to it's own changes, or just do them both by merging this.  Both are ATP now, but I think we want to do some more  finalizing on the new/updated generators and wagn sub-commands.


GerryG:
You have any idea what that random failure is about?  Maybe a 'within on second' error?  Race condition?

GerryG:
I'm trying to figure out a test scenario for the platform case.  I think it is something like this:

rails new my_platform_app
cp /home/runner/Gemfile /home/runner/my_platform_app/
# Gemfile is a configuration mainly with: gem 'wagn', :path => '../wagn'
cp (database.yml to new app/config)
cd my_platform_app
rails generate card base # should probably use table prefix and mount point to test those features

Note, rails generate vs. wagn.  I think this is how it should work, the generators available for both scenarios, wagn app and rails app with cards.  The maybe we can demonstrate a use case with another framework (Sinatra or ?)

PR #292, opened 2015-01-17
load sets with more than one anchor

xithan:
loads also sets in subdirectories


GerryG:
Was wondering when we would need this.  Do we have any modules that would test it yet?

xithan:
not yet, so we don't have automated tests for this, but I verifed it locally with a type_plus_right set

PR #297, opened 2015-01-21
updated set generator

xithan:
to handle multiple anchors


xithan:
improved performance logger

Configuration is still the same:
set `config.log_level= :wagn` and use `performance_logger` hash for configuration
```` 
  config.performance_logger = {
      :methods => [:event, :search, :fetch, :view],  # choose methods to log
      :limit => 100,                                 # show only method calls that are slower than 100ms
      :level => 3,                                   # show nested method calls only up to level 3
      :details=> true                                # show method arguments and sql
   }
````

and you get something like this:
1__bundle_exec_wagn_s__ruby__und_home_-_my_wagn_und_log_rb_ _wagn__git__master_

**Comment by ethn** [comment]
code looks solid at first glance.  looking forward to playing with it.

File: lib/card/generators/set/USAGE [hunk 1]
```diff
@@ -1,9 +1,16 @@
...
 
-Example:
+Examples:
     wagn generate set standard type tshirt
 
     This will create:
-        mods/05_standard/sets/type/tshirt.rb
-        spec/mods/05_standard/sets/type/tshirt_spec.rb
+        mods/standard/sets/type/tshirt.rb
+        mods/standard/spec/sets/type/tshirt_spec.rb
```
ethn: we should probably singularize mods and sets in these examples at some point ;)

PR #298, opened 2015-01-26
include search box in ace editor

chuenlok:
the browser's default search function is not working well in ace editor. Just include the ace editor's search function.


chuenlok:
I also put the jasmine rails description in this pr.

PR #299, opened 2015-01-27
Inherited type

GerryG:
An attempt at this ticket: http://wagn.org/Inherit_Type_from_Structure_or_Default

It may need some cache clearing when Type+*type+*default rules are added or removed.

### DIFF
--- a/lib/card/set.rb
+++ b/lib/card/set.rb
@@ -200,9 +200,11 @@ class Card
         # FIXME - this does not properly handle anchorless sets
         # There are special hacks for *all, but others (like *rstar) will not be found by
         # include_set_modules, which will look for Card::Set::Rstar, not Card::Set::Rstar::Blah
-        
+        # This issue appears to be addressed by making the entries, in modules arrays.
+        # If yes remove this comment.
+
         to_file = "#{Wagn.paths['tmp/set'].first}/#{set_pattern}/#{seq}-#{anchors.join '-'}.rb"
-        anchor_modules = anchors.map { |a| "module #{a.camelize};" }.join
+        anchor_modules = anchors.map { |a| "module #{a.camelize};" }.join # use explicit '', not $,
         file_content = <<EOF
 # -*- encoding : utf-8 -*-
 class Card; module Set; module #{set_pattern.camelize}; #{anchor_modules}
--- a/lib/card/set_pattern.rb
+++ b/lib/card/set_pattern.rb
@@ -77,16 +77,8 @@ EOF
         else
           Card.fetch_id @anchor_name
         end
-        if self.class.pattern_code == 'type' and
-             default_rule = Card.fetch(card.type_cardname.trait_name(:type).trait_name(:default),
-                                       :skip_modules=>true, :skip_virtual=>true) and
-             type_code = default_rule.type_code and
-             mod_key = "Type::#{type_code.to_s.camelize}" and
-             (Card::Set.modules[:nonbase_format].values + [Card::Set.modules[:nonbase]]).
-               any?{|hash| hash[mod_key]}
-          @inherited_type_module_key = mod_key
-        end
       end
+        @inherited_key = nil
       self
     end
 
@@ -100,17 +92,17 @@ EOF
       end
     end
 
-    def inherited_key modules_hash
+    def inherited_module_list modules_hash
       module_key && modules_hash[ module_key ] ||
-        @inherited_type_module_key && modules_hash[ @inherited_type_module_key ]
+        @inherited_key && modules_hash[ @inherited_key ]
     end
 
     def module_list
-      inherited_key Card::Set.modules[ :nonbase ]
+      inherited_module_list Card::Set.modules[ :nonbase ]
     end
 
     def format_module_list klass
-      hash = Card::Set.modules[ :nonbase_format ][ klass ] and inherited_key hash
+      hash = Card::Set.modules[ :nonbase_format ][ klass ] and inherited_module_list hash
     end
 
     def anchor_codenames
@@ -145,5 +137,24 @@ EOF
       end
     end
   
+  end
+  class TypeSet < SetPattern
+    def initialize card
+      super
+      if mod_key = lookup_inherited_key(card)
+        @inherited_key = mod_key
+      end
+    end
+
+    def lookup_inherited_key card
+      default_rule = card.rule_card(:default) and
+        type_code = default_rule.type_code and
+        default_rule.cardname.size > 2 and
+        default_rule.left.right.codename == self.class.pattern_code
+        mod_key = "Type::#{type_code.to_s.camelize}" and
+        ( Card::Set.modules[:nonbase_format].values +
+          [Card::Set.modules[:nonbase]] ).any?{|hash| hash[mod_key]} and
+        mod_key
+    end
   end
 end
--- a/lib/card/set_pattern.rb
+++ b/lib/card/set_pattern.rb
@@ -141,7 +141,7 @@ EOF
   class TypeSet < SetPattern
     def initialize card
       super
-      if mod_key = lookup_inherited_key(card)
+      if !module_key && mod_key = lookup_inherited_key(card)
         @inherited_key = mod_key
       end
     end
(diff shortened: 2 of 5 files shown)


GerryG:
Noted in test output that this line is logging to the output: +        puts "resetting local cache"

Looks like a left-over debug thing, probably not wanted in the merged code.

lib/card/set.rb
```diff
@@ -200,9 +200,11 @@ def write_tmp_file set_pattern, anchors, from_file, seq
         # FIXME - this does not properly handle anchorless sets
         # There are special hacks for *all, but others (like *rstar) will not be found by
         # include_set_modules, which will look for Card::Set::Rstar, not Card::Set::Rstar::Blah
-        
+        # This issue appears to be addressed by making the entries, in modules arrays.
+        # If yes remove this comment.
+
```
GerryG: No real change here.  Just noticing that the comment may be obsolete.  If not, we should finally fix this.  Seems like a necessary part of the recent PR about multi-anchor sets.

lib/card/set_pattern.rb
```diff
@@ -91,12 +97,17 @@ def module_key
       end
     end
 
+    def inherited_key modules_hash
+      module_key && modules_hash[ module_key ] ||
+        @inherited_type_module_key && modules_hash[ @inherited_type_module_key ]
```
GerryG: We could avoid setting this when none of the hashes have an entry for this key, but this test isn't all that expensive anyway.

mod/01_core/set/all/type.rb
```diff
@@ -14,9 +15,11 @@ def type_code
...
 
 def type_name
-  return if type_id.nil?
-  type_card = Card.fetch type_id.to_i, :skip_modules=>true, :skip_virtual=>true
-  type_card and type_card.name
+  type_card.try :name
+end
+
+def type_cardname
+  type_card.try :cardname
```
GerryG: I think this change is a good thing.  It makes all the type_* methods use a more parallel method.  I think the skip_virtual is redundant since the type has to be 'real' and have a type_id.  Before, type_card didn't do the skip_modules, but probably better if it does.  Any usage that needs the mods will auto-load them at that point.

I mainly wanted to add type_cardname, and it isn't really needed that much.  I could just do type_name.cardname.

ethn:
closing: no specs!

GerryG:
How many do I need?

I added a couple to one of the pointer types that inherit from a Pointer as the default rule.  I though there was at least one test for the pointer settings, or the the pointer editor.  These cases would fail if we didn't respond to some pointer functions.

Most importantly, it isn't breaking any existing features.

GerryG:
More importantly, any comments?  It seems pretty straightforward.  Any hazards that we should be worried about.

mod/01_core/set/all/type.rb
ethn: these are slower

ethn:
This is pretty impressively tight, now that I look more closely.  should get merged in soon.

lib/card/set_pattern.rb
```diff
@@ -77,6 +77,15 @@ def initialize card
         else
           Card.fetch_id @anchor_name
         end
+        if self.class.pattern_code == 'type' and
+             default_rule = Card.fetch(card.type_cardname.trait_name(:type).trait_name(:default),
+                                       :skip_modules=>true, :skip_virtual=>true) and
+             type_code = default_rule.type_code and
+             mod_key = "Type::#{type_code.to_s.camelize}" and
+             (Card::Set.modules[:nonbase_format].values + [Card::Set.modules[:nonbase]]).
```
GerryG: Check all the possible hashes of modules, the Card and all Formats.

GerryG:
And, to repay the complement.  All made possible by how tight the module handling is now in both Card and the Formatters.  That stuff is really awesome now.  Getting that to work was really fun to do and the final product is even better.

lib/card/set_pattern.rb
```diff
@@ -77,6 +77,15 @@ def initialize card
...
+               any?{|hash| hash[mod_key]}
+          @inherited_type_module_key = mod_key
```
ethn: wait a minute.  what does this have to do with write_tmp_file?

ethn: would be nice to put this is a well-named method and make it more readable.   This is a lot to tuck inside initialization with no explanation or clean naming.

actually, hold on.  there are a couple of ways to improve this:
1.  Override initialize in Card::TypeSet (can be in this file).  then you don't have to check the pattern_code
2. there should be a method for getting the set cardname for these instances, which are basically sets.

I like this quote: https://github.com/bbatsov/ruby-style-guide#naming

ethn: you should also consider doing the initial test for a default rule by using the rule cache rather than a fetch; the rule cache would mean no new lookups (unless the rule exists)

GerryG: Both great comments.  I was thinking that I wanted to put this in the set pattern file for type, but the standard there is all class methods.  Overriding it here will work great.  And I'll have to check out how the rule_cache gets used.

lib/card/set_pattern.rb
```diff
@@ -132,4 +138,23 @@ def rule_set_key
...
+  class TypeSet < SetPattern
+    def initialize card
+      super
+      if mod_key = lookup_inherited_key(card)
+        @inherited_key = mod_key
+      end
+    end
+
+    def lookup_inherited_key card
+      default_rule = card.rule_card(:default) and
```
GerryG: I think I should test at the top for whether the card's type has a codename (or equivalent, module_key) because it will never inherit because it has a type mod bound to the codename.

lib/card/set_pattern.rb
```diff
@@ -132,4 +138,23 @@ def rule_set_key
...
+        type_code = default_rule.type_code and
+        default_rule.cardname.size > 2 and
+        default_rule.left.right.codename == self.class.pattern_code
```
GerryG: This tests that the *default rule is a *type set, but I wonder if that is what we want.  Are there any hazards of accepting any *default rule found here?  Are there any benefits?  Most cases, the *default value is saved (after editing) along with it's type.  It's that when the rule was for the type, we don't want to change the type that caused it to match the rule in the first place.

Maybe we just need more tests for any use cases that we think could be important.  If they work, great, if not we probably can fix it here.

lib/card/set_pattern.rb
```diff
@@ -130,6 +135,34 @@ def rule_set_key
...
+
+  class TypeSet < SetPattern
+    def initialize card
+      super
+      @inherit_card = card unless module_key
+    end
+
+    def lookup_inherited_key
+      return unless @inherit_card
+      card, @inherit_card = @inherit_card, nil
```
GerryG: Seemed like we wanted to keep the card out of these objects, but I was getting infinite recursion through rule_card if I did it when initializing the SetPattern instance.  We'll use it the first time this method is called where it will also clear the attribute this forgetting the card.  I guess the instance is really for that specific card.  The remnant of that will be in the return value to be stored in @inherited_key.  This could be any rule with the current code (two lines commented out would restrict it to type rules).

lib/card/set_pattern.rb
```diff
@@ -130,6 +135,34 @@ def rule_set_key
...
+        ( Card::Set.modules[:nonbase_format].values +
+          [Card::Set.modules[:nonbase]] ).any?{|hash| hash[mod_key]} and
+        mod_key
+    end
+
+    def inherited_key
+      (defined? @inherited_key) ? @inherited_key : @inherited_key = lookup_inherited_key
+    end
+
+    def lookup_module_list modules_hash
```
GerryG: This is now an override in the Type subclass, so the inherited_key stuff is only use here and only defined here.

lib/card/set_pattern.rb
```diff
@@ -130,6 +135,34 @@ def rule_set_key
...
+      return unless @inherit_card
+      card, @inherit_card = @inherit_card, nil
+
+      default_rule = card.rule_card(:default) and
+        type_code = default_rule.type_code and
+        #default_rule.cardname.size > 2 and
+        #default_rule.left.right.codename == self.class.pattern_code
+        mod_key = "Type::#{type_code.to_s.camelize}" and
+        ( Card::Set.modules[:nonbase_format].values +
+          [Card::Set.modules[:nonbase]] ).any?{|hash| hash[mod_key]} and
```
ethn: minor point, but it would be nice to stick these two lines in a separate method for readability.

GerryG: I almost did that.  One of us can add it on a small fixes kick.

PR #300, opened 2015-02-02
advanced performance logger

xithan:
log arbitrary methods using the following config options

````ruby
config.performance_logger = {
    :min_time => 100,                              # show only method calls that are slower than 100ms
    :max_depth => 3,                               # show nested method calls only up to depth 3
    :details=> true                                # show method arguments and sql
    :methods => [:event, :search, :fetch, :view],  # choose methods to log
}

# If you give :methods a hash you can log arbitrary methods. The syntax is as follows:
#  class =>  method type => method name => log options

# Example:
  Card  => {
             :instance  => [ :fetch, :search ],
             :singleton => { :fetch    => { :title => 'Card.fetch' } },
             :all       => {
                             :fetch    => {
                                            :message => 2                           # use second argument passed to fetch
                                            :details => :to_s                       # use return value of to_s in method context
                                            :title => proc { |method_context| method_context.name }
                                          },
                           },
           },
````
class, method type and log options are optional.  Default values are 'Card', ':all'  and { :title => method name, :message => first argument, :details=> remaining arguments }.

### CODE REVIEW

File: lib/wagn/log.rb [hunk 1]
```diff
@@ -162,26 +378,26 @@ def stop
...
-      def with_timer method, message, args, &block
-        if args[:context] 
-          
-          # if the previous context was created by an entry on the same level 
+
+      def with_timer method, args, &block
+        if args[:context]
+
+          # if the previous context was created by an entry on the same level
           # the finish the context if it's a different context
```
GerryG: then finish

lib/wagn/log.rb
```diff
@@ -34,24 +36,238 @@ def self.write_log_entry controller
...
+            else              value
+            end
+        end
+        #   self.class.instance_variable_set(store_name, hash )
+        #   end
+        #   log_args = self.class.instance_variable_get store_name
+          Wagn::Log::Performance.with_timer(method_name, log_args) do
+            bind_block.call(self, args, &block)
+          end
+        end
```
ethn: so this is where the magic happens?

Impressive code.

lib/wagn/migration.rb
```diff
@@ -1,34 +1,59 @@
...
+      if filename
+        self.paths.each do |path|
+          path_to_file = File.join path, 'data', filename
+          return path_to_file if File.exists? path_to_file
+        end
+      else
+        File.join self.paths.first, 'data'
+      end
+    end
+  end
```
ethn: I agree that it would be nice to be confident the migrations are all working after all these changes.  Any chance we can rerun the migrations from 1.10?

PR #301, opened 2015-02-05
make it impossible to delete accounts with edits

xithan:
got lost when we replaced the revision table with acts and actions


mod/01_core/spec/set/all/trash_spec.rb
```diff
@@ -11,5 +11,24 @@
...
+      expect(card.errors[:delete].first).to match("Edits have been made with #{name}'s user account")
+      expect(Card[name]).to be
+    end
+  end
+  
+  it 'deletes account without edits' do
+    Card::Auth.as_bot do
+      name = 'born to die'
+      card = Card.create! :name=>name, :type_code=>:user
+      card.delete
```
xithan: It fought bravely and gave its live for the dream of a better world with flawless code  ;)  https://www.youtube.com/watch?v=E_jWcIDqXq0

GerryG: This is the right PR for these tests.

Yay testss!

PR #302, opened 2015-02-05
don't make uri in email subject a link

xithan:
this was a reason why wikirate.org verifications emails were marked as spam


mod/05_email/set/type/email_template.rb
```diff
@@ -44,7 +44,7 @@ def email_config args={}
   
   [:subject, :text_message].each do |field_name|
     config[field_name] = process_email_field( field_name, args ) do |field_card|
-      field_card.contextual_content context_card, {:format=>'email_text'}, args
+      field_card.contextual_content context_card, {:format=>'email_text'}, args.merge(:content_opts=>{ :chunk_list=>:inclusion_and_link})
```
GerryG: This prevents the free link recognition?

Is this an issue for text emails?  It seems to me that our link formatting should do the correct thing with free links in a text formatter (that is render it as itself, it gets processed, but shouldn't be changed).

If you are doing html emails, that's a different story, although it still isn't clear when they wold want this difference in link handling.

xithan: Yes, for example "wagn.org" became "wagn.org[wagn.org]". Spam filters don't like that. I guess Ethan introduced this when he overwrote the link_to method (https://github.com/wagn/wagn/blame/master/mod/05_standard/set/all/links.rb).  
The format doesn't deal with this because chunk_list is only defined on sets and not on formats.
But maybe I should create a right/subject.rb and right/text_message to handle this more genereally.

GerryG: Then I would say that this is the wrong fix for that.

Ethan just fixed up what I did a bit.  I'm not sure if this was an oversight, or to deal with other problems that came up.  The code should not be passing the url as both the text and link parameter when the chunk is rendered as a link.  That way the text expander will just render the url part and not the text in brackets because it shouldn't pass that parameter.

There should be specs that fail if the text rendering includes two copies of the url.  I'm pretty sure those got added with the link_type merge.

GerryG:
Seems like links in the Subject is bad even if they are [[]] links.  Is this both for the body and header fields?

mod/05_email/set/type/email_template.rb
GerryG: Look at content_specs:
116 TEXT_RENDERED = {
117   :three => [
118     "Some URIs and Links: ", 'http://a.url.com/',
119     "\n        More urls: ",
120     "wagn.com/a/path/to.html[http://wagn.com/a/path/to.html]",
121     "\n        ",
122     "http://localhost:2020/path?cgi=foo&bar=baz",
123     "  ",
124     "extra[http://brain.org/Home]",
125     "\n        [ ",
126     "http://gerry.wagn.com/a/path",
127     " ]\n        { ",
128     "https://brain.org/more?args",
129     " }"
130   ],
131 }
You can look at the input too, there is one case where it does a free link with the [] added on, but that's because the link and the text are not the same.  The link version has http: added because it is a host without protocol url.

We need a clearer spec for this, and if there is special handling for subject, maybe it is to pass an additional option arg to tell the link formatter to just render the original text (which is what this will do, right?).

mod/01_core/spec/set/all/trash_spec.rb
```diff
@@ -11,5 +11,24 @@
...
+  end
+  
+  it 'deletes account without edits' do
+    Card::Auth.as_bot do
+      name = 'born to die'
+      card = Card.create! :name=>name, :type_code=>:user
+      card.delete
+      expect(Card[name]).not_to be
+    end
+  end
```
ethn: yay tests!

GerryG: Wait, but these are tests for the other PR, right?  I guess this branch also has the other branch in it?

mod/05_email/set/type/email_template.rb
xithan: I was imprecise. wagn.org became wagn.org\[http://wagn.org] ( not wagn.org[wagn.org] ). That's probably fine and how it is supposed to be for text format.

For emails I suggest that we process the following chunk types:
* subject: only inclusions, no url, no link,
* text body: inclusions and links (with default text format), no url
* html body: all chunks
* address: inclusions, error for url,  special format for links  [[href | text]] -> text\<href>

specs are coming soon

GerryG: That seems a bit better, but I don't think changing the chunks processed is the best way to handle this for the url case.  If the [link] syntax is what is throwing off the spam processing, then I think that should be disabled through render options rather than by changing the list of chunk types.  The point is that except for the odd bare hostname case, url processing in the text format is a noop.  Presumably the 'address' case can use a special format class where the link format is overriden?

This is really not a deep technical issue with performance implications or anything, but I think it is important enough to make a collective decision about which approach is better and to be preferred.  Making a couple of Format subclasses with the special link handling is probably the cleanest solution.  Overriding the chunk lists isn't so bad, but I don't like it as much.  I'm not liking the idea of using render args as much as I think about it, and I think Ethan may have a stronger prohibition for that option.

The content specs test file is a good place to add more cases for this.  There are example tests that use non-default formatters (otherwise it always does rich html).

lib/card/content.rb
```diff
@@ -26,7 +26,7 @@ def card
     end
     
     def chunk_list
-      @opts[:chunk_list] || card.chunk_list
+      @opts[:chunk_list] || @format.chunk_list
```
xithan: that's a notable change. I moved chunk_list from the Card class to the Format class.

mod/05_email/format/email_address_format.rb
```diff
@@ -0,0 +1,15 @@
...
+      "#{text} <#{href}>"
+    else
+      href
+    end
+  end
+
+  def chunk_list
+    :references
+  end
+end
```
xithan: I've just introduced that format  but I'm going to remove it (unless there is somebody who wants to defend it).
The only difference compared to EmailText is the link_to method and this doesn't make a lot of sense, at least not how links currently work. It was supposed to render [[my@mail.com | name]] as 'name <my@mail.com>'  but Wagn interprets  my@mail.com as a card name.  One could make [[mailto:my@mail.com | name]] work. link_to has to remove the 'mailto:' from the href variable for that. But I doubt that this is useful.  I think it's more convenient to write directly 'name <my@mail.com>'

lib/card/content.rb
GerryG: We do turn off chunk processing based on card types, and it might make sense that the kinds of chunks are more related to the card sets than the format classes.  On the other hand a format instance has a card and it's sets loaded for views.

mod/05_email/format/email_address_format.rb
GerryG: Can't we just make the external link recognizer for [[]] syntax also match without the mailto: ? It also begs the question of just what URL forms we recognize there, maybe that needs to be expanded to anything recognized as a free link.

mod/01_core/set/all/content.rb
```diff
@@ -19,8 +19,11 @@ def raw_content
   structure ? template.db_content : db_content
 end
 
-def chunk_list #override to customize by set
-  :default
+
+format :all do
+  def chunk_list #override to customize by set
+    :default
+  end
```
GerryG: Why can't this just be defined in base and inherited?

mod/01_core/set/all/trash.rb
```diff
@@ -31,9 +31,9 @@ def delete!
     errors.add :delete, "#{name} is an indestructible rule"
   end
   
-  # if account && Card::Revision.find_by_creator_id( self.id )  #ask ethan
-  #   errors.add :delete, "Edits have been made with #{name}'s user account.\n  Deleting this card would mess up our revision records."
-  # end
+  if account && Card::Act.find_by_actor_id( self.id )
+    errors.add :delete, "Edits have been made with #{name}'s user account.\nDeleting this card would mess up our history."
+  end
```
GerryG: As long as the other branch gets merged first, this shouldn't be a problem, but if that changes, we better clean this up first.

mod/01_core/set/all/content.rb
GerryG: Thinking about this further, I sort of see the need for this, but not in this case.  The set/all modules get included into the format class and will be inherited.  But for other sets where they get loaded onto the singleton class (i.e. format instances for format declarations), I'm not sure.  I'd have to look at the code some more.  It would be nice to have a test that would fail if it does require this kind of declaration.  That would be something like testing if a method or view declared on a non-all set in the base format class is available in subclasses.

In other words, I'm not completely against this method of including a format block into multiple formatters, but we should consider 1) is it needed, and 2) is there a better way to do this in the singleton module inclusion process.

GerryG: 98     def include_set_format_modules
 99       self.class.format_ancestry.reverse.each do |klass|
100         card.set_format_modules( klass ).each do |m|
101           singleton_class.send :include, m
102         end
103       end
104     end

Seems to me that this handles the case I was talking about.  The loop on 99 will load set modules from the inheritance classes with the right order so inheriting class' methods are overloaded on top to override methods from the classes up the line.  Add the tests and see if you still need this stuff.  There may be a use case when you want something in two formatters where one isn't an ancestor of the other, but I'd bet that is very rare.

mod/05_email/format/email_address_format.rb
GerryG: Another point, automatically adding mailto: or http: to the link part of a chunk reference should really be part of rendering the link and not in the chunk processing.  In a way, that is the origin of this bug.  When generating an html link, it is needed on the href= attribute, but that is format dependent and should not be in the chunk processing.

mod/01_core/set/all/content.rb
wagn: It's pretty clear to me that format(:all) is a valuable API option.  As to whether it's needed here,  I do agree that it would be nice to have a test case that fails without it.

It's pretty important to get this merged in to stop the email subject problem (though for now I changed *title on Wikirate.org to "Wikirate" from "Wikirate.org", which should buy us a little time)

mod/05_email/format/email_address_format.rb
wagn: I'm in favor of backing out of this and doing some design work on the api.

mod/01_core/set/all/content.rb
xithan: I thought there is a `chunk_list` defined for html format but there isn't. It's on the html cardtype set.
So I agree. We don't need it for chunk_list. At least not at the moment.
But I think in general it's useful. By default format beats set. Whenever you need it the other way you need `format(:all)`.
For example for css cards we don't want to render urls in any format. So we define
`Set::Type::Css::Format.chunk_list`. We are lucky because there is no `chunk_list` defined on a broader set in any format. A `Set::All::HtmlFormat.chunk_list` would mess up the html format of css cards.

@ethn: You are still acting as 'wagn'. Maybe you forgot to transform back to 'ethn' after I asked you to use your super powers ;)

GerryG: I'm not sure that we will need it, though I can see that it could be useful at some point.

I don't think the change to allow multiple formats is  very costly, and it's all at load time, so it can't hurt.

mod/01_core/spec/set/all/trash_spec.rb
GerryG: I'm expecting this to disappear when you merge latest master.

GerryG:
This can probably be merged, it does look pretty good.  We can address the stuff in the comments later.  It would be nice to know if those format :all cases are needed for this.

mod/01_core/spec/set/all/trash_spec.rb
ethn: didn't I merge it already?

xithan:
No, they are not. The format :all cases are all gone in this PR. But the support of this feature is still there.

xithan:
merge conflicts are solved

mod/05_email/set/right/bcc.rb
```diff
@@ -18,7 +22,7 @@ def process_email_addresses context_card, format_args, args
 format :email_text do
   view :email_addresses do |args|
     context = args[:context] || self
-    card.item_names.map do |item_name|
+    card.item_names(:context=>context.cardname).map do |item_name|
```
GerryG: Hmm, what does the :context arg do here?  It might make sense that it allows item_names to absolutize the names to the context, but doesn't that happen in the next line?

ethn: I looked into this.  item_names absolutizes already (unless you use :context=>:raw), so it looks like we can drop the next line.

xithan: One of the new tests failed without that. I think it was about a search card with contextual names in the query.
We need the next line for non-pointer cards: https://github.com/wagn/wagn/blob/master/mod/01_core/set/all/collection.rb#L40

ethn: interesting. Thanks for explaining. 

I'll merge this, but it seems like a wart if some types absolutize names by default and other don't.  I'll add a comment to that effect.

GerryG: Agreed.  This can be fixed so it doesn't need that line, but fine for now.  Maybe drop in a TODO comment.

ethn: I did that when I merged.  It's ok not to send PR's for comments, right? ;)

On Thu, Feb 12, 2015 at 6:52 AM, Gerry Gleason <notifications@github.com>
wrote:

> In mod/05_email/set/right/bcc.rb
> <https://github.com/wagn/wagn/pull/302#discussion_r24581360>:
>
> > @@ -18,7 +22,7 @@ def process_email_addresses context_card, format_args, args
> >  format :email_text do
> >    view :email_addresses do |args|
> >      context = args[:context] || self
> > -    card.item_names.map do |item_name|
> > +    card.item_names(:context=>context.cardname).map do |item_name|
>
> Agreed. This can be fixed so it doesn't need that line, but fine for now.
> Maybe drop in a TODO comment.
>
> —
> <https://github.com/wagn/wagn/pull/302/files#r24581360>.
>



-- 
Ethan McCutchen
One of the Wagneers, Wagn.org

Wagn. How pioneers roll.

s: ethan.mccutchen
t: @intogreater

PR #305, opened 2015-02-09
fix logger loading

xithan:
some rake tasks didn't work if logger is turned on

### DIFF
--- a/lib/card/loader.rb
+++ b/lib/card/loader.rb
@@ -10,6 +10,10 @@ class Card
         load_set_patterns
         load_formats
         load_sets
+        
+        if config.performance_logger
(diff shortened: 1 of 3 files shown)


GerryG:
I guess I'd better merge master again.  More changes to conflicting areas.  Maybe I'll wait for the queue of changes that will come before the big 1.15 split Ethan has mentioned.  The sooner we get to that point the happier.  I hate hazardous merges.

ethn:
Gerry, I did this already in the 1_15 branch.  There should be no further non-critical merges to master.

GerryG:
Thanks.  Looking good.

PR #304, opened 2015-02-09
fix migrations

xithan:
Migrations tested with Wagn v1.10.0. Seems to work now.

PR #306, opened 2015-02-09
remove 'require "jasmine" in rake file. this kills new deck's wagn seed

chuenlok:


xithan:
Noticed this too today and took care of it here: https://github.com/wagn/wagn/pull/304
Don't we need this gem anymore? I removed it only for non-developer decks.

chuenlok:
we need this gem(for core or mod-dev)  but I don't think we need it in Rakefile?

xithan:
Does `rake spec:javascript` work without it?

chuenlok:
I tried in my local and it works without `require 'jasmine'`

ethn:
if we can get things working with this elsewhere, that would be preferable, imo.  Seems out of place here.

Would it be possible to put this in a task file, for example?

ethn:
Sorry, just saw Henry's previous comment.  Obviously, if it works without it, that's even better!

PR #307, opened 2015-02-10
welcome email

xithan:
if a "welcome email" card of type "email template" exists send it when an account is activated


mod/05_email/set/type/email_template.rb
```diff
@@ -55,12 +55,11 @@ def email_config args={}
...
   
   from_name, from_email = (config[:from] =~ /(.*)\<(.*)>/) ? [$1.strip, $2] : [nil, config[:from]]
-      
   if default_from=Card::Mailer.default[:from]
     config[:from] = from_email ? "#{from_name || from_email} <#{default_from}>" : default_from
     config[:reply_to] ||= config[:from]
-  else
-    config[:from] ||= Card[Card::WagnBotID].account.email
+  elsif config[:from].empty? 
+    config[:from] = Card[Card::WagnBotID].account.email
```
ethn: do you mean blank?  I think this will break if config[:from] is nil, no?

xithan: It can't be nil because at that point it contains a rendered view. But if that changes some day we will be safer with `blank?`

GerryG:
This is what I love about this architecture.  It is really simple to add/modify this kind of thing.  Nice branch, specs and everything :)

ethn:
Here here!  Go us!

PR #308, opened 2015-02-12
Versions

ethn:
Fixed breakage on *version card.  In "getting things working" mode, so this is not a lovely solution.


ethn:
doh.  went to master.

PR #309, opened 2015-02-12
fixed *version bug

ethn:
*version card was blowing up (in browser)

Works now, and with 30 fewer lines of code ;)


card/VERSION
```diff
@@ -1 +1 @@
-1.14.8s
+1.15.pre
```
ethn: I think, like the various rails gems, we want to keep all these versions synced, so ideally we'd put them all in one place.

card/mod/05_standard/set/self/version.rb
```diff
@@ -1,8 +1,8 @@
 # -*- encoding : utf-8 -*-
-require 'card/version'
+require 'wagn/version'
 
 view :raw do |args|
-  Card::Version.release
+  Wagn::Version.release
 end
```
ethn: shouldn't be referring to Wagn from here, I know.

card/VERSION
GerryG: Sounds good, I'll move it into the base directory.  One question I have is what to we call the repo base dir for development purposes?  That's the directory that will have this file and the one you'll need to supply to configure the gem path pointers for development mode.  It's one level up from the three gem dirs.

ethn: Actually, I think it needs to stay here (I made that comment before I'd finished tidying up).  The issue is that the card gem actually needs direct access to the version from the self/version card, and that has to work in a normal deployed gem environment.  In that environment, the base directory doesn't even exist ;)

There could be other solutions for that, but let's make sure we don't break that case!

GerryG: Interesting.  Is there a spec for that?

Then we want any version reference in all three gems to reference (however indirected) card/VERSION ?

GerryG: The name of the repo base is still something we should be in sync on because it is needed in the tools. (at least for now, we can move to individual roots if we find a good way to do that at some point)

GerryG: I do really like that you removed a bunch of copied code.  Wish I'd done it in the first place.  We'll need to be sure all the wagn commands and tasks work still, some don't like references to Card::Anything.

ethn: > The name of the repo base is still something we should be in sync on because it is needed in the tools.

I don't think I follow.  What tools?  Like rake tasks?

Basically, I think it's fair to assume that core developers will be working from a local copy of the gem, so any code intended for core developers (including building the gem from the gemspec file) can safely refer to the repo base.

Any other code, however (code that does stuff for wagneers or developers) should NOT refer to the repo base, because the base usually won't be there at all.

Agree that there could be command issues, but they should be easy enough. to fix.  We could also make it so that, for example, Wagn::Version gets its methods from Card::Version one way or another.

There is a version spec, yes, and at one point it was working locally (where the repo base existed) but not on semaphore (where it didn't).  That's how I got clear on the need to have it somewhere in a gem dir ;)

ethn:
Any objections to merging?

card/VERSION
GerryG: Mostly, I want a good way to refer to it when we do.  And totally agree about where we can refer to what.

PR #310, opened 2015-02-12
fix for fetching virtual cards

xithan:
this fixes a problem that affects the upcoming following stuff.
There are cards that are defined to be virtual by code rules, e.g. 
right/followers.rb:
````ruby
def virtual?; true end
````

`Card.fetch "Home+*followers"` returns nil because the set modules are loaded after the virtual check.

### DIFF
--- a/mod/01_core/set/all/fetch.rb
+++ b/mod/01_core/set/all/fetch.rb
@@ -57,7 +57,7 @@ module ClassMethods
       else
         card.include_set_modules unless opts[:skip_modules]  # need to load modules here to call the right virtual? method 
         opts[:skip_modules] = true                           # don't load twice
-        return unless !opts[:skip_virtual] && card.virtual?
+        return if opts[:skip_virtual] || !card.virtual?
       end
       card.name = mark.to_s if mark && mark.to_s != card.name
     end
--- a/mod/01_core/set/all/fetch.rb
+++ b/mod/01_core/set/all/fetch.rb
@@ -55,6 +55,8 @@ module ClassMethods
       if opts[:new]
         return card.renew(opts) if !clean_cache_opts? opts
       else
+        card.include_set_modules unless opts[:skip_modules]  # need to load modules here to call the right virtual? method 
+        opts[:skip_modules] = true                           # don't load twice
         return unless !opts[:skip_virtual] && card.virtual?
       end
       card.name = mark.to_s if mark && mark.to_s != card.name
--- a/mod/01_core/set/all/fetch.rb
+++ b/mod/01_core/set/all/fetch.rb
@@ -54,10 +54,11 @@ module ClassMethods
     if card.new_card?
       if opts[:new]
         return card.renew(opts) if !clean_cache_opts? opts
+      elsif opts[:skip_virtual]
+        return
       else
         card.include_set_modules unless opts[:skip_modules]  # need to load modules here to call the right virtual? method 
-        opts[:skip_modules] = true                           # don't load twice
-        return if opts[:skip_virtual] || !card.virtual?
+        return unless card.virtual?
       end
       card.name = mark.to_s if mark && mark.to_s != card.name
     end

### CODE REVIEW

File: mod/01_core/set/all/fetch.rb [hunk 1]
```diff
@@ -55,7 +55,9 @@ def fetch mark, opts={}
       if opts[:new]
         return card.renew(opts) if !clean_cache_opts? opts
       else
-        return unless !opts[:skip_virtual] && card.virtual?
+        card.include_set_modules unless opts[:skip_modules]  # need to load modules here to call the right virtual? method 
+        opts[:skip_modules] = true                           # don't load twice
+        return if opts[:skip_virtual] || !card.virtual?
```
GerryG: Let me see if I get this.  Before, the card would not be virtual? here when it the fact that it was virtual? was defined in code only.

But wait, wouldn't the card 'Home+*followers' already be virtual? if there is a rule *followers+*right+*structure ?  I can see that you might put that completely in code, but would it be a solution to create this structure card anyway?  Presumably you are defining :raw and :core views for the set in code, but why not create the support to know it is virtual without the model override for virtual?

xithan: I overwrite raw_content and include Set::Type::Pointer. That's what makes the card behave as a virtual card. It would work with a dummy *followers+*right+*structure but the card doesn't need a structure.

GerryG: Right, that's what I meant.  The rule card just triggers the instantiating of a virtual? card, then it's content and type would never be used.  BTW, does a code-virtual card have a type?

Change is fine, but I don't think you need to set :skip_modules.  The loading of modules is always conditioned on them being already loaded.  There are lots of paths where it can get called multiple times, so that has to work or we'd be re-doing that step a lot.

ethn: If we're skipping virtual, we should return right away and not run #include_set_modules. 

> BTW, does a code-virtual card have a type?

Every card has a type.  These cards can be of any type.  It just means #virtual? is overridden in a set module.  Am I misunderstanding the question?

> I don't think you need to set :skip_modules. The loading of modules is always conditioned on them being already loaded.

Agree. See `@set_mods_loaded`.

mod/01_core/set/all/fetch.rb
```diff
@@ -54,8 +54,11 @@ def fetch mark, opts={}
     if card.new_card?
       if opts[:new]
         return card.renew(opts) if !clean_cache_opts? opts
+      elsif opts[:skip_virtual]
```
GerryG: Yes, this is better.  The original logic did just return in this case without testing virtual?.  We might look at the :skip_virtual vs. :skip_modules matrix.  I'm not sure we use all the case.  Do we do skip_virtual without skip_modules?  The other way around?

xithan: >The original logic did just return in this case without testing virtual?

yes, the original logic was `return unless !opts[:skip_virtual] && card.virtual?`

ethn: xithan: looks good to me!

GerryG, see fetch.rb:
````
  def [](mark)
    fetch mark, :skip_virtual=>true
  end

...
  
  def known? mark
    card = fetch mark, :skip_modules=>true
    card.present?
  end
````

GerryG: Isn't the second case one where :skip_virtual is implied?  Or is this a case that this branch is fixing?  known? is true for virtual cards, right?  Now we will include the modules anyway for virtual cards.

Yeah, these do seem like distinct cases.

Do we often do known?(name) && fetch(name) (or similar).  Seems like we should be able to get the card without another fetch when known? returns truthy.  It could just return card or nil, maybe a version without the question mark?

ethn: A card can be one of three things: unknown, virtual, or real.
known is, of course, not unknown(virtual or real).  new is not real (unknown or virtual)

[] is effectively shorthand for "fetch real" (it skips virtual).

And yes: #known? must return true if virtual and must therefore include modules for unknown cards, but at least it skips them for real cards.

As far as your last question: by default, fetch only returns known cards.  if it's unknown, it returns nil.  So you're describing the default case.

If you want a new card returned, you have to add a :new arg.  (eg `fetch mycardname, :new=>{}`). I could imagine further shorthand for that.  It does turn up with an empty hash often. I believe that as of yet there is no Card.get... Perhaps this:
````
def get mark, args={}
  fetch mark, :new=>args
end
````

GerryG: Took me a while to parse that.  You're saying that Card.fetch mark is the same as the hypothetical known I was describing?  I like the get idea.  we could probably use that for a lot of calls.

ethn: > You're saying that Card.fetch mark is the same as the hypothetical known I was describing?
exactly.

PR #311, opened 2015-02-12
Generator

GerryG:
New PR to 1_15 branch.

To summarize, this branch will be all about updating the developer command line api for the cards within an app case. I'll need your input to make sure this is right.


card/db/migrate_core_cards/20141119001955_make_symlinks_relative.rb
```diff
@@ -2,7 +2,7 @@
 
 class MakeSymlinksRelative < Card::Migration
   def up
-    Cardio.paths['files'].each do |file_path| 
+    Card.paths['files'].each do |file_path|
```
GerryG: Lot's of changes here are just this change that you requested, but card-gem was already way to unweildy.  Card.(X) are just aliases for the other, and sometime we still have to use Cardio when Card is not yet loaded (most migrations and tasks).

card/lib/card/cache.rb
```diff
@@ -25,7 +25,7 @@ class Cache
     class << self
       def [] klass
         raise "nil klass" if klass.nil?
-        cache_by_class[klass] ||= new :class=>klass, :store=>(@@using_rails_cache ? nil : Rails.cache)
+        cache_by_class[klass] ||= new :class=>klass, :store=>(@@using_rails_cache ? nil : Cardio.cache)
```
GerryG: Not sure, but you may have changed this to Rails in the merge.  Was that right?  We are not supposed to reference Rails, Wagn or Decko in Card, right?  Let me know if you have a different requirement on that.

card/lib/card/format.rb
```diff
@@ -8,20 +8,18 @@ class Format
     INCLUSION_MODES  = { :closed=>:closed, :closed_content=>:closed, :edit=>:edit,
       :layout=>:layout, :new=>:edit, :setup=>:edit, :normal=>:normal, :template=>:template } #should be set in views
     
-    cattr_accessor :ajax_call, :registered, :max_depth
+    cattr_accessor :ajax_call, :registered
     [ :perms, :denial_views, :closed_views, :error_codes, :view_tags, :aliases ].each do |acc|
       cattr_accessor acc
       self.send "#{acc}=", {}
     end
-    # TODO: these should come from Cardio.config
```
GerryG: Also threw in this TODO for free :)

card/lib/card/loader.rb
```diff
@@ -1,8 +1,18 @@
...
+  class << self
+    def config
+      Cardio.config
+    end
+
+    def paths
+      Cardio.paths
+    end
+  end
+
```
GerryG: It's a puzzle from the card-gem branch where class level Card stuff should go when it shouldn't be in card.rb.  I don't think this is bad for these two, but you can move if you want.

card/lib/card/loader.rb
```diff
@@ -121,25 +131,26 @@ def load_implicit_sets_from_source set_pattern
...
         else
-          @@rewrite = !( Rails.env.production? and Cardio.paths['tmp/set'].existent.first )
+          @@rewrite = !( Rails.env.production? and Card.paths['tmp/set'].existent.first )
         end
       end
 
       def load_dir dir
         Dir[dir].sort.each do |file|
 #          puts Benchmark.measure("from #load_dir: rd: #{file}") {
+warn "load file #{file}"
```
GerryG: Will remove, debug code.  I think I was debugging the build and it was another case of stuff not loading because seeds didn't load, so modules don't load.

card/lib/card/migration.rb
```diff
@@ -29,11 +29,11 @@ def find_unused_name base_name
     end
 
     def paths mig_type=type
-      Cardio.paths["db/migrate#{schema_suffix mig_type}"].to_a
+      Card.paths["db/migrate#{schema_suffix type}"].to_a
```
GerryG: That looks like a bug.  The merge of the last refactoring of the migrations support didn't go completely right.

card/lib/cardio.rb
```diff
@@ -11,27 +11,11 @@ module Cardio
...
-      CARD_GEM_ROOT
-    end
-
-    def add_card_paths paths, config, root
-      @@config = config
-      @@paths = paths
-      @@root = root
+  class << self
+    def card_config *args
+      @@config, @@paths, @@root, @@cache = args
```
GerryG: Most of this merged automatically when I merged 1_15, so I'm a bit puzzled about what happened here.  Maybe you could summarize what you changed in merging card-gem and I'll make sure it is reflected here.

card/lib/generators/card.rb
```diff
@@ -0,0 +1,43 @@
+# -*- encoding : utf-8 -*-
+
```
GerryG: This one is the new one to generate a migration that adds cards to your app.  I'm not sure if this runs the way it should.  Ideally, this is an add-on to rails generate (rails generate card [args]) when the card gem is installed.  The rails app should need decko-rails and card, but not wagn.  We'll need to address how assets will work in this case so that the developer can integrate assets from these two gems with their asset frameworks.

card/lib/generators/card/card_generator.rb
```diff
@@ -0,0 +1,33 @@
+# -*- encoding : utf-8 -*-
+
```
GerryG: This one is the new one to generate a migration that adds cards to your app. I'm not sure if this runs the way it should. Ideally, this is an add-on to rails generate (rails generate card [args]) when the card gem is installed. The rails app should need decko-rails and card, but not wagn. We'll need to address how assets will work in this case so that the developer can integrate assets from these two gems with their asset frameworks.

card/lib/generators/card/format/USAGE
```diff
@@ -0,0 +1,9 @@
+Description:
```
GerryG: This is just the original generators moving into a place where they show up in the help lists, etc.  Try out the generator help subcommands on the branch.

card/lib/generators/card/templates/create_cards_tables.erb
```diff
@@ -0,0 +1,106 @@
+# encoding: UTF-8
```
GerryG: This is instead of referencing a schema.rb because the app will already have its own.  So we add our tables as a migration.

card/lib/generators/card/templates/create_cards_tables.erb
```diff
@@ -0,0 +1,106 @@
...
+    end
+
+    add_index "<%= prefix %>schema_migrations_core_cards", ["version"], :name => "<%= prefix %>unique_schema_migrations_cards", :unique => true
+
+    create_table "schema_migrations_deck_cards", :id => false, :force => true do |t|
+      t.string "version", :null => false
+    end
+
+    add_index "<%= prefix %>schema_migrations_deck_cards", ["version"], :name => "<%= prefix %>unique_schema_migrations_deck_cards", :unique => true
+  end
```
GerryG: To be resolved: loading seed/fixture data into these tables and running migrations from gems (particularly when updating gems to a newer version).

I did split the rake tasks so that loading the seed data is a distinct task from loading the schema, which we should prevent when we are embedded.

decko-rails/config/routes.rb
```diff
@@ -1,11 +1,7 @@
 # -*- encoding : utf-8 -*-
 
-Rails.application.routes.draw do
+Decko::Engine.routes.draw do
```
GerryG: This one looks suspicious.  It may have been in card-gem, but you wanted to put it off to this branch maybe?

decko-rails/lib/decko/engine.rb
```diff
@@ -46,9 +46,16 @@ class Decko::Engine < Rails::Engine
...
-        Cardio.add_card_paths Decko::Engine.paths, Rails.application.config, Wagn.root
+        Cardio.card_config Rails.application.config, Decko::Engine.paths, Wagn.root, Rails.cache
         ActiveRecord::Base.establish_connection(Rails.env)
       end
+      ActiveSupport.on_load(:after_initialize) do
+        begin
+          require_dependency 'card' unless defined?(Card)
+        rescue ActiveRecord::StatementInvalid => e
+          Rails.logger.warn "database not available[#{Rails.env}] #{e}"
+        end
```
GerryG: I think this may have been necessary for one of the migration or date loading cases, not sure.

decko-rails/spec/controllers/card_controller_spec.rb
```diff
@@ -4,8 +4,13 @@
 
   include Capybara::DSL
   describe "- route generation" do
+    routes { Decko::Engine.routes }
 
     it "should recognize type" do
+      all_routes = Rails.application.routes.routes
+
+   require 'rails/application/route_inspector'
```
GerryG: Magic so tests know about engine routes.  Probably why the Decko::Engine declaration isn't in the card-gem branch.  Breaks tests without this in the tests.

decko-rails/spec/controllers/card_controller_spec.rb
```diff
@@ -48,6 +53,8 @@
   end
 
   describe "#create" do
+    routes { Decko::Engine.routes }
```
GerryG: Magic so tests know about engine routes. Probably why the Decko::Engine declaration isn't in the card-gem branch. Breaks tests without this in the tests.

wagn/lib/wagn/generators/wagn/templates/config/routes.erb
```diff
@@ -0,0 +1,12 @@
+# -*- encoding : utf-8 -*-
```
GerryG: Hmm, not sure this should be here now, maybe go duplicated in the merge.

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -13,7 +13,9 @@ end
 
 namespace :wagn do
   desc "create a wagn database from scratch"
-  task :create do
+  task :recreate do
```
GerryG: Point on task names.  We need something that means 'clear and reload the data, but don't drop and add the schema', vs. what create means originally.  In the embedded case, the equivalent of reloading schema will be to revert the migration that added cards, and then run a migration (possibly different) that add the current cards schema.  Maybe we need Card.embedded? or similar so code/config/tasks can know which it is?

GerryG:
Original reference PR here: https://github.com/GerryG/wagn/pull/43/files

Most of the comments are covered with new comments here, and the card-gem branch that it targets is merged into 1_15, so this PR won't change, and get more confusing as more things are merged from 1_15 and not into card-gem.

GerryG:
At least it passes.  The build on my semaphore account is broken.  I could use access to yours so I can at least look at it to copy stuff.

card/lib/card/cache.rb
```diff
@@ -16,7 +16,7 @@ def real_file_path(name)
   class Cache
 
     @@prepopulating     = [ 'test','cucumber' ].include? Rails.env
-    @@using_rails_cache = Rails.env =~ /^cucumber|test$/
+    @@using_rails_cache = Rails.env =~ /^cucumber|test$/ # why two versions of same test?
```
ethn: I think the separate class vars may be defensible (because we have played with different configurations), but the two tests are pretty ugly ;)

card/lib/card/cache.rb
ethn: I changed it to rails because what was there before broke in the browser.  If this works, it's better.

Even with our tests, it's probably worth a browser sanity check for any significant amount of code change.  (Especially something like this, which explicitly does not get run in test environments!)

card/lib/cardio.rb
ethn: uhhh. this file was new in card-gem, so I doubt merging master and card-gem changed this?

card/lib/card/migration.rb
ethn: our standard way to test migrations is to run all migrations from 1.10.  See data dump in wagn/test.

Migration debugging process is roughly:
1. import sql to db
2. check out 1.10
3. config to use db
4. run `wagn update`
5. fix bugs
6. repeat

card/lib/card/cache.rb
GerryG: Exactly.  Fine to want to have two things that happen to use the same test.

decko-rails/lib/decko/engine.rb
GerryG: This could also be something that fixes the console error.  I will check and let you know.

card/lib/cardio.rb
GerryG: Good point, unless there is a move history that caused a change.

card/lib/card/cache.rb
GerryG: It is supposed to work.  If it doesn't I'll have to fix it.  It may work correctly this way on this branch.

decko-rails/lib/decko/engine.rb
GerryG: I verified this (wagn console works on 1_15 plus this addition.  Let me now what you want to do.  You could just add it to 1_15, or I can make a branch for this and maybe a few more critical details.

ethn: I wish there were an easy way to see the updated diffs (now that stuff is
merged in). sigh.

at least we're in the hundreds now (diff lines) and not tens of thousands ;)

On Thu, Feb 12, 2015 at 5:57 PM, Gerry Gleason <notifications@github.com>
wrote:

> In decko-rails/lib/decko/engine.rb
> <https://github.com/wagn/wagn/pull/311#discussion_r24636438>:
>
> >          ActiveRecord::Base.establish_connection(Rails.env)
> >        end
> > +      ActiveSupport.on_load(:after_initialize) do
> > +        begin
> > +          require_dependency 'card' unless defined?(Card)
> > +        rescue ActiveRecord::StatementInvalid => e
> > +          Rails.logger.warn "database not available[#{Rails.env}] #{e}"
> > +        end
>
> I verified this (wagn console works on 1_15 plus this addition. Let me now
> what you want to do. You could just add it to 1_15, or I can make a branch
> for this and maybe a few more critical details.
>
> —
> <https://github.com/wagn/wagn/pull/311/files#r24636438>.
>



-- 
Ethan McCutchen
One of the Wagneers, Wagn.org

Wagn. How pioneers roll.

s: ethan.mccutchen
t: @intogreater

GerryG:
https://github.com/GerryG/wagn/pull/44  I thought I already made that one for comparison on my part.

GerryG:
This doesn't seem to be updating now.  I don't see why not.

ethn:
We keep seeing that PR's don't update when their target updates. Definitely a bummer.

I think there might be some situations where it does update (eg, new commits to target are included in PR, maybe?).   Not sure.

GerryG:
The last commit is pretty key.  I think I have the new generator concept working.  I will create some more scriptish tests for some use cases we'll want.

card/lib/generators/card.rb
```diff
@@ -0,0 +1,43 @@
+# -*- encoding : utf-8 -*-
```
GerryG: This file just defines a couple of classes that are used by the actual generators.  The main reason *rails g* wasn't working quite right was I wasn't including this file in each generator.

GerryG:
It is actually mostly working.  The generator mostly work, there is some cleanup to do, but there is a problem to solve about migrations.  It seem like the paths entry for deck_cards isn't getting populated as it should be.

For the embedded case, I'm able to run some generators under *rails*, but the migrations don't work because the wagn application stuff isn't loading.  Some of the stuff that is in wagn/application.rb probably needs to be in decko-rails instead.

card/lib/generators/card/format/USAGE
ethn: did you delete and add in separate commits?  looks like git didn't pick this up as a rename.

wagn/lib/wagn/tasks/wagn.rake
ethn: Dunno what "embedded" means.  Like rails decks?

Does anyone actually need to do any of this stuff?  Dropping and reloading the schema seems to be a simple way to avoid schema issues in this case.  When is it important not to do that?

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -13,7 +13,9 @@ end
...
+    ENV['SCHEMA'] ||= "#{Cardio.gem_root}/db/schema.rb"
+
     puts "dropping"
     #fixme - this should be an option, but should not happen on standard creates!
     begin
```
ethn: if it's there, it should drop it, no?

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -24,16 +26,40 @@ namespace :wagn do
...
+    conn = ActiveRecord::Base.connection
+
+    puts "delete all data in bootstrap tables"
+    WAGN_BOOTSTRAP_TABLES.each do |table|
+      conn.delete "delete from #{table}"
+    end
+  end
+
+  desc "Load bootstrap data into database"
+  task :seed do
```
ethn: from the command line `wagn seed` calls `rake wagn:create`.  This is confusing!  

Since `wagn seed` is the most public part of the api (see https://github.com/wagn/wagn/blob/master/README.rdoc#3-create--seed-database), perhaps we could make it so that "seed" means "create and pre-populate".

In other words, `wagn seed` would call `rake wagn:seed` which would call the :create and :prepopulate tasks.

wagn/lib/wagn/tasks/wagn.rake
GerryG: I think we're ok on terminology.  Embedded means in another rails app, or what you are suggesting is a *deck app*.  Then you started calling the standard case, *app deck* which is hard for a dyslexic like me.

I guess I wasn't thinking about it quite right.  I was thinking because you create the card dependency in a migration, that reloading the schema is out, but the migration expands the schema and so it can load the schema that will have all the apps other tables and the card tables.

I still need a task for loading the data.  I want to put it in the decko namespace (vs. wagn tasks where we already have them).  Maybe initially we have the same tasks in wagn and decko.  Ideally, we won't actually depend on wagn in the deck app case.

wagn/lib/wagn/tasks/wagn.rake
GerryG: I thought (not totally sensibly) that adding cards to an existing app would mean that often we would not want to *start fresh by dropping and re-loading the schema*, but as I commented above that isn't quite right.

I was playing with adding couple of new subcommands and tasks.  The standard way, which is just like it does now, drop and reload, then seed.  The other way is clear data from all the card tables and then seed.

Let's see:
LOAD_TASKS= {
  'reseed' => 'create',
  'seed' => 'recreate',
  'load' => 'seed'
}
Sorry for that, it isn't pretty, but it will be ok when we finalize this stuff.  It maps the wagn subcommands to wagn: namespace tasks.  So, seed => recreate, which is what the old create did (that's part of what I'm apologizing for, it's not right, but I think I flipped them when I realized we didn't want the external commands changing (wagn X) because the are used in the build).

To me, recreate is what we do when we drop, reload schema and seed, and create should just create (isn't that would db:create does ?).  It is *wagn load* to just run rake wagn:seed.

wagn/lib/wagn/tasks/wagn.rake
ethn: fwiw, I never used the term "deck app" for anything.

A Decko installation is a *deck*.  I suggested you can have a *rails deck* or an *app deck*.  Perhaps it would be less confusing to say *engine deck* vs *app deck*?  And for now you mentioned using *free* deck to mean something that is neither. So we could add:
````
#app_deck? 
#engine_deck?
#free_deck?
````
We know that technically, a rails app is also an engine, but an app deck is not an engine deck, ok?

If you're working with an *engine deck*, you should be installing decko-rails but the wagn/decko gem should not even be present, correct?

None of this answers my question above.  How the heck is this currently passing semaphore??

File: wagn/lib/wagn/tasks/wagn.rake (hunk 22 above)
Comment (ethn) [inline wagn/lib/wagn/tasks/wagn.rake]: YIKES!  If we're talking about CLI to wagn task, it should be direct mapping, eg:

seed -> seed
reseed -> reseed
load -> load....

AFAIC, you can rename rake tasks with abandon, but please don't change the basic installation process for app decks.

But first, can you please explain how this passes Semaphore?  How does the database ever get created??

Comment (ethn) [inline wagn/lib/wagn/tasks/wagn.rake]: ohhh. I see now. sorry.  I was being thick.

I get why it works, but let's please make it so that there is just one meaning.  The wagn meanings don't have to map directly to the rails meanings, but the CLI ones and the rake tasks need to line up.

File: card/lib/generators/card/USAGE [hunk 23]
```diff
@@ -0,0 +1,21 @@
+Description
+
+Examples
+  wagn generate card new cards pre_fix
```
ethn: card, new, cards, and pre_fix are all variables?  Can you choose some more plausible variable names?

card/lib/generators/card/USAGE
```diff
@@ -0,0 +1,21 @@
...
+
+Creates:
+  config/new_routes.rb
+  db/migration/2015010112000000_create_cards_tables.rb
+
+
+  The table names with prefix 'pre_fix' (i.e. pre_fixcards, pre_fixcard_references ...)
+  The Decko::Engine routes will be mounted at /cards/
```
ethn: are you sure we want to do this prefixing rather than adding a deck id?

for now I'd be comfortable with a "one deck per rails app" restriction".

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -24,16 +26,40 @@ namespace :wagn do
...
+
+    Rake::Task['wagn:seed'].invoke
+  end
+
+  desc "clear and load fixtures with existing tables"
+  task :create do
+    ENV['SCHEMA'] ||= "#{Cardio.gem_root}/db/schema.rb"
+
+    Rake::Task['wagn:clear'].invoke
+
```
ethn: So "create" doesn't actually create a database?

card/lib/generators/card/templates/create_cards_tables.erb
ethn: This is an odd use of a generator, isn't it?

Also obviously very non-DRY.  Should find a way to use the schema.rb.

As mentioned elsewhere, I don't really like the prefix idea.  It would cause a massive headache.

Really, before you worry too much about automating the setup of an engine deck, I would just experiment with setting one up.

I would expect, for example, that creating a new engine class.

For app decks you have a MyApp::Application class that inherits from Wagn::Application (soon Decko::Application) which in turn inherits from Rails::Application.

I would think that when you create an Engine you would get a MyDeck::Engine class that inherits from Decko::Engine (which inherits from Rails::Engine).  Have you experimented with that?  Do we have a hello world of a deck within a rails app?

wagn/lib/wagn/commands.rb
```diff
@@ -36,19 +36,27 @@ def find_spec_file filename, base_dir
...
+  require 'generators/card' if command == 'generate' # without this, the card generators don't list with: wagn g --help
   require 'rails/commands'
 else
   command = ARGV.shift
   command = ALIAS[command] || command
 
   case command
-  when 'seed'
+  when 'seed', 'reseed', 'load'
+    task = LOAD_TASKS[command]
```
ethn: kept not seeing this line :)

card/lib/generators/card/USAGE
ethn: Do I understand that this is supposed to be the main mechanism for creating a new engine deck?

I would think the process for setting up a new deck would be something more like:
1. add 'decko-rails' to Gemfile of rails app
2. run bundle install
3. run `decko-rails new wiki`.  

The last one would run a rake task to set up db tables, generate the necessary stuff, and probably output some instructions (add X to your config/routes.rb)

The big question is: what stuff do we need that generator to create?  Here's a starting place: http://guides.rubyonrails.org/engines.html#generating-an-engine.

Those generators make lots of things we know we don't want: app directories and helpers and view and so forth.  But I do suspect we want a new engine class at a minimum and perhaps more.

Step 1 is getting all that to work.  Step 2 is automating its generation :)

ethn: the API is debatable, of course.  But I think the idea is that decko-rails does not require wagn, correct?  So it seems like `wagn` shouldn't be the base call.  Other ideas:
`card new --engine wiki`
`decko-rails mount wiki`
...

card/lib/generators/card/format/USAGE
GerryG: I didn't think so, but maybe things got messed up from a merge or something.  It might not be a bad idea to re-create this branch fresh and see if that can be fixed.

card/lib/generators/card/templates/create_cards_tables.erb
GerryG: If you want to re-use one source of the schema, I think schema.rb should be derived from a migration like this, not the other way around.  After all, in general an app starts without a schema and migrations build it up.

I'm not totally attached to the prefix concept, and in general maybe we encourage people not to use it, but it seems a necessary option for deck apps.  What if they already have a table named 'card', for example?  We probably don't need to support for the app deck case, and no doubt there are a lot of fixes needed to actually support it.

Is there any reason for adding a MyDeck::Engine?  You could create one, but in general I don't think it is needed.  Is there something wrong with doing Decko::Engine.config.foo_bar = 'baz' ?

I'm just at the point of trying to test generating a new rails app, adding the card engine and trying to run it.  I was planning on adding a script like the one for the migration test as a developer guideline.

Don't you agree that

wagn/lib/wagn/tasks/wagn.rake
GerryG: I think you used *deck apps* in a more recent comment where it seemed to be in contrast to an *app deck*, but I think you are right *engine deck* maybe better (or *rails deck* which could be considered an alternate for *engine deck*).  Those seem fine for the boolean methods.

My thinking was that eventually, there would be just the decko and card gems.  The wagn gem is already getting pretty empty.  In my view, we only split them (wagn and decko-rails) as a transitional thing because we aren't ready to re-brand the whole thing yet.  So, we may find a good reason to keep this split, but I think the final state will be that you install the decko gem, then you can do either 1) run decko commands to generate an app deck, or 2) add decko to your Gemfile and run rails generator commands to add the models and engine to your app.

wagn/lib/wagn/tasks/wagn.rake
GerryG: This is where I wanted to change the old 'create' task to be called 'recreate' because that better describes what it does, but I had to change it back so create is the original drop, load, re-populate task.  This is the wrong name for the new task which clears tables and re-populates.  I think that is a valid task to want, but the names are not right yet.

wagn/lib/wagn/tasks/wagn.rake
GerryG: Agreed, that's why I'm apologizing.  It's an ugly shim right now, but I'm confident it won't be hard to clean up once we are clear and the commands we need and want.

card/lib/generators/card/USAGE
GerryG: Ideally, yes.  The main reason I still need wagn for the engine deck case is because I want the rake tasks.  Should we copy the rake tasks into the new namespace and have two for a while?  Start using decko:task now?  I don't think we need the wagn command for this case as long as the tasks are available.

One of the important things on this branch is getting the generators to load in rails as they should.

card/lib/generators/card/USAGE
GerryG: One set of card tables, but that could be multiple decks when we have namespaces implemented.

Are you dead-set against prefix?  There is certainly some work to do for it to be even usable, but won't we be asked for it in the engine deck cases?  I think we have to have a pretty strong argument for not supporting it in the long run.  I'm fine with holding off on including even basic support for now.

wagn/lib/wagn/tasks/wagn.rake
GerryG: That isn't right as a name.  This is more like 'reseed' as a better descriptive name.  I can change it to that and then make create what it used to be again (even though it does more than rails db:create)

wagn/lib/wagn/commands.rb
GerryG: Yeah, kind of critical to know what tasks run when.  And I agree once we settle the API the mapping shouldn't be needed.

card/lib/generators/card/USAGE
GerryG: I just added a commit to this branch that improves things.  This is the basic scenario as I see it:

1. You have a rails app (generate a skeleton app for testing).
2. Add decko-rails (later just decko) to the Gemfile
3. Generate a migration to add cards to my app: rails generate card [optional args, defaults are just like a deck app]
4. Follow the instructions output from step 3.
5. Edit config/routes.rb, config/application.rb and Rakefile
6. Run migrations: rake db:migrate
7. Load card seeds: rake decko:seed
8. Run tests: rake decko:rspec (this doesn't exist yet, but I think this is what we want for starters)

We could make it so wagn (later decko) command will run the tests for either case, but right now decko doesn't exist and wagn rspec won't run in an engine app.  We could fix it so it does and eventually port that to decko when we factor out wagn.

wagn/lib/wagn/tasks/wagn.rake
ethn: Fwiw, you did propose *deck apps* here: https://github.com/wagn/wagn/pull/319 but that was for the same meaning as *app deck*. I strongly prefer the latter, but the good news is that even if we confuse the order, the meaning should be clear :)  Afaik, nobody wants to use one of those to mean engine deck.

I think, long term, we still want three gems, but they are not currently organized as I think they should be.

As you say, there are (and should be) two main gems: decko and cards.  But I do think decko-rails should continue to exist, only in a different form.  It's decko-rails (not decko) that should be super thin.  What I originally pushing for (and still want to see) is that decko should not depend on decko-rails.  At present we've gone the opposite route.

I get that conceptually an engine is lower level than an application.  But organizing our gems around this idea assumes that we're going to retain a railsy architecture for Decko.  That's not my assumption.  Wagn::Application may some day not inherit from Rails::Application just as Card may not inherit from ActiveRecord::Base.

Just for the hell of it, imagine that two years from now, Wagn has completely moved away from Rails.  We should still be able to plug it into rails with a rails engine.  It's an external application using our API.  Not thinking of it that way gets rails stuff baked too deeply into our architecture.

It will really help if you think of things this way.  Wagn is a pioneering new development framework supporting a MoFoS architecture.  It appreciates the Rails community, it owes a great debt to the Rails community, it hopes to contribute back to the Rails community, and it still makes use of Rails code.  But we are not committing to supporting the Rails API; we're developing the Wagn API.

What I *still* want is this:
- Card stands alone (currently failing here because of the way configuration works.  We couldn't really load paths without an external system calling Cardio#set_config with some rather involved objects from Railites)
- Decko depends on Cards and turns it into an app.  Decko does NOT depend on decko-rails, because, conceptually, why the hell would it?  This is the default case.
- decko-rails is a (thin) wrapper that makes it so that decko can plug into rails.

We're really not that terribly far away from this.  But public, spec, lib, config.... all those things should be in decko and referred to by decko-rails, not the other way around.  

I doubt you'll want to take this on, which is fine; I'm willing to.  I don't think it's going to be that hard.  But I don't think the decko-rails case should be in decko, both for functional and marketing reasons.

card/lib/generators/card/templates/create_cards_tables.erb
ethn: Looks like this thought got cut off :)

> What if they already have a table named 'card', for example?

Feels like you're getting lost in bogus edge cases. I doubt this will come up, and when it does, the namespace issues are coing to be much bigger than table names.

We need to get an engine working today or tomorrow and not keep wandering in the wilderness. This prefix stuff is wilderness.

> Is there any reason for adding a MyDeck::Engine?

Yes.
1. If we keep the Engine configuration closely parallel to the Application configuration, we'll make it easy for someone to go from a rails deck to and app deck; this is a case we want to support.
2. The boot process is set up so that many things happen when the engine definition is complete.  Assigning configuration values afterwards means they will get ignored by anything that happens in the mean time.
3. I think there's actually a significant error in setting the engine root equal to Rails.root.  Typically, I think the engine root is elsewhere, and this is determined by where the Engine itself is defined.  If it's defined in the gem, the engine root will by default be in the gem; that's not what we want.
4. If you don't create a new engine class, you have to create a whole new pattern for how everything works.  Unlike with code inside our gems, in code in other people's rails apps we definitely want to follow railsy patterns as closely as makes sense for Wagn.

card/lib/generators/card/USAGE
ethn: > Should we copy the rake tasks into the new namespace and have two for a while?

Lord no!  Stay thirsty and DRY, my friend. The tasks should be defined in wagn / card and loaded into decko-rails as needed.

> later just decko
imo, always decko-rails.  we don't want to imply that this usage is what *decko* is. Nor do we want decko to be burdened with supporting a special plug-in case.

> Generate a migration to add cards to my app

I don't think we want to mess with their migrations; let's just use our own.   We're not going to add migrations to their app every time we update our structure.  We already have 2 custom migrations tables in our app case (card and deck); in the engine case we should probably have 3, to include the card_structure.

> Follow the instructions output from step 3.
> Edit config/routes.rb, config/application.rb and Rakefile

are these two the same?  I would think that only config/routes.rb would be necessary.  There's no mention of either of these two in the engine guide.

> Run migrations: rake db:migrate
> Load card seeds: rake decko:seed

So many commands.  We're going to be exhausting folks.

> Run tests: rake decko:rspec

You're not saying that running tests is part of standard installation are you?  I don't think I've ever done that with any software before. This would mean they'd have to install many more dependencies, and would dramatically up the odds of problems.

How's this:
1. add 'decko-rails' to Gemfile of existing rails app
2. `bundle install`
3. `decko-rails new wiki`.
4. follow output instructions to edit config
5. `decko-rails seed`

This mirrors our normal installation process very closely; in fact, you can drop the "-rails" and it's basically the app deck commands.

GerryG: I'm not sure you quite understand that there really isn't much to decko-rails.  Or rather, there isn't much in wagn -> decko, so I'm not sure I understand what you want here.

I don't see the problem with implementing the 2nd case as an extension that you don't really see unless you use it a certain way.  Both cases are a rails app, but in one you don't open up the app at all.  Extensions are modules within the app.

I think I'm finally seeing what you mean about just having our migrations.  The generated migration doesn't have to change, or rather, it only changes when the schema changes.  In fact, if we just load the schema by running the *initial* migration into a database with no tables instead of the traditional rails schema load, that solves the DRY issue with the base schema.  It's just a migration that creates an empty deck.

I don't think we need a decko-rails command.  That still should be a generator (I guess you want it to be called wiki (rails g wiki [card])), but it doesn't have much to do unless you make it actually edit (and backup) the files it needs to config.  I'm trying to reduce the number of files to a minimum.  I will also try to get it to attach our migrations to the base tasks (db:migrate).

I never wanted to duplicate things, but I also don't really want to reference wagn tasks from decko-rails.  Can we move them to the decko now? (or will it be deck for rake tasks?)  So decko:migrate and sub-tasks should be the same for both cases (aka wagn:migrate).

ethn: > I'm not sure you quite understand that there really isn't much to decko-rails

I'm very clear on this and have been saying this all along.  Very little. It should be a very lightweight gem.  You've put a lot more in there than I would want there.  NO core decko code should be there.

> Both cases are a rails app, but in one you don't open up the app at all. Extensions are modules within the app.

Wagn happens to be a rails app now.  For the purposes of this architectural decision, it's better to pretend it's not.

ethn: > I don't think we need a decko-rails command

We don't need it, but I think we want one.  It teaches the decko way.  And for decko experts, you don't really have to learn anything new.  I would personally have a hard time trying to learn the commands you're proposing.

> I'm trying to reduce the number of files to a minimum

I want to keep those down, too, but it's more important to have an awesome user interaction. You can start with generators if you want, but when we're ready to promote decko-rails, we'll want something much better than than `rails generate card` which is totally baffling.  Rails isn't generating anything; decko-rails is.  And what's being generated is not a card. (rails g migration creates a migration, rails g engine creates an engine.  controller, model, view.... this would be a weird generator that doesn't generate what it says it does)  

> I never wanted to duplicate things, but I also don't really want to reference wagn tasks from decko-rails. Can we move them to the decko now? (or will it be deck for rake tasks?) So decko:migrate and sub-tasks should be the same for both cases (aka wagn:migrate).

To me it's obvious that decko-rails should depend upon decko (currently wagn).   And decko/wagn should not depend upon decko-rails. I really don't get why this is in question.  We want people to find decko via decko-rails and not the other way around.   decko-rails a plugin for putting decko in your rails app.  It's just so clear that it's a separate gem.

As for how the commands work, the plan is:

The tasks will all subsitute wagn with decko:
`wagn new` = create app deck now
`decko new` = create app deck future

What I've been proposing is this:
`decko-rails new` = create engine deck soon

We could do it other ways, like
`wagn new` (soon `decko new`) is smart and recognizes when it's in a non-wagn rails app.  But this is the kind of logic I don't think belongs in the wagn gem.  

~~~
most other CLI options will translate directly to rake tasks
`wagn seed`  = `rake wagn:seed` = create and populate app deck database now/soon
`decko seed` = `rake decko:seed` = create and populate app deck database future

Again I'm proposing that we follow something like this pattern with:
`decko-rails seed`, but  I don't have any deep proposals for the namespacing of the rake tasks triggered.

> Can we move them to the decko now? (or will it be deck for rake tasks?) So decko:migrate and sub-tasks should be the same for both cases (aka wagn:migrate).

I'm definitely fine with having both namespaces handled in the decko-rails gem.

To me, it seems like the big next step is getting it so that wagn no longer depends on decko-rails.   

I keep saying this, but why don't you try hacking together an engine?  (don't worry about generators or special CLI -- just get it to work) It doesn't sound like we've even been playing with engines at all, and at this point that seems overdue, no?

GerryG: This is really the wrong place for this discussion.  I think we need to talk so we are sure we are really on the same page.

What frustrates me is the only reason there is a decko-rails is that we are not ready to change the branding and go away from wagn.  Lets imaging a future in which there is also a decko-sinatra or decko-pick_some_framework, and in that context, generating an app deck (what wagn new does now) would be just one of the frameworks it generates for.  So, decko new sinatra or decko new my_framework would implicate decko-sinatra or decko-my_framework.  Maybe plain 'decko new' implied decko-rails by default, or maybe it is included in decko.

Any way you slice that, it is wrong to move the basic engine support from decko-rails back to wagn (or decko).  The decko-rails gem would need to have all the engine stuff.  Engines are a rails concept, so you wouldn't have it for decko-sinatra, right?  I guess that things like assets are ideally shared, with adapting frameworks going into deck-platform gems, but the java/coffeescript, images and such should be in a shared gem (wagn/decko).  On the other hand, I think it is ok to combine decko and decko-rails and have the other glue modules adapt the shared parts from that combined gem.

To me, decko-rails is a shim we need to not have to re-brand all at once.  To you it is the support for this use case that you don't want to get in the way of "normal" wagn functioning.  I'm supportive of the idea of not interfering with the common (now only) use case.  I'm even supportive of the idea of making the usage of this new case (engine decks) smoother from a UI standpoint, but it is pre-mature to optimize for that now.

Note that I agreed with your point about not generating a migration in 'rails g card', that file should be included in the card gem.  Did you get my point about using that file in loading the schema in the standard flow?  Rather than compressing all the migrations up to a point into a schema.rb file, they get compressed into a single migration that takes the db structure from empty to the current structure.  It still needs to be set apart from the rest of the migrations as it may take you past some of the other migrations still needed by decks in the field.

I'm trying to get feedback on the CLI, and I do appreciate that.  I'm also trying to improve the generators as they are.  We should not need the wagn command to access the card generators.  Those are rails things and should extend the rails framework.  I'm more concerned about this sort of question about where they belong code-wise.  That's why I'm objecting so strongly to your desire to move code back into wagn.

Note that given your stated goals here, it is actually wrong for wagn command to build off of the rails command stuff.  Well, it's fine for wagn, but not what we would want for decko.  Maybe for decko-rails if it implemented a command (decko most naturally and a conflict).

Yes, we want to make the commands as natural as possible, but no, we don't care that much about the details of what command you run for which case.  We can create wrapper scripts for that when we know what the use cases are.  The whole reason I created these PRs way before they were ready is so I could follow your feedback as I went along.  Now, I'm not feeling particularly acknowledged for the work done.

I'm not that worried about getting an engine to work in a simple application started from scratch.  I'm concerned about my use case of an enormous rails app that I wan't to add cards to.  If I try to present this as a solution to my colleagues when the basic plumbing isn't right, I'll be sunk from the start.  They won't care if the CLI is natural, but about whether the steps work and don't break things.

I'm working on the basic functioning of these work flows because the first accomplishment I want is to add cards to an empty rails app and basically get a functioning wagn.  Then I'll be ready to try to add cards to an existing app.  I'll want namespaces before I can add it to our complex app.

ethn: Hey man,

A lot here to respond to, but the main thing I want to say is that I think we're very close to a great solution, and it's all thanks to you.   I just moved some files around, but we're retaining all your major restructuring.

> I'm not that worried about getting an engine to work in a simple application started from scratch. I'm concerned about my use case of an enormous rails app that I wan't to add cards to. 

I know we can do it, but I really want to build one so we can agree upon the endpoint.  Imo, it's weird that we're getting stuck generators; we're debating the mold when we haven't made the statue.  On that level, it seems like the CLI is the *only* part we're ready to clarify.

> If I try to present this as a solution to my colleagues when the basic plumbing isn't right, I'll be sunk from the start. They won't care if the CLI is natural, but about whether the steps work and don't break things.

I hear you.  I don't think I really changed the plumbing that much other than gem organization, right?  As mentioned over there, that's the part that's important to me.  We can now do a LOT of wagn -> decko renaming within the wagn gem.  Then, when it's time to pull the switch, the gem rename should be a snap.

> To me, decko-rails is a shim we need to not have to re-brand all at once.

That's completely different from my conception. The way you described it in the other gem was much better aligned.  Wagn/Decko is everything you need for an app deck.  Decko-Rails is a (very thin) layer wrapped around that with support for making it so that it's easy to plug decks into rails apps.  As you say, there currently is almost nothing to that.

As I understand things, decko-sinatra would be used to plug decko into a sinatra app.  It would not change the internal workings of decko (which, for now at least, would still be based on rails).  If you wanted to build a decko-like app using sinatra for the innards, that would be a different application.

This is what I'm trying to emphasis.  Decko is decko.  There's only one of them.  It may be able to plug into multiple frameworks, but that plugging-in won't hollow out its innards.  The fact that it is currently rails makes it very easy to make it rails compatible.  But it should never depend on decko-rails, which is the mechanism for deploying it as an engine within a host app.

GerryG:
I think we are on the same page here.  The generator issue is really about making decko(wagn) generators rails friendly.  That actually isn't that hard and is mostly solved.  The 'add your engine model to my app' doesn't have to be solved completely, but I have to be able to do it (and script a test for it).

The issues with rake tasks is bigger.  I don't care if they are still in the wagn namespace, but they have to load and work in the engine deck context.  I think you are ok will changing the task namespace sooner rather that later, but in truth, there is no rush on it.  I do want to be able to run tests in order to test the whole process, but you are right the gem users don't have to do that.  Gem developers do.  The command can actually be considerably stupid and complex, but shortcuts will be nice.

I'm fine making the test step that adds a card schema be manual or with a 'special' script/migration.  I'm just thinking out loud about what would be nice if we can add it while testing the hard parts.

GerryG:
Closing for a new one to master.

PR #312, opened 2015-02-13
fix *default handling bug

ethn:
This is fairly critical, because it fixes brokenness in setting default content in forms (eg wagn.org/new/Ticket, in which +status should default to "open", based on the existing *default rule)

### DIFF
--- a/mod/01_core/spec/set/all/templating_spec.rb
+++ b/mod/01_core/spec/set/all/templating_spec.rb
@@ -57,6 +57,7 @@ describe Card::Set::All::Templating do
       Card::Auth.as_bot  do
         @bt = Card.create! :name=>"birthday+*right+*default", :type=>'Date', :content=>"Today!"
       end
+      @bb = Card.new :name=>'Bob+birthday'
       @jb = Card.create! :name=>"Jim+birthday"
     end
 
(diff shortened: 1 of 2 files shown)


mod/01_core/spec/set/all/templating_spec.rb
```diff
@@ -69,7 +70,9 @@
     end
     
     it "should apply to new cards" do
-      expect(Card.new(:name=>"Pete+birthday").content).to eq('Today!')
+      pb = Card.new :name=>"Pete+birthday"
+      expect(pb.raw_content).to eq('Today!')
```
ethn: the above line fails without the change. "raw_content" is what most views are supposed to use.  (you could easily make the case that #raw_content and #content should be switched, or that we need deeper changes to content handling, but that is for another day ;)

GerryG: The real deeper issue is developer documentation, where it is and should be, and creating tickets to work on it at the right moment in time.  I started to write a book http://gerry.wagn.org/It_s_Cards_all_the_Way_Down where I'm starting from a more general perspective, but ultimately we should be able to attach more detail to each part.  I intend to use our technology to make that detail *multi-dimensional*, each component of **REWD** being one of the dimensions (at a minimum).

It is a collective work in progress, so come on in and edit, add content or discussions to any of it.

PR #313, opened 2015-02-13
Email template

ethn:
I was actually just looking at this to work on what I thought was a bug but was a configuration issue on one site.

Still this clean up might be worth it. 

Better still would be to move the special from handling into a process_email_address override in right/from


mod/01_core/set/all/content.rb
```diff
@@ -16,7 +16,7 @@ def content=(value)
 end
 
 def raw_content
-  structure ? template.db_content : db_content
+  structure ? template.db_content : content
 end
```
ethn: whoops, these aren't supposed to be here (see other PR)

PR #315, opened 2015-02-13
Some of the general fixups from generator branch

GerryG:
Let me know what other changes in generator that you would want to see here, or in another branch.

Also, please feel free to suggest where we might best add test coverage for this stuff.  I think we need to cover and extend the command line tests, migrations, generators, etc.  Automated or not.  I work on doing the migration from 1.10 test soon.

### DIFF
--- a/card/lib/card/cache.rb
+++ b/card/lib/card/cache.rb
@@ -15,8 +15,9 @@ class Card
 
   class Cache
 
-    @@prepopulating     = [ 'test','cucumber' ].include? Rails.env
-    @@using_rails_cache = Rails.env =~ /^cucumber|test$/
+    TEST_ENVS = %w{test cucumber}
+    @@prepopulating     = TEST_ENVS.include? Rails.env
+    @@using_rails_cache = TEST_ENVS.include? Rails.env
     @@prefix_root       = Cardio.config.database_configuration[Rails.env]['database']
     @@cache_by_class    = {}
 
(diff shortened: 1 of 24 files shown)


card/lib/card/cache.rb
```diff
@@ -25,7 +26,7 @@ class Cache
     class << self
       def [] klass
         raise "nil klass" if klass.nil?
-        cache_by_class[klass] ||= new :class=>klass, :store=>(@@using_rails_cache ? nil : Rails.cache)
+        cache_by_class[klass] ||= new :class=>klass, :store=>(@@using_rails_cache ? nil : Cardio.cache)
```
ethn: you've tried this in a browser, right?

GerryG: In apache/passenger or is wagn s good enough?

ethn:
This looks good to me.  Will merge it in now. Bummer it won't update the diffs on the Generator branch.  Would be much cleaner.

There is this: http://makandracards.com/makandra/13043-force-github-pull-requests-to-update-the-diff-against-its-target-branch

...but I suspect if there aren't comments to preserve, it may be easier to do a new PR.

card/lib/card/cache.rb
ethn: just webrick is fine.  when I first looked at card-gem the browser failed on the cache before it loaded the first page.

PR #316, opened 2015-02-13
fix scss chunk list problem (processing pseudo urls)

ethn:
critical scss fix (was processing sorta-url-like things, even though it's not supposed to process urls at all)

### DIFF
--- a/mod/03_machines/set/type/scss.rb
+++ b/mod/03_machines/set/type/scss.rb
@@ -1,16 +1,13 @@
-include Card::Set::Type::Css
+include Type::Css
 
-format :html do
-  view :core, :mod=>Css::HtmlFormat
-  view :editor, :mod=>PlainText::HtmlFormat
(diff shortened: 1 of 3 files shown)


mod/03_machines/set/type/scss.rb
```diff
@@ -21,7 +18,11 @@ def compile_scss scss, style=:expanded
     e
   end
   
-
 end
 
 
+format( :html ) { include Css::HtmlFormat }
```
ethn: this suffices to replace html stuff above.

mod/03_machines/set/type/scss.rb
```diff
@@ -1,16 +1,13 @@
...
-  view :content_changes, :mod=>CoffeeScript::HtmlFormat
-end
-  
 def diff_args
   {:format=>:text}
 end  
 
+
 format do
+  include Css::Format
```
ethn: this fixes the bug by including the chunk_list method.

PR #317, opened 2015-02-13
jump back to top when after clicking the page/next

chuenlok:
Vignesh found that users want to jump back to top when after clicking the page/next.


mod/03_machines/lib/javascript/wagn.js.coffee
```diff
@@ -150,6 +150,9 @@ $(window).ready ->
   $('body').on 'ajax:success', '.slotter', (event, data, c, d) ->
     unless event.slotSuccessful
       $(this).slotSuccess data
+      # should scroll to top after clicking on new page
+      if $(this).hasClass "card-paging-link"
+        $("body").scrollTop 0
```
GerryG: Does this apply to next and prev?  Should it go to bottom for prev?  Does it already?

ethn: wait, they don't necessarily want to scroll to the top of the page, they want to scroll to the top of the list (the slot), right?  Remember that the thing we're paging through might be waaaay down the page.

ethn: also, please remember to make non-critical PR's to 1_15 for now ;)

ethn: (should not go to bottom for prev. we want standard paging behavior)

GerryG: Right, not the page, but of the part of the page with the pager in it.  I was thinking of the process of going through a list, if you are paging down from the bottom of one page, you want the top of the next and vice versa.  If the page is long and doesn't fit on one screenful, what is the desired UX?  A continuous pager that is more like scrolling is more typically what the designers are asking for.

ethn: I like the continuous pager idea, too.  This is specifically for the case where there are (old school) paging links at the bottom of a list.  I suspect we'd like to have both.

PR #318, opened 2015-02-14
Some constants move to config, and fix Cardio not to overwrite config va...

GerryG:
More stuff factored off generator.


card/lib/cardio.rb
```diff
@@ -18,25 +18,26 @@ def card_config *args
...
-      config.email_defaults        = nil
+      config.max_char_count        = 200                      unless config.respond_to? :max_char_count
+      config.max_depth             = 20                       unless config.respond_to? :max_depth
+      config.email_defaults        = nil                      unless config.respond_to? :email_defaults
 
-      config.token_expiry          = 2.days
-      config.revisions_per_page    = 10
+      config.token_expiry          = 2.days                   unless config.respond_to? :token_expiry
+      config.revisions_per_page    = 10                       unless config.respond_to? :revisions_per_page
+      config.closed_search_limit   = 50                       unless config.respond_to? :closed_search_limit
```
ethn: it doesn't need to hold up this merge, but I'd like better API than this respond_to? stuff.  Perhaps something like
````
config_default :toxen_expiry, 2.days
````

ethn:
Good stuff!  Thank you!

PR #319, opened 2015-02-14
Working migrations

GerryG:
Splitting off another thing from generator.  Mostly refactoring related to migrations and finding the directories, etc.  The idea is to run through all the developer use cases and migration from 1.10 forward.

As it turns out, the more critical change pulled onto this branch fully implements the engine abstraction.  The generated app now mounts Decko::Engine in the application routes.

The rest of what is on generator is just moving the old generators and adding the embedded app use case support.  It won't impact the current use case much at all (after this change).


wagn/lib/wagn/application.rb
```diff
@@ -74,8 +74,8 @@ def add_gem_path paths, path, options={}
     paths = Decko::Engine.config.paths
     # should we have add_deck_paths for these?
     paths['local-mod'] = Rails::Paths::Path.new(paths, 'local-mod', "#{Rails.root}/mod") unless approot_is_gemroot?
-    paths['db/migrate'] = Rails::Paths::Path.new(paths, "#{Rails.root}/db/migrate")
-    paths['db/migrate_deck_cards'] = Rails::Paths::Path.new(paths, "#{Rails.root}/db/migrate_cards", "#{Rails.root}/db/migrate_cards")
+    paths['db/migrate'] = Rails::Paths::Path.new(paths, 'db/migrate', r="#{Cardio.gem_root}/db/migrate")
+    paths['db/migrate_deck_cards'] = Rails::Paths::Path.new(paths, 'db/migrate_deck_cards', "#{Rails.root}/db/migrate_cards", "#{Rails.root}/db/migrate_cards")
```
GerryG: I think these are correct, but not  working yet.  Working on migrations on a different branch.

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -73,8 +74,8 @@ namespace :wagn do
 
   desc "migrate structure and cards"
   task :migrate =>:environment do
-    ENV['SCHEMA'] ||= "#{Wagn.root}/db/schema.rb"
-    
+    ENV['SCHEMA'] ||= "#{Cardio.gem_root}/db/schema.rb"
```
GerryG: Then I guess all of these should be like this one?  If schema is always in the gem.  I think that is one thing that will change a little to complete the generator work.  It may be that for the cards in another rails app case we have to extend the rails command vs. having our own (wagn or decko).  In other words, we'd have both, one for the current case of the app (wagn) and in decko-rails we extend the rails command for the new case.

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -104,9 +105,7 @@ namespace :wagn do
   desc 'insert existing card migrations into schema_migrations_cards to avoid re-migrating'
   task :assume_card_migrations do
     require 'card/migration'
-    Card::CoreMigration.schema_mode do
-      ActiveRecord::Schema.assume_migrated_upto_version Card::Version.schema(:core_cards), Card::CoreMigration.paths
-    end
+    Card::CoreMigration.assume_migrated_upto_version
```
GerryG: No schema stuff in Version now, and moved this logic into the shared class method (new).  This is adapted to the new way we handle the two types in these support classes.  I suppose tests for this would be useful now.

GerryG:
This is getting really close.  One issue remains, but at least I'm starting to understand where it comes from.

The actual code problem is the paths object, and that there is a different one for the engine and the application.  The structure migrations (db:migrate*) are looking at the application paths and we are (mostly) using the one on Decko::Engine.

I could use some help making sure we are going after the right solution.  It makes sense to me that all the migrations should be associated with the engine (and card classes).

GerryG:
Oh, and the reason the tests were failing was a missing template.  You need the routes.rb to mount the engine, and so you can set the relative root there (I guess if the code gets its relative url root from somehow introspecting this mount point ...)

Interesting that the specs don't depend on routes.rb.  That makes some sense to me.

GerryG:
Made a little more progress with migrations.

As a temporary shim, I link the app's db/migrate to the gem.  The core db:* tasks also want to write local db/schema.rb too, to db/ needs to be there.

The latest issue in card migrations is a *account+*right rule shows up as blank and throws an error.  Maybe that should be a warning, but think there could be a different migration issue that mangles some card content.

test_work/run_mig.sh
```diff
@@ -0,0 +1,29 @@
...
+wagn new test_wagn_app -c --gem-path='../../'
+cd test_wagn_app
+# load seed db (customize to your database.yml)
+echo "Load seed db (enter the pw)"
+mysql -u decko_user -p mdummy_test < ${SEEDSQL}
+# copy db config (populate .local with you user/pw)
+cp ${DBCFG} config/database.yml
+echo "migrating"
+#RAILS_ENV=test bundle exec rake db:migrate --trace
+RAILS_ENV=test bundle exec rake wagn:migrate --trace
```
GerryG: Run the migrations, should do both structure and core cards.  Needs a little help still.

test_work/run_mig.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+VER=1.15.pre
+DBCFG=../database.yml.local
+SEEDSQL=../mdummy_test.sql
```
GerryG: Configure the script here to load a different base version. Configure location of db.yml prototype.  You may have to adjust the load command for your user/dbname (mdummy_test/development, user decko_user).

In any case, this is so much to run as to give an almost working example for developers to use.

wagn/lib/wagn/generators/wagn/wagn_generator.rb
```diff
@@ -28,10 +28,10 @@ class WagnGenerator < Rails::Generators::AppBase
   
 ## should probably eventually use rails-like AppBuilder approach, but this is a first step.  
   def dev_setup
-    @gem_path = options['gem-path']
+    @gem_path = ENV['WAGN_DEV_GEM_PATH'] || options['gem-path']
```
GerryG: I think this may have been introduced on master and lost in a merge.  Here now in any case so we can do this either way.  Note the parameter version is used (and not verified in generated Gemfile) in the migration test script.

wagn/lib/wagn/generators/wagn/wagn_generator.rb
```diff
@@ -126,6 +125,7 @@ def config
 
     inside "config" do
       template "application.rb"
+      template 'routes.erb', "routes.rb"
```
GerryG: This is the main change we needed, but we got the whole file.

decko-rails/app/controllers/card_controller.rb
```diff
@@ -2,11 +2,6 @@
 
 require_dependency 'card'
 
-require_dependency 'card/content'
-require_dependency 'card/action'
-require_dependency 'card/act'
-require_dependency 'card/change'
-require_dependency 'card/reference'
```
ethn: yeah, that's better.

wagn/lib/wagn/generators/wagn/templates/config/routes.erb
```diff
@@ -0,0 +1,12 @@
+# -*- encoding : utf-8 -*-
+
+require 'decko/engine'
+
+Rails.application.routes.draw do
+
+  if !Rails.env.production? && Object.const_defined?( :JasmineRails )
+    mount Object.const_get(:JasmineRails).const_get(:Engine) => "<%= mount_point %>/specs"
+  end
```
ethn: we don't really want this Jasmine stuff in decks, do we?

wagn/lib/wagn/generators/wagn/wagn_generator.rb
ethn: no, this isn't what I was suggesting.  We want the Gemfile to have something like:
````
  gem 'wagn', :path=>"ENV['WAGN_GEM_DEV_PATH']"
````
the point is to be able to have a standard path in your profile but set a different one whenever you run a server, tests, etc.

File: .gitignore [hunk 10]
```diff
@@ -23,3 +23,4 @@ decko*gem
 *.sqlite3
 *.DS_Store
 *.swp
+test_work/*
```
ethn: test_work is here, yet there is stuff in the directory below?

test_work/run_mig.sh
```diff
@@ -0,0 +1,29 @@
...
+cd wagn/
+gem build wagn.gemspec
+mv *.gem ../card/
+cd ../decko-rails
+gem build decko.gemspec
+mv *.gem ../card/
+cd ../card
+gem build card.gemspec
+gem install *-${VER}.gem
```
ethn: why are you building/installing the gems AND setting the gem-path?

test_work/run_mig.sh
```diff
@@ -0,0 +1,29 @@
...
+# load seed db (customize to your database.yml)
+echo "Load seed db (enter the pw)"
+mysql -u decko_user -p mdummy_test < ${SEEDSQL}
+# copy db config (populate .local with you user/pw)
+cp ${DBCFG} config/database.yml
+echo "migrating"
+#RAILS_ENV=test bundle exec rake db:migrate --trace
+RAILS_ENV=test bundle exec rake wagn:migrate --trace
+#RAILS_ENV=development bundle exec rake db:migrate --trace
+#RAILS_ENV=development bundle exec rake wagn:migrate --trace
```
ethn: this is actually starting to look like a job for semaphore!

wagn/lib/wagn/generators/wagn/wagn_generator.rb
ethn: are you proposing to make it so that every deck now has to have its own routes file?  Does this mean existing wagns won't work without one?  I would prefer to make it so that sites can work without their own routes.rb file (and can have one if they want to customize); is that workable?

wagn/lib/wagn/generators/wagn/templates/config/routes.erb
```diff
@@ -0,0 +1,12 @@
...
+
+  mount Decko::Engine => '<%= mount_point %>/'
```
ethn: this is cool.

ethn: wondering about the word "mount_point".  not sure that's how folks will think of it.  it's really just the "deck_path", right?

wagn/lib/wagn/generators/wagn/wagn_generator.rb
GerryG: I guess we can still do that with a routes.rb in the wagn gem.  This is a really small routes file and it gives them a good place to set the mount point (relative root).

wagn/lib/wagn/generators/wagn/templates/config/routes.erb
GerryG: Maybe only in development mode?  I think this may be the right place for it if we make it conditional in the generation.

wagn/lib/wagn/generators/wagn/templates/config/routes.erb
GerryG: Sure, that might be a better name for it.

wagn/lib/wagn/generators/wagn/wagn_generator.rb
GerryG: Instead of --gem-path, or as one of the ways to set it?  Or maybe what you get if you don't set it?

test_work/run_mig.sh
GerryG: Mainly for dev purposes so I don't have to re-install the gems when working on them.  Also wanted to test that code path.

The point is more that we can do it a number of ways if we want, and add other scenarios as their own scripts.  I added the whole directory to .gitignore with that thought.  We'll only want to add stuff to the repo explicitly and ignore them most of the time.

test_work/run_mig.sh
GerryG: Yes, I think we can run a bunch of other scenarios there, but we probably don't want all of them running continuously.  Do they have a way of squirelling away alternative build configurations?  Seems like they directly connect the build config to the repo name.

I could put it up on my account, I guess.

.gitignore
GerryG: The idea is to ignore working script changes and explicitly add them.  I guess you may have to disable this line to do it, but I expect it won't be that common of an action.

wagn/lib/wagn/generators/wagn/wagn_generator.rb
ethn: right, if you don't set it.

wagn/lib/wagn/generators/wagn/templates/config/routes.erb
ethn: perhaps if you're doing core-dev, yeah.  But a normal decko-rails user should definitely not have this stuff.

wagn/lib/wagn/application.rb
```diff
@@ -71,11 +71,13 @@ def add_gem_path paths, path, options={}
     config.revisions_per_page    = 10
     config.request_logger        = false
 
+    # this needs to be on the application's paths object.
+    # maybe if we finally understand how these are supposed to be connected in railties we can fix lots of stuff
+    paths['db/migrate'] = Rails::Paths::Path.new(paths, 'db/migrate', "#{Cardio.gem_root}/db/migrate")
     paths = Decko::Engine.config.paths
+    paths['db/migrate'] = Rails::Paths::Path.new(paths, 'db/migrate', "#{Cardio.gem_root}/db/migrate")
     # should we have add_deck_paths for these?
     paths['local-mod'] = Rails::Paths::Path.new(paths, 'local-mod', "#{Rails.root}/mod") unless approot_is_gemroot?
```
ethn: didn't we get rid of approot_is_gemroot a long time ago?

decko-rails/spec/controllers/card_controller_spec.rb
```diff
@@ -296,7 +307,9 @@
       expect(response.status).to eq(404)
     end
   end
+
   describe "unit tests" do
+    routes { Decko::Engine.routes }
```
ethn: any shot we could put all these in a single before block?

test_work/run_mig.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
ethn: might we move this into script?  Little warty for root imo.

decko-rails/spec/controllers/card_controller_spec.rb
GerryG: I'll give it a try.  You just mean so we don't have to repeat this, right?

test_work/run_mig.sh
GerryG: We can move it, but I'm not sure I'd use script.  These probably are most related to wagn, so wagn/test/script might be good.  What do you think?

There isn't much of a dependency on the script location, but where the working directory goes is pretty important.  The script needs to know where it (the working dir, not the script dir) is relative to the repo-root, so it would be good if we all put it at the same level in the dir hierarchy.

wagn/lib/wagn/application.rb
GerryG: I may have missed it in a merge.  Or maybe there was a leftover that I kept.  So I should totally remove it?

wagn/lib/wagn/tasks/wagn.rake
ethn: I think the reason for this chunk of code is that schema migrations tend to try to update the schema.rb and we used this to hack around that problem somehow.

In general, my priority is trying NOT to put stuff in decks (schemas, routes, etc) that we may want to update.  I'm fine with making it so that folks can override default stuff in their decks; whenever you override behavior, you're willfully taking on a maintenance chore.

That said, the decko-rails case is almost certainly going to involve more overrides on average, because it's trying to squeeze wagn into a different app. So yeah, we need to be prepared for that.

test_work/run_mig.sh
ethn: wagn/test/script is fine by me.

wagn/lib/wagn/tasks/wagn.rake
GerryG: Should we?
1) Never write a schema?  Does it need that schema during the processing of structure migrations (regular rails type migrations)?
2) Never write a gem file (probably, so if we do write one, it must be in the deck?)

Is it ok to install a link for db/migrations if not present in the deck?  If it is present, we could copy/link files from the gems.

ethn: 1. We still use schema.rb in the gem for generating new decks.  Core developers update that schema when there is a new structure migration.  We don't actually need it during structure migrations for decks; the rails pattern is to update the schema.rb file then so that new copies of the app are structurally synched with old ones.
2. We have to write a Gemfile in the decks.  Otherwise the decks can't use bundler to manage their gem updates, including wagn updates.

For now, I'll use the term "rails decks" to decks not connected to a rails app (via decko-rails) and "free decks" for traditional stand-alone ones.

Free decks don't currently have any links to db/migrations in them and I don't see why it would be an improvement to have one. For rails decks, that would be seem admissible if it's the best way to get things working.

wagn/lib/wagn/generators/wagn/wagn_generator.rb
```diff
@@ -41,7 +43,7 @@ def dev_setup
       @spec_path = 'mod/'
       @spec_helper_path = './spec/spec_helper'
       @simplecov_config = "card_simplecov_filters"
-      @gem_path = options['gem-path'] || ask("Enter the path to your local wagn gem installation: ")
+      @gem_path = ask("Enter the path to your local wagn gem installation: ") unless @gem_path
```
xithan: This means we never ask. `@gem_path` is set to "ENV['WAGN_DEV_GEM_PATH']" above

xithan:
When we merge this we have to adapt Wikirate rake tasks. `Migration.paths` is used there.

card/lib/card/migration.rb
```diff
@@ -28,8 +27,12 @@ def find_unused_name base_name
       test_name
     end
 
-    def paths mig_type=type
-      Card.paths["db/migrate#{schema_suffix type}"].to_a
+    def migration_paths mig_type=type
+      Cardio.migration_paths mig_type
+    end
```
GerryG: The wikirate comment didn't seem to be anchored to code.  I think you mean this method.  I think this is a necessary rename.  All the other places, 'paths' is a rails Paths object, and this isn't.  It's an array, I think.  And it may be loaded from the other paths configuration.

And for sure we should make sure I'm not breaking any API we use unless we are making a change we can all agree to.

wagn/lib/wagn/tasks/wagn.rake
GerryG: For 1., then for the developer case, update in the gem and the dev will checking/checkout as needed.

And I meant 'write to the gem space' when I said 'gem file', not the Gemfile in the deck.

I'm not sure I get the terms quite yet: *rails decks* makes sense as a term, but it does involve decko-rails.  I might use *free decks* for the non-rails case (just model, maybe Sinatra, other web framework, or some other framework), which nobody is really considering yet.  We would know that case by the existence of a new glue gem or library (say decko-sinatra).

Maybe *deck app* for the current stand-alone rails app case.

card/lib/card/migration.rb
ethn: I think this is probably a good change.  not a big deal to change it in wikirate.  I think Philipp was making that comment as a flag to me mostly, so I'll remember to update wikirate when we push this out.

wagn/lib/wagn/generators/wagn/wagn_generator.rb
ethn: what if that's not there in the ENV?

card/lib/card/migration.rb
GerryG: That is what I thought, but wanted to be sure we were talking about the same thing.

wagn/lib/wagn/tasks/wagn.rake
ethn: app decks v. engine decks vs. whatever other future decks?

wagn/lib/wagn/generators/wagn/wagn_generator.rb
GerryG: Yes, I fixed at least one of these.  Thanks for the catch.

card/lib/card/migration.rb
ethn: I'm merging this, but should we consider making it a rails paths object?  It's basically a paths object converted to an array, no?

wagn/lib/wagn/generators/wagn/wagn_generator.rb
ethn: I'm not sure I agree that this is a bug.  When --core-dev is flagged, if there is no gem-path option and not ENV variable set, then we should ask, no?

card/lib/card/migration.rb
GerryG: It is what path[key] returns.  Is that a Paths thing.  If so, then it is, if not, that should be fine too,right?

wagn/lib/wagn/generators/wagn/wagn_generator.rb
xithan: The problem was it doesn't use the content of the ENV variable it uses the string "ENV['WAGN_DEV_GEM_PATH']" as path. So it doesn't ask even when the ENV variable is not set.

ethn: ahh, so instead of `unless @gem_path` we need `if @gem_path.blank?`, perhaps?

PR #320, opened 2015-02-15
bugfix: email links need relative root paths

ethn:
fixed verify_url and reset_password_url on sites with relative root path


mod/05_standard/set/right/account.rb
```diff
@@ -33,7 +33,7 @@ def authenticate_by_token val
   end
 
   view :reset_password_url do |args|
-    wagn_url "/update/#{card.cardname.url_key}?reset_token=#{card.token_card.refresh(true).content}"
+    wagn_url "update/#{card.cardname.url_key}?reset_token=#{card.token_card.refresh(true).content}"
```
ethn: eg, if site root is mydomain.com/wagn, then this needs to be mydomain.com/wagn/update, not mydomain.com/update

PR #321, opened 2015-02-16
bootstrap and following

xithan:


xithan:
there is one problem left: the test  "returns denial when no read permission" fails with "stack level too deep". When I try to explore the loop (for example with a breakpoint in view :layout) the loop disappears and the test fails with "expected response 403, got 200"

GerryG:
You'll still need to properly merge to 1_15.  The paths all change with that branch so there are three main subdirectories, each with a gem.  They are wagn, card and decko-rails.

ethn:
I had told him I'd work on the merge.

GerryG:
Got it.  Let me know if I can help with it.  Probably mostly just needs to move the changes to the new locations.  Unfortunately that make git a lot less help sometimes.  A lot of the file moves can be tracked.

GerryG:
We should be closing this to avoid confusion, right?  I saw a related one close, and I should it was this one ...

card/lib/cardio.rb
```diff
@@ -8,7 +8,11 @@
 module Cardio
 
   ActiveSupport.on_load :card do
-    Card::Loader.load_mods if Card.count > 0
+    if Card.count > 0
+      Card::Loader.load_mods 
+    else
+      Rails.logger.warn "empty database"
```
ethn: nice.  we may go farther, but thanks for starting this.

card/mod/01_core/set/all/collection.rb
```diff
@@ -48,8 +48,13 @@ def item_type
...
 
-def include_item? cardname
-  item_names.map{|name| name.to_name.key}.member? cardname.to_name.key
+def include_item? item
+  key = if Card === item
+    item.cardname.key
+  else
+    item.to_name.key
+  end
+  item_names.map{|name| name.to_name.key}.member? key
```
ethn: not an issue for this release (you didn't change this), but it would be nice to use find here.  (stop looking when you find it)

card/mod/01_core/set/all/initialize.rb
```diff
@@ -1,49 +1,49 @@
...
-def include_set_modules
-  unless @set_mods_loaded
-    set_modules.each do |m|
-      singleton_class.send :include, m
-    end
-    @set_mods_loaded=true
-  end
-  self
-end
+JUNK_INIT_ARGS = %w{ missing skip_virtual id }
```
ethn: is it just me, or is this a diff without much difference?

card/mod/01_core/set/all/rules.rb
```diff
@@ -1,25 +1,41 @@
...
   select rules.id as rule_id, settings.id as setting_id, sets.id as set_id, sets.left_id as anchor_id, sets.right_id as set_tag_id
-  from cards rules join cards sets on rules.left_id = sets.id join cards settings on rules.right_id = settings.id
+  from cards rules 
+  join cards sets     on rules.left_id  = sets.id 
+  join cards settings on rules.right_id = settings.id
   where sets.type_id     = #{Card::SetID }    and sets.trash     is false
   and   settings.type_id = #{Card::SettingID} and settings.trash is false
-  and                                             rules.trash    is false;
+  and                                             rules.trash    is false
+  and   (settings.codename != 'follow' or rules.db_content != '');
```
ethn: meaning you want follow rules so long as they're not blank?

card/mod/01_core/set/all/rules.rb
```diff
@@ -1,25 +1,41 @@
...
-  from cards read_rules join card_references refs on refs.referer_id = read_rules.id join cards sets on read_rules.left_id = sets.id
+  from cards read_rules 
+  join card_references refs on refs.referer_id    = read_rules.id 
+  join cards sets           on read_rules.left_id = sets.id
   where read_rules.right_id = #{Card::ReadID} and read_rules.trash is false and sets.type_id = #{Card::SetID};
-}  
+}
+
 
 def is_rule?
```
ethn: might break these out and do:
````
def is_rule?
  is_standard_rule? || is_user_rule?
end
````

File: card/mod/01_core/set/all/rules.rb [hunk 6]
```diff
@@ -32,6 +48,15 @@ def rule_card setting_code, options={}
...
   fallback = options.delete( :fallback )
+  
+  if Card::Setting.user_specific? setting_code 
+    user_id = options[:user_id] || (options[:user] and options[:user].id) || Auth.current_id
+    if user_id
+      fallback = "#{setting_code}+#{Card[:all].id}"
+      setting_code = "#{setting_code}+#{user_id}"
+    end
+  end
+
```
ethn: this is cool. I'm tempted to find ways to break the user stuff with a parallel api around the word "preference", but I really like what you've done here.

card/mod/01_core/set/all/rules.rb
```diff
@@ -72,26 +130,155 @@ def toggle val
...
+  def write_user_ids_cache hash
+    Card.cache.write 'USER_IDS', hash
+  end
+  
+  def write_rule_keys_cache hash
+    Card.cache.write 'RULE_KEYS', hash
+  end
+  
+  
+
```
ethn: man, this is a chunk of work!

I feel like I should write up comments for it to be sure that I understand it.  But I can't spot any problems with a quick look, and despite all the new functionality, if anything the code is more handsome than what it replaces.

card/mod/01_core/set/all/tracked_attributes.rb
```diff
@@ -41,8 +41,9 @@ def extract_subcard_args! args={}
   if is_rule?
```
ethn: this code, on the other hand, is still an ugly duckling. It's not your duckling, but it's still ugly ;)

card/mod/03_machines/lib/javascript/wagn.js.coffee
```diff
@@ -253,6 +253,11 @@ $(window).ready ->
 
   $('body').on 'change', '.edit-type-field', ->
     $(this).closest('form').submit()
+    
+  $('body').on 'change', '.submit-select-field', ->
+    field = $(this).closest('.card-editor').find('.card-content')
+    field.val( $(this).val() )
+    $(this).closest('form').submit()
```
ethn: this is standard bootstrappiness?

ethn:
Wow. I new this was a big PR, but I think I may still have underestimated all the work involved here.

I won't pretend to grok it all yet, but I'll keep working my way through it as we test on wikirate and such.  I may continue to comment on this PR even after merging.  I think it's time to do that now!

card/lib/cardio.rb
xithan: I wasn't sure whether we want to raise an exception here

card/mod/01_core/set/all/initialize.rb
xithan: it's without any difference. I didn't touch this file

card/mod/01_core/set/all/rules.rb
xithan: exactly, blank follow rules are treated as non-existent.  I wanted to write `!(settings.codename = 'follow' and rules.db_content = '')` but I suspected that to be invalid sql syntax

card/mod/01_core/set/all/rules.rb
xithan: good point, changed.

card/mod/03_machines/lib/javascript/wagn.js.coffee
xithan: no, that was part of an outdated 'follow' interface (list of text fields without submit buttons to update and create follow rules). I'll remove it.

card/mod/01_core/set/all/rules.rb
xithan: "preference" is a good name for user rules. We can make Set+User the preference view similar to the view of Set cards.

card/lib/cardio.rb
GerryG: I've wasted a lot of debug time because I didn't notice the seeds didn't load and therefor I had no mods loaded ...

I think yes, raise an exception.  Although, I think the point of the test is to make sure it does not throw and exception because lots of mods depend on data (WagnbotID at a minimum).  That would mean we'd have to catch the exception we throw here when we need to load with an empty database.  If not an exception, maybe use 'error' logging level.

card/mod/01_core/set/all/rules.rb
GerryG: I think you can use NOT (... and...), or for an exercise, express with AR relations and joins :)

card/mod/01_core/set/all/rules.rb
GerryG: This looks like something pretty new.  The name 'setting_code' is a little confusing.  That has to be a card name/key, right?  Converted from a codename, but not a codename?  Any wagneer docs yet?

ethn: setting_code is the code name of a setting card.  that's what's used in the rule cache.

GerryG: Oh, that's a cache key.  Nevermind.

GerryG:
Were there migrations with this change?  I was re-testing migration from 1.10 on my generator branch and saw: 
==  WatchersToFollowing: migrating ============================================
rake aborted!
StandardError: An error has occurred, all later migrations canceled:

undefined method `add_item' for #<Card###70309891776460[basic:3](Sara+*following){new:I:nil}>/Users/ggleason/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/gems/activemodel-3.2.16/lib/active_model/attribute_methods.rb:407:in `method_missing'
/Users/ggleason/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/gems/activerecord-3.2.16/lib/active_record/attribute_methods.rb:149:in `method_missing'
/Users/ggleason/Sites/p2p/wagn/card/db/migrate_core_cards/20141204061304_watchers_to_following.rb:26:in `block (2 levels) in up'

GerryG:
Seems like there needs to be a rule for User+*following is that a search or does it default to Pointer?

xithan:
That's a bigger issue. Thanks for finding that one. I forgot to test the 1.10 migration.  
There used to be a card_accessor :following for user cards where we stored what users are following. But we have user rules for this now. I have to fix that migration and add another one that translates +*following cards to follow rules.

PR #322, opened 2015-02-16
Boot and follow - merged

ethn:
...hmm, but it says it can't automatically merge? so maybe I goofed?


ethn:
crap.  I had this all working, and then something went horribly wrong with my commit history.  getting late. will fix tomorrow.

PR #323, opened 2015-02-16
Bofo (bootstrap and following -- merged)

ethn:
not atp yet, but not too far off.  might be some easy fixes here, but I'm off to bed ;)


card/lib/card/format.rb
```diff
@@ -224,7 +224,11 @@ def hide_views args
     end
     
     def parse_view_visibility val
-      (val || '').split( /[\s\,]+/ ).map { |view| canonicalize_view view }
+      if val.kind_of? Array
+        val.map { |view| canonicalize_view view }
+      else
+        (val || '').split( /[\s\,]+/ ).map { |view| canonicalize_view view }
```
GerryG: This will squeeze out empty elements of the list.  Is that was is wanted?  (i.e.  a,b,,c won't have an empty string between b and c, so you can't get that if that use case is wanted)

card/mod/03_machines/lib/stylesheets/functional.scss
```diff
@@ -150,9 +150,9 @@ fieldset {
 }
 
 
-.navbox-form {
-  display: inline
-}
+//.navbox-form {
+//  display: inline
+//}
```
GerryG: Is this really meant to stay with comment marks?

card/mod/05_standard/set/all/links.rb
```diff
@@ -31,7 +31,11 @@ def web_link href, opts={}
   # link to a specific card
   def card_link name, opts={}
     text = (opts.delete(:text) || name).to_name.to_show @context_names
-    name = Card[name].cardname if Symbol===name
+    name = case name
+           when Symbol then Card[name].cardname
+           when Card   then name.cardname
+           else             name
+           end
```
GerryG: I'm not sure this is right.  I don't think this method is meant to take a Card, and the symbol case probably should just used Codename and not instantiate a card.  I think this should work for a symbol even if there is no card with that codename.

card/mod/05_standard/set/rstar/rules.rb
```diff
@@ -1,9 +1,10 @@
+
 format :html do
 
   view :closed_rule, :tags=>:unknown_ok do |args|
     return 'not a rule' if !card.is_rule? #these are helpful for handling non-rule rstar cards until we have real rule sets
       
-    rule_card = card.new_card? ? find_current_rule_card[0] : card
+    rule_card = card.new_card? ? find_current_rule_card : card
```
GerryG: Was this just broken, or did find_current_rule_card change?

card/mod/05_standard/set/type/set.rb
```diff
@@ -1,6 +1,6 @@
 
 format :html do
-
+
```
GerryG: Let's try to stay clean on invisible spaces.  I know that TextMate is bad about introducing this stuff.

card/lib/card/format.rb
xithan: The method is used for the 'show' and 'hide' inclusion options. An empty entry would be uncommon. Moreover the string handing didn't change. I added support for arrays so that I can call
`nest my_card, :hide=>[ view1, view2, view3]` and don't have to generate a string first or convert the array to a hash  `{view1=>:hide, view2=>:hide, view3=>:hide}`

card/mod/03_machines/lib/stylesheets/functional.scss
xithan: This should be removed in bootstrap_functional.scss and not in functional.scss

card/mod/05_standard/set/rstar/rules.rb
xithan: It changed. It returns only the current_rule_card now and not an array with additional stuff.

card/mod/05_standard/set/all/links.rb
xithan: Why not allowing to use a Card object?

card/lib/card/format.rb
GerryG: Yes, probably what we want here, and it did not change.  I hope this doesn't need to be redone much, although canonicalize_view isn't that expensive and these lists will rarely be long.  You can DRY up the map since it is the same for both paths.

card/mod/05_standard/set/all/links.rb
GerryG: It seems like it can't hurt to be more specific in how we call it.  If you have the card, adding .cardname isn't ugly, and if you look at the uses, I think cardname is usually what it has already.  I can go either way and this and will defer to Ethan's opinion.

I think the point about looking up the card for the symbol case is more important.  At the very least it should :skip_modules, but we are guaranteed to find the codename in the codenames hash, and that is the fastest lookup we can do here.  We don't need the card here, just the name.

card/mod/05_standard/set/rstar/rules.rb
```diff
@@ -217,17 +267,106 @@ def find_current_rule_card
     # self.card is a POTENTIAL rule; it quacks like a rule but may or may not exist.
     # This generates a prototypical member of the POTENTIAL rule's set
     # and returns that member's ACTUAL rule for the POTENTIAL rule's setting
-    set_prototype = card.trunk.prototype
-    rule_card = if card.new_card?
-      setting = card.right and set_prototype.rule_card setting.codename   
+    if card.new_card?
+       ((setting = card.right) && card.set_prototype.rule_card(setting.codename, :user=>card.rule_user)) ||
+            Card.new(:name=> "#{Card[:all].name}+#{card.rule_user_setting_name}")
```
GerryG: Ok, I get it now.  It used to return the prototype, but now you can get that from the card if you need it.  Just returns the rule card.  Good stuff.

card/mod/05_standard/set/all/links.rb
GerryG: Scratch the comment about should work if it doesn't exist.  My proposal wouldn't make that work.  There can't be a codename without a card.  Still good to avoid any Card stuff for that case.

card/mod/03_machines/lib/stylesheets/functional.scss
ethn: I'm going to have a look at the css organization and (potentially) propose and work on re-arrangements.  But that can happen after merging.

card/lib/card/format.rb
ethn: not a bad idea.

card/mod/05_standard/set/all/links.rb
ethn: :skip_modules is a good idea.

ethn:
I'm closing this in favor of xithan's PR #321

card/mod/05_standard/set/all/links.rb
GerryG: Better to not even fetch in that case.  The Card case, is fine, you have a card.  Best if this method never does card fetch, no?

ethn: The codename cache is just a mapping between codenames and ids.  If you want the name you have to use fetch.  Otherwise the codename cache would have to get involved in cache clearing for name changes, which would suck.

GerryG: D'oh.

But, then skip_modules is still needed as mentioned.  I suppose it wouldn't be a bad thing to switch the default for [].  then Card[foo].include_sets (or what it is really called) is still nice and compact.

PR #326, opened 2015-02-18
migration fix

xithan:


wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -73,12 +74,15 @@ namespace :wagn do
...
-    
+    ENV['SCHEMA'] ||= "#{Cardio.gem_root}/db/schema.rb"
+
     stamp = ENV['STAMP_MIGRATIONS']
 
     puts 'migrating structure'
+    
+    Wagn.paths['db/migrate'] = ["#{Cardio.gem_root}/db/migrate"]
     Rake::Task['db:migrate'].invoke
+    #Rake::Task['wagn:migrate:structure'].invoke
```
xithan: Here is a serious problem. db:migrate uses the wrong path. The new 'wagn:migrate:structure' task has the problem that it doesn't update schema.rb

wagn/features/step_definitions/wagn_steps.rb
```diff
@@ -213,7 +207,7 @@ def signed_in_as username
 end
 
 When /^I hover over the main menu$/ do
-  page.execute_script "$('#main > .card-slot > .card-header > .card-menu-link').trigger('mouseenter')"
+  page.execute_script "$('#main > .card-slot > .panel-heading > .card-header > .card-menu-link').trigger('mouseenter')"
 end
```
xithan: That file shouldn't be here. It's used for 'follow' cukes

wagn/lib/wagn/tasks/test.rake
```diff
@@ -74,7 +74,7 @@ namespace :test do
     ActiveRecord::Base.establish_connection
     tables.each do |table_name|
       i = "000"
-      File.open("#{Wagn.gem_root}/test/fixtures/#{table_name}.yml", 'w') do |file|
+      File.open("#{Cardio.gem_root}/test/fixtures/#{table_name}.yml", 'w') do |file|
         data = ActiveRecord::Base.connection.select_all(sql % table_name)
         file.write data.inject({}) { |hash, record|
           record['trash'] = false if record.has_key? 'trash'
```
xithan: Is `Wagn.root` correct at line 2 and 6?

ethn: I think so.  That basically still means "deck root"

wagn/lib/wagn/tasks/wagn.rake
xithan: Github is acting weird. It shows different file content when I click on "View full changes" (the code on  the right side in the screenshot is what I commited) and it shows a lot of code changes that should be already merged in with PR #319

migration_fix_by_xithan_ _pull_request__326_ _wagn_wagn_und_migration_fix_by_xithan_ _pull_request__326_ _wagn_wagn_und_databases_rake_ _activerecord-4_1_1__git__migration-fix_

ethn: OK, yeah, that's very weird.

On Wed, Feb 18, 2015 at 4:29 AM, xithan <notifications@github.com> wrote:

> In wagn/lib/wagn/tasks/wagn.rake
> <https://github.com/wagn/wagn/pull/326#discussion_r24894928>:
>
> >      Rake::Task['db:migrate'].invoke
> > +    #Rake::Task['wagn:migrate:structure'].invoke
>
> Github is acting weird. It shows different file content when I click on
> "View full changes" (the code on the right side in the screenshot is what I
> commited) and it shows a lot of code changes that should be already merged
> in with PR #319 <https://github.com/wagn/wagn/pull/319>
>
> [image: migration_fix_by_xithan_ _pull_request__326_
> _wagn_wagn_und_migration_fix_by_xithan_ _pull_request__326_
> _wagn_wagn_und_databases_rake_ _activerecord-4_1_1__git__migration-fix_]
> <https://cloud.githubusercontent.com/assets/13584/6246391/bff7a98e-b768-11e4-8a08-70428122b164.png>
>
> —
> <https://github.com/wagn/wagn/pull/326/files#r24894928>.
>



-- 
Ethan McCutchen
One of the Wagneers, Wagn.org

Wagn. How pioneers roll.

s: ethan.mccutchen
t: @intogreater

wagn/lib/wagn/tasks/test.rake
GerryG: It does seem right, but what actually uses this?  It seems kind of aggressive to be setting the 'database' for production and development in that method.  Does it restore the original when it is done?  What if it is interrupted?

ethn: not sure I get this comment. it's a rake task.  "what actually uses this" is an administrator (directly).  where is it setting the database?

GerryG: He referred to line 2 and 6 in the top comment.  That code reads, updates and writes database.yml, which is pretty special purpose.  At first I thought it was related to generating an app, but it is for configuring the database names in database.yml.  Is that for an app generated by a test script?  That would make some sense.  Otherwise, why are we generating a database.yml in a rake task?

ethn: Ah. I think it's fair to say that code needs an update.

It's basically legacy code from the pre-gem days when changing from working on one site to another would generally mean editing config files.  Lewis created code shortcuts for this and used the trick for populating the template database, but now there are better ways!

Afaict, it's only used in generating fixtures at the moment, but still... 

Let's move the conversation off this PR, but good point!

PR #329, opened 2015-02-20
"follow" updates

xithan:


card/db/migrate_core_cards/20141204061304_watchers_to_following.rb
```diff
@@ -22,7 +22,7 @@ def up
     
       follower_hash.each do |user, items|
         if card=Card.fetch(user) and card.account
-          following = card.fetch :trait=>"following", :new=>{}
+          following = Card.fetch("#{card.name}+*following", :new=>{:type=>'pointer'})
```
GerryG: The trait fetch with the type should work the same, right?  The new version will be wrong if the *following card is renamed.  Likewise on the :type param, should be :typecode or :type_id forms, I think.

card/db/migrate_core_cards/20150220134731_following_to_follow_rule.rb
```diff
@@ -0,0 +1,19 @@
...
+    Card.search(:right=>{:codename=>'following'}, :left=>{:type=>'user'}).each do |following_card|
+      user_name = following_card.cardname.left
+      following_card.item_names.each do |followed_set_name|
+        
+        set_card = Card.fetch(follwed_set_name, :new=>{})
+        if (!set_card.type_code == :set)
+          set_card = set_card.default_follow_set_card
+        end
+        rule = Card.fetch set_card.follow_rule_name(user_name), :new=>{:type=>'pointer'}
+        rule.content = "[[always]]"
```
GerryG: I think you can put this in the :new params.  Can it exists already?  If so, then I guess this needs to be separate.  Unless it is a warning/error situation if it exists.  Then we should test whether it is new.

card/db/migrate/20150220141031_create_deck_routes_file.rb
```diff
@@ -0,0 +1,14 @@
...
+    routes_path = File.join Wagn.root, 'config', 'routes.rb'
+    if !File.exist? routes_path
+      template_path = File.join Wagn.gem_root, 'lib', 'wagn', 'generators', 'wagn', 'templates', 'config', 'routes.erb'
+      @include_jasmine_engine = false
+      decko_path = ''
+      routes = ERB.new(File.read template_path).result binding
+      File.open(routes_path, "w") { |f|  f.write routes }
+    end
+  end
+end
```
ethn: mentioned this in call, but I don't think file management belongs in migrations.  that would mean, eg, you could copy a migrated db to an installation without a routes.rb file, and it would then be broken and there wouldn't be a clear mechanism for fixing it.

If we were sure we wanted routes.rb files in app decks, we would probably be better off with a rake task that creates a routes.rb file when it doesn't exist.  Then we could have `wagn update` run that task.

HOWEVER, I think I would suggest this, instead:
1. for now, app decks should use the old pattern: routes.rb in gem.
2. deck apps, on the other hand, will need it generated.
3. if app deck managers want to override this, they can add something like this to their application.rb
````
config.paths.add 'config/routes', :with=>'config/routes.rb'
````
We could also add a method like `use_local_routes` as a shortcut.

card/db/migrate_core_cards/20141204061304_watchers_to_following.rb
ethn: good points.

I keep thinking we might want to consider some new API for this.

card/db/migrate_core_cards/20150220134731_following_to_follow_rule.rb
ethn: I suspect this approach is probably fine, as it's unlikely for it to exist, and it's possible for it to exist and not cause a problem if updated this way.

I should note that it can really cause headaches when we raise errors over minor problems in content migration.  It kills the whole migration and reverts everything it's done, often for pretty minor errors that would not otherwise prevent the site from running normally.  This is particularly a problem on Cloudstore, when we're usually trying to get through a bunch of migrations at once, and tracking down individual migration errors can get tricky.

I think what I would suggest is that we support an optionally separate "migration errors" log for both fatal and non-fatal migration errors.  As a maintainer of lots of wagn sites, this would be a big help.

None of this should block this PR!

card/db/migrate/20150220141031_create_deck_routes_file.rb
GerryG: I think I agree, but one question.  How should we configure this difference between the two cases?  I guess it is at the config/application.rb level.  For deck apps, I think we will have a small number of lines we need to add to an existing config/application.rb.  It seems like the deck app case should not require wagn/application, and they won't inherit from Wagn::Application either.  With that in mind, I think some things in wagn/application.rb need to be in a decko file, whether engine or a railtie thing.  We're close.

ethn: My previous comment explains how app decks handle it, right?  The application.rb can override the default setting to get the routes file from the gem.

Engines, obviously, don't have an application.rb.  They have an engine.rb.  There you can override the default setting to get the routes file from the deck.

GerryG: I think you meant that reverse, but yes.  With nothing in config/application.rb, wagn/application.rb tells it to find it in the gem, but you can override to get it from the deck instead of the gem.  The engine deck case will not use wagn/application.rb and get the standard config/routes.rb where the developer has added the Decko::Engine mount.  If this gets any significant use, then a shortcut might be good.

And, in the engine deck case, there still is a config/application.rb, but we don't generate it, and it is for the hosting application.  Now, most of application.rb is comments about config settings that you could change, and those might be useful for engine developers too.  I'm not worried about that now, although it might be nice to have all those comments somewhere so they could add them manually if desired.

ethn: No, not reversed.

When you create an app deck, it creates an application.rb file (in config) that inherits from Wagn::Application (soon Decko::Application).

When you create an engine deck, it should create an engine.rb file that inherits from Decko::Engine.  When I said "engines don't have an application.rb", I meant in the engine itself.   *Engine decks should generally not require editing the config/application.rb file of the host rails app.*  Imo they should require an edit to the host app's config/routes.rb file and, generally speaking, nothing else.

The engine deck app's engine.rb file should be a lot like an app deck's application.rb file.  It should have a lot of common options demonstrated and commented out.   These templates belong in decko-rails.

GerryG: I may be finally getting something here.

My understanding:
We will have wagn/lib/decko/engine.rb, which is more or less the file from the decko-rails gem.  That is what is mounted in an engine app, and what is mounted in wagn/.../config/routes.rb.

Are you saying there will be another class MyAppEngine < Decko::Engine ?  I'm not sure we would need that.  My original thinking was that Decko::Engine.config would be connected by railties with MyApp.config, but now I'm not sure if that is desirable or possible.  Next thought is that we can copy the ones Decko needs from the application.config object in one of the initializers.

Do you have any better understanding and suggestions about how this should work?

ethn: Yes!

If you generate a standard rails engine, you'll get:
````
module Blorgh
  class Engine < ::Rails::Engine
  end
end
````

if you generate a decko engine, you'll get:
````
module Blorgh
  class Engine < ::Decko::Engine
     # here' how you add recaptcha to decko:
     #config.recaptcha_keys blah blah blah
  end
end
````
I wouldn't go so far as to suggest this is the only possible way, but it's highly desirable.  That's how you configure the engine; it's how you give people a good running start on an engine deck by giving them editable defaults.  And if you decide, "hey, I actually think I want this to be a stand-alone wagn now (an app deck), then you can just change "Engine" to "Application", move a few files around, and roll on.

If there are major distinctions between what app decks and engine decks want the defaults of Decko::Engine to be, then we can set them in decko (gem) and override them in decko-rails.

ethn: I'd still like to take this migration out (config/routes.rb is no longer needed). If we do that we can get this merged in.  Gerry's comments on renaming *following cards is valid, but there's probably very low risk there.  In other words, it would be awesome if you could address those, but wagn is more broken without this PR than with ;)

xithan:
The postgres tests are failing because of this line `Card.search(:editor_of=>card.name, :return=>:id).include? user_id`. `Card.search(:editor_of=>card.name, :return=>:id)` returns Fixnums with mysql but Strings with postgres :scream:

PR #330, opened 2015-02-21
Routes - back to gem

ethn:
also moved Decko.gem_root (et al) to Decko::Rails.gem_root.  Much more forward compatible, since Wagn will soon be Decko!


ethn:
doh, wrong target.

PR #331, opened 2015-02-21
Routes - back to gem

ethn:
also moved Decko.gem_root (et al) to Decko::Rails.gem_root.  Much more forward compatible, since Wagn will soon be Decko!

### DIFF
--- a/decko-rails/lib/decko/engine.rb
+++ b/decko-rails/lib/decko/engine.rb
@@ -12,7 +12,6 @@ require 'kaminari'
 require 'diff/lcs'
 require 'diffy'
 
-DECKO_GEM_ROOT = File.expand_path('../../..', __FILE__)
 
 module Decko
 
@@ -27,7 +26,7 @@ module Decko
 
   end
 
-  class Decko::Engine < Rails::Engine
+  class Engine < ::Rails::Engine
     
     paths.add "app/controllers", :eager_load => true
     paths.add 'gem-assets',      :with => 'public/assets'
@@ -39,17 +38,17 @@ module Decko
           Decko::Engine.paths['request_log'] = Wagn.paths['request_log']
           Decko::Engine.paths['log']         = Wagn.paths['log']
         else
-          Cardio.card_config Rails.application.config
+          Cardio.card_config ::Rails.application.config
         end
-        Cardio.cache == Rails.cache
+        Cardio.cache == ::Rails.cache
         
-        ActiveRecord::Base.establish_connection(Rails.env)
+        ActiveRecord::Base.establish_connection(::Rails.env)
       end
       ActiveSupport.on_load(:after_initialize) do
         begin
           require_dependency 'card' unless defined?(Card)
         rescue ActiveRecord::StatementInvalid => e
-          Rails.logger.warn "database not available[#{Rails.env}] #{e}"
+          ::Rails.logger.warn "database not available[#{::Rails.env}] #{e}"
         end
       end
     end
--- a/decko-rails/lib/decko/rails.rb
+++ b/decko-rails/lib/decko/rails.rb
@@ -1,0 +1,15 @@
+DECKO_RAILS_GEM_ROOT = File.expand_path('../../..', __FILE__)
+
+module Decko
+  module Rails
+    class << self
+      def root
+        ::Rails.root
+      end
+
+      def gem_root
+        DECKO_RAILS_GEM_ROOT
+      end
+    end
+  end
+end
--- a/wagn/lib/wagn/application.rb
+++ b/wagn/lib/wagn/application.rb
@@ -84,14 +84,14 @@ module Wagn
         
         paths.add 'files'
         
+        #more consistent to call this deck mod...
         paths.add 'local-mod', :with=>'mod'
         paths['app/models'] = []
         paths['app/mailers'] = []
         
-    
-        # should we have add_deck_paths for these?
-        add_gem_path paths, "lib/tasks",           :with => "lib/wagn/tasks", :glob => "**/*.rake"
+        add_gem_path paths, "lib/tasks",     :with => "lib/wagn/tasks", :glob => "**/*.rake"
         add_gem_path paths, 'gem-assets'
+        add_gem_path paths, 'config/routes', :with => 'lib/wagn/config/routes.rb'
         
         paths
       end
--- a/wagn/lib/wagn/config/routes.rb
+++ b/wagn/lib/wagn/config/routes.rb
@@ -1,0 +1,3 @@
+Wagn.application.routes.draw do
+  mount Decko::Engine => '/'
+end
--- a/wagn/lib/wagn/config/routes.rb
+++ b/wagn/lib/wagn/config/routes.rb
@@ -1,3 +1,6 @@
 Wagn.application.routes.draw do
+  if !Rails.env.production? && Object.const_defined?( :JasmineRails )
+    mount Object.const_get(:JasmineRails).const_get(:Engine) => "/specs"
+  end
   mount Decko::Engine => '/'
 end
--- a/wagn/lib/wagn/generators/wagn/templates/config/routes.erb
+++ b/wagn/lib/wagn/generators/wagn/templates/config/routes.erb
@@ -1,8 +1,6 @@
 # -*- encoding : utf-8 -*-
 
-require 'decko/engine'
-
-Rails.application.routes.draw do
+Wagn.application.routes.draw do
 <% if @include_jasmine_engine %>
   if !Rails.env.production? && Object.const_defined?( :JasmineRails )
     mount Object.const_get(:JasmineRails).const_get(:Engine) => "<%= decko_path %>/specs"
--- a/wagn/lib/wagn/generators/wagn/wagn_generator.rb
+++ b/wagn/lib/wagn/generators/wagn/wagn_generator.rb
@@ -16,7 +16,7 @@ class WagnGenerator < Rails::Generators::AppBase
     desc: "Prepare deck for wagn core testing"
     
   class_option 'gem-path', :type => :string, aliases: '-g', :default => false, :group => :runtime, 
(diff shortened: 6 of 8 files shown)


decko-rails/lib/decko/engine.rb
```diff
@@ -27,7 +26,7 @@ def gem_root
 
   end
 
-  class Decko::Engine < Rails::Engine
+  class Engine < ::Rails::Engine
     
     paths.add "app/controllers", :eager_load => true
     paths.add 'gem-assets',      :with => 'public/assets'
```
ethn: note: when we inherit from this, we won't be able to add these paths like this any more!

wagn/lib/wagn/generators/wagn/wagn_generator.rb
```diff
@@ -16,7 +16,7 @@ class WagnGenerator < Rails::Generators::AppBase
     desc: "Prepare deck for wagn core testing"
     
   class_option 'gem-path', :type => :string, aliases: '-g', :default => false, :group => :runtime, 
-    desc: "Path to local gem installation (Default, use env WAGN_DEV_GEM_PATH)"
+    desc: "Path to local gem installation (Default, use env WAGN_GEM_PATH)"
```
ethn: Also shortened this constant name.

I set it up so my bash profile now has
````
export WAGN_GEM_PATH="~/dev/wagn/gem"
````
Loving this pattern!

Sorry if you've already done this and I broke it :(

**Comment by GerryG** [comment]
Looks like you just have to get the Jasmine engine mounted for test/dev environments.  I think it is fine to have the conditional in routes.rb and have it there are not based on the presence of the right gems.

File: decko-rails/lib/decko/rails.rb [hunk 3]
```diff
@@ -0,0 +1,15 @@
+DECKO_RAILS_GEM_ROOT = File.expand_path('../../..', __FILE__)
+
+module Decko
+  module Rails
```
GerryG: Do we need this module?  I was probably just following the pattern that included Wagn.root and Rails.root.

decko-rails/lib/decko/rails.rb
```diff
@@ -0,0 +1,15 @@
...
+    class << self
+      def root
+        ::Rails.root
+      end
+
+      def gem_root
```
GerryG: And in particular, gem_root isn't related to Rails at all.

wagn/lib/wagn/application.rb
```diff
@@ -84,14 +84,14 @@ def paths
         
         paths.add 'files'
         
+        #more consistent to call this deck mod...
```
GerryG: Agreed, why don't you go ahead and make that change.

wagn/lib/wagn/application.rb
```diff
@@ -84,14 +84,14 @@ def paths
...
         paths.add 'local-mod', :with=>'mod'
         paths['app/models'] = []
         paths['app/mailers'] = []
         
-    
-        # should we have add_deck_paths for these?
-        add_gem_path paths, "lib/tasks",           :with => "lib/wagn/tasks", :glob => "**/*.rake"
+        add_gem_path paths, "lib/tasks",     :with => "lib/wagn/tasks", :glob => "**/*.rake"
         add_gem_path paths, 'gem-assets'
+        add_gem_path paths, 'config/routes', :with => 'lib/wagn/config/routes.rb'
```
GerryG: I think it is more consistent for this file to be in wagn/config/routes.rb
The directory wagn/lib/wagn/config should not exist.

wagn/lib/wagn/config/routes.rb
```diff
@@ -0,0 +1,3 @@
+Wagn.application.routes.draw do
```
GerryG: You need the Jasmine mount here, or you need to have two files and a condition on assigning the path for routes.

wagn/lib/wagn/generators/wagn/templates/config/routes.erb
```diff
@@ -1,8 +1,6 @@
 # -*- encoding : utf-8 -*-
 
-require 'decko/engine'
-
-Rails.application.routes.draw do
+Wagn.application.routes.draw do
```
GerryG: This is wrong.  Now this will only be used for the engine deck cases, and we don't want any wagn dependencies there.  The actual Rails.application will have a class determined by the application, and it will not be Wagn related.  Does this even work?  I'm not sure Wagn.application actually exists, nor should it.

GerryG:
That's an odd failure.  Random or due to other changes?

decko-rails/lib/decko/rails.rb
ethn: I think (a) we will likely want to have a module based on the gem name for auto-require, and (b) this is the correct location:

http://guides.rubygems.org/name-your-gem/

`def root` may make more sense in the engine, but `def gem_root` belongs here, imo.

decko-rails/lib/decko/rails.rb
ethn: the name of the gem is decko-rails.

wagn/lib/wagn/application.rb
ethn: we discussed this before and moved it here (as in the master branch).  the short of it is that we didn't really like config as a root directory.  it makes sense in an app, but less so in our development framework.

We want to organize the gem directories around teaching Wagn, not around conforming to rails patterns that we may soon break away from.

decko-rails/lib/decko/engine.rb
GerryG: I'm not sure about that.  Not sure I totally understand how it should work either.

What inherits from what?  I never got clear on whether an Application and the Engines in may mount actually share config and paths.  Presumably if the can/should, I haven't configured it right with Railties, and if not, then I don't know what this comment implies.  I don't think Rails::Engine will have those in inherit, only instances of subclasses would (at least I think that is the case).

wagn/lib/wagn/generators/wagn/templates/config/routes.erb
ethn: this is in Wagn.  Engine decks won't use this.

Please try to avoid using the Rails constant wherever possible.  Bit by bit we may be able to drop our dependencies on rails gems.

card/mod/01_core/spec/set/all/rules_spec.rb
```diff
@@ -121,8 +121,8 @@
           Card.create(:name => "Book+*type+Joe User+*follow", :content => "[[Home]]")
           Card::Auth.current_id = Card.fetch('Joe Admin').id
           Card.create(:name => "Book+*type+Joe Admin+*follow", :content => "[[Home]]")
-          user_ids = Card.all_user_ids_with_rule_for( Card.fetch("Book+*type"), :follow )
-          expect(user_ids).to eq [Card['Joe User'].id, Card['Joe Admin'].id]
+          user_ids = Card.all_user_ids_with_rule_for( Card.fetch("Book+*type"), :follow ).sort
+          expect(user_ids).to eq [Card['Joe User'].id, Card['Joe Admin'].id].sort
         end
```
ethn: I think this sorting should fix the tests (confession: I pushed before testing.  whoops).

decko-rails/lib/decko/rails.rb
GerryG: Oh, then '-' in a name implied '/' in the require name?  I didn't completely understand that.

decko-rails/lib/decko/rails.rb
GerryG: I finally get that now, sorry.  Then this goes away if we combine wagn and decko-rails into decko.  That now seems correct to me.

wagn/lib/wagn/application.rb
GerryG: I mad this change in my PR to your routes branch.  I think it is a good idea.

wagn/lib/wagn/application.rb
GerryG: I guess I'm ok with that.  As I think about it, I'm not sure which is the recommended rails practice anyway.  This is for supporting a rails app with its routes in a gem, so I think best rails practice is relevant, but as I said, I'm not sure which one it is.  Seems like putting *un-namespaced* stuff down under lib/wagn or lib/decko would be at least as confusing as putting them in config.  I had a really hard time finding this when I was trying to change where it found it in the initial splitting.

wagn/lib/wagn/generators/wagn/templates/config/routes.erb
GerryG: Sorry, I thought this was in the engine deck template.  I'll make sure I got the locations right.

For the engine deck case, should this now be Decko::Rails.application?  That would still need to reference Rails, but I don't see that would be a problem in that file, right?

decko-rails/lib/decko/engine.rb
ethn: I think I spelled out an inheritance proposition in the Generator branch.

wagn/lib/wagn/application.rb
ethn: This does make some since.  Maybe we should have one root directory called "rails-app" or something?

wagn/lib/wagn/generators/wagn/templates/config/routes.erb
ethn: I don't think we need to mention an application there. See this part from http://guides.rubyonrails.org/engines.html
````
Blorgh::Engine.routes.draw do
end
```

decko-rails/lib/decko/engine.rb
GerryG: By "never got clear on", I'm not so much talking about what you wrote, but what I've seen when I tried in and looked into how it is supposed to work.  As with most rails and ruby things, the code is better than the documentation, and lots of examples and descriptions are at best partly right.

Is the Decko::Engine.paths object the same or different from the MyApp::Application.paths object?  Same question for config.  I think they are necessarily different, but I'm also not sure that is right.

PR #332, opened 2015-02-21
Routes

GerryG:
You can merge what you want into your branch and close this one.  Mostly I want to make semaphore run on it as my config is currently broken.  Did you ever look into whether you can give me access so I can read the settings and files in the build?

### DIFF
--- a/decko-rails/lib/decko/engine.rb
+++ b/decko-rails/lib/decko/engine.rb
@@ -12,7 +12,6 @@ require 'kaminari'
 require 'diff/lcs'
 require 'diffy'
 
-DECKO_GEM_ROOT = File.expand_path('../../..', __FILE__)
 
 module Decko
 
@@ -27,7 +26,7 @@ module Decko
 
   end
 
-  class Decko::Engine < Rails::Engine
+  class Engine < ::Rails::Engine
     
     paths.add "app/controllers", :eager_load => true
     paths.add 'gem-assets',      :with => 'public/assets'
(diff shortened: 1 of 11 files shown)


decko-rails/lib/decko/engine.rb
```diff
@@ -27,10 +26,11 @@ def gem_root
 
   end
 
-  class Decko::Engine < Rails::Engine
+  class Engine < ::Rails::Engine
     
     paths.add "app/controllers", :eager_load => true
     paths.add 'gem-assets',      :with => 'public/assets'
+    paths.add 'lib/tasks',       :with => 'lib/decko/tasks', :glob => '**/*.rake'
```
GerryG: I didn't actually mean to add this.  It shouldn't hurt, right?  Unless it throws errors when the directory isn't there, but I doubt it does that.

wagn/lib/wagn/application.rb
```diff
@@ -84,14 +84,14 @@ def paths
         
         paths.add 'files'
         
-        paths.add 'local-mod', :with=>'mod'
+        #more consistent to call this deck mod...
```
GerryG: I guess I should remove the comment since the work is in here.

wagn/lib/wagn/config/routes.rb
```diff
@@ -0,0 +1,8 @@
+Wagn.application.routes.draw do
+
+  if !Rails.env.production? && Object.const_defined?( :JasmineRails )
+    mount Object.const_get(:JasmineRails).const_get(:Engine) => '/specs'
+  end
```
GerryG: This will fix the tests (I'm hoping).  It shouldn't be a problem for this to be here, since the file is now in the gem.

wagn/lib/wagn/generators/wagn/templates/config/routes.erb
```diff
@@ -1,6 +1,6 @@
 # -*- encoding : utf-8 -*-
 
-require 'decko/engine'
+#require 'decko/engine'
```
GerryG: I'm going to test this use case further on the generator branch.  It shouldn't be used in the *app deck* case anyway.  I think it is better if we don't need this, and this require should probably happen automatically via whatever is auto-included.

I don't know if you noticed, but I took away some of the eager load options in the config (as part of the initial card-gem split).  Best practice is not to do that for lib directories.

GerryG:
I guess I can just point this to your fork since you fixed the build issue.

PR #333, opened 2015-02-21
Rm local mod

GerryG:
This is the change to remove local-mod and consolidate these in mod.

### DIFF
--- a/wagn/lib/wagn/application.rb
+++ b/wagn/lib/wagn/application.rb
@@ -83,9 +83,15 @@ module Wagn
         Cardio.set_paths paths
         
         paths.add 'files'
-        
-        #more consistent to call this deck mod...
-        paths.add 'deck-mod', :with=>'mod'
+
+        if  mods = paths['mod']
+          if -d (deck_mods = "#{Wagn.root}/mod")
+            warn "mods #{mods.inspect}, #{deck_mods.inspect}"
+            mods << deck_mods
+          end
+        else
+          Raisl.logger.warn "Missing path for \'mod\'"
+        end
         paths['app/models'] = []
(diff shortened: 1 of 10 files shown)


wagn/lib/wagn/application.rb
```diff
@@ -83,9 +83,15 @@ def paths
...
-        #more consistent to call this deck mod...
-        paths.add 'local-mod', :with=>'mod'
+
+        if  mods = paths['mod']
+          if Dir.exists?(deck_mods = "#{Wagn.root}/mod")
+            warn "mods #{mods.inspect}, #{deck_mods.inspect}"
+            mods << deck_mods
+          end
+        else
+          Raisl.logger.warn "Missing path for \'mod\'"
```
ethn: typo!

wagn/lib/wagn/generators/wagn/templates/config/application.rb
```diff
@@ -43,7 +43,8 @@ class Application < Wagn::Application
     # config.paths['files'] = 'files'
     # where uploaded files are actually stored. (eg Image and File cards)
 
-    # config.paths['local-mod'] = 'mod'
+    # This probably needs to be a full path or it will be gem relative, which isn't really what we want here
+    # config.paths['mods'] << 'mod'
```
ethn: let's get rid of this comment (a template is a bad place for a code comment).

Also, I think it's incorrect.  by default it will be handled as deck relative (all gem paths must be made absolute via add_gem_path)

GerryG: D'oh, I thought I was in the lib/wagn/application.rb

Then it is wrong too :)

PR #334, opened 2015-02-22
moved decko::engine stuff into wagn

ethn:
This is an attempt to move most of the core wagn functionality from decko-rails and back into the wagn gem, but preserving Gerry's engine reorganization.

If we go with this model, decko-rails is going to be super thin. The idea is that decko-rails can be made to generate engine classes that inherit from the same engine class (currently Wagn::Engine, soon Decko::Engine again) that standard wagn apps use to mount their routes.

(In other words, it's basically what Gerry did before, but it refines decko-rails to be little more than an installation utility.)

Still more cleanup to do, but I hope this clarifies some of the recent architectural conversations.


GerryG:
I think you are making extra work here.  I don't think we should try to finalize this stuff until we are actually doing the final move from wagn to decko.  Why rename it to Wagn::Engine when it will be named back later?  I see no good coming from moving stuff back into the Wagn namespace.

GerryG:
To be clear, this is the position I am advocating.

1. If decko-rails exists in the long run, it is very thin.
2. The current split is the first appearance of decko as a codename, and as such the first step in the re-branding.  Wagn is still public branding and therefore the command UI and the original use case still use it.
3. Internally, we are shifting from Wagn -> Card and when Card isn't the right place Wagn -> Decko.  This can and should continue.
4. When it is time for the re-brand, the few things left in wagn move to decko, and  most of decko-rails moves as well.
5. Even if it is completely empty, we could keep decko-rails around anyway.  Maybe it is just the CLI to add it to an app and not even needed after that.  It is so thin, it won't really matter.

GerryG:
I'm ok if you move the files from one gem to the other, but don't change the ruby namespaces.

card/spec/lib/card/cache_spec.rb
```diff
@@ -76,7 +76,7 @@
 
   describe "with file store" do
     before do
-      cache_path = "#{Decko::Rails.root}/tmp/cache"
+      cache_path = "#{Wagn.root}/tmp/cache"
```
GerryG: This is a spec, but the deeper problem is that neither of those namespaces should be referenced in Card.  Truth is none of them should be referenced in the test and this should come from the class under test directly.  That may just move the problem there, but at least it will be in the right place.

decko-rails/decko.gemspec
```diff
@@ -22,11 +22,6 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 1.8.7'
 
-  [
-    [ 'rails',        '3.2.16'  ],
-    [ 'card',     version  ]
-  ].each do |dep|
-    s.add_runtime_dependency *dep
-  end
+  s.add_runtime_dependency 'wagn', version
```
GerryG: We are agreed on changing the direction of this dependency.

wagn/lib/wagn/application.rb
```diff
@@ -1,6 +1,6 @@
 # -*- encoding : utf-8 -*-
 
-require 'decko/engine'
+require 'wagn/engine'
```
GerryG: This is what I'm totally against.  Wagn::Engine is not something we want in the long run, so we shouldn't create it now.

Like I said in the general comment, move the files between gems, but don't change any names.

wagn/lib/wagn/application.rb
```diff
@@ -56,9 +56,13 @@ def add_gem_path paths, path, options={}
     def config
       @config ||= begin
         config = super
+
         Cardio.set_config config
+        config.autoload_paths += Dir["#{Cardio.gem_root}/mod/*/lib/**/"]
```
GerryG: Where was this before?  I think it needs to be with the component that triggers module loading on initialization or we have to make engine developers include it (or we repeat ourselves somewhere).

wagn/lib/wagn/application.rb
```diff
@@ -90,17 +94,12 @@ def paths
         paths['app/mailers'] = []
         
         add_gem_path paths, "lib/tasks",     :with => "lib/wagn/tasks", :glob => "**/*.rake"
-        add_gem_path paths, 'gem-assets'
-        add_gem_path paths, 'config/routes', :with => 'lib/wagn/config/routes.rb'
+        add_gem_path paths, 'config/routes', :with => 'rails/application-routes.rb'
```
GerryG: Interesting.  I like that better, what suggested it to you?

And, are you putting gem-assets with decko-rails, gem-wise?  I think that's right, it should be in a different place for initializing the path.

GerryG:
Now I get that you move the general stuff to rails/..., that's good either way you slice it.

If other web/service adapters can use it, they can get it from this path is whatever the eventual gem-name is by direct reference.  And we are agreed that wagn -> decko is where they belong.

ethn:
> To be clear, this is the position I am advocating.

I'm *totally* on board with those five steps. These make me feel like we're now very well aligned. Personally, I've been much more concerned about gem organization than namespaces.  Separating out Card was a massive architectural accomplishment.  Renaming Wagn to Decko will be a big undertaking, too, but my hope is that almost none of that change is *architectural*.  

I would suggest that in wagn/lib (the lib dir of the wagn gem), we can now add a decko dir and go crazy moving over anything that we can without breaking existing wagns.  Definitely move the engine there, because it's new.  We can start changing all kinds of stuff to Decko.  Migration classes, rake tasks....  we can even make Wagn::Application inherit from Decko::Application and put all the logic in the latter.  We can even create "decko" in the bin and make it do what `wagn` does. _Let the renaming begin!_

We just have to make sure we don't break existing decks (especially ones without mods; only a very few have mods, so we can handle updating those if needed).  But I think really, the only things existing decks use by name are: `Wagn::Application` and the cli commands.

Before, we were conflating the renaming with the gem structure, which was probably causing us a good bit of confusion.  I was thinking "name it what you want, so long as it's in the right gem", and you were thinking more along the lines of "put it where you want, so long as it has the right name".  I think we're both going to be happy now!

ethn:
> And, are you putting gem-assets with decko-rails, gem-wise?

There's basically nothing in decko-rails now.  wagn no longer depends on it.  it's just a palette ready for whatever adaptations are needed to get engine decks working.

I should have said this more explicitly, but your engine organization is awesome.  I just think the base classes belong in the wagn/decko gem.  What I did yesterday was generally very easy because you had already done the heavy lifting.

It's possible that some day Decko will not use Rails at all.  We likely would no longer have the railsy organization of engines and routes and so forth.  If that happens, then decko-rails will need to create these kinds of classes and use them to connect to Decko/Cards with more use of connector code.

But so long as Decko *is* railsy, it makes perfect sense for us to split out our controller/route handling into a separate engine and make it available for use/inheritance by decko-rails.  To me this is largely about what we're committing to.  I'd say we're making a very strong commitment to keeping things such that decko can be used as an engine in a rails context.  We're making no such commitment to decko depending upon rails gems directly.

GerryG:
Yes, and thanks.  I feel a lot more appreciated now :)

I'll finish up this branch for the most part.  This morning I was playing with putting decko/engine under the wagn/rails path, but if you're good with that, I'm putting it in wagn/lib/decko/ now and that will avoid some initialization pain.  Sorry for being dense about the gem organization, or I would have gone down that route a while back.  Good work moving lots of stuff to wagn/rails/

card/spec/lib/card/cache_spec.rb
ethn: right, this should be Cardio.root

decko-rails/decko.gemspec
ethn: woot!

wagn/lib/wagn/application.rb
ethn: yeah, it was in the engine and should probably be there.  Still several things to iron out here.  I especially want to figure out how to make it so cardio works without being fed a Railties-specific paths object.

wagn/lib/wagn/application.rb
ethn: > what suggested it to you?

you, basically ;)

PR #335, opened 2015-02-22
Engine with Decko::Engine

GerryG:
Hoping this works.  I'll make a PR to your version of this for comparison too.


GerryG:
https://github.com/ethn/wagn/pull/28

card/lib/cardio.rb
```diff
@@ -22,6 +22,8 @@ def set_config config
       @@config = config
       @@root = @@config.root
         
+      config.autoload_paths += Dir["#{Cardio.gem_root}/mod/*/lib/**/"]
+
```
ethn: yeah, this belongs here.

wagn/lib/wagn/railtie.rb
```diff
@@ -0,0 +1,12 @@
+
+require 'decko/engine'
```
ethn: given that engines inherit from Railtie, it's especially odd to do this.

GerryG: I'm not sure how those things relate.  I'm willing to go with another theory of what should require what.  I think this file becomes a bit more important with the latest fixes on the generator branch.

ethn: Engines depend on Railties, so they require railties.  

I'm not up for merging this railtie stuff in without seeing more rationale for this stuff.  I really think it's time to create an engine in a dummy rails app.  No doubt we'll learn a bunch.  Perhaps it will probe necessary to tweak railties, but at present these changes aren't even used (here or in the generator branch), so I can't really justify merging them in.

In the meantime, 1.15 is really being held up by performance and CSS issues, so there's really no rush, now that we're generally clear on namespaces and gem organization.

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -1,4 +1,5 @@
 
+require 'wagn/railtie'
 require 'rails/application'
```
ethn: any idea why this is rails/application and not wagn/application?

wagn/lib/wagn/railtie.rb
GerryG: Did you look at the test script on generator?  That's what I'm doing, but I'm having trouble running some of the steps I need to set-up and seed the database.

wagn/lib/wagn/all.rb
```diff
@@ -1,2 +1,2 @@
-# backwards compatibility
-require 'wagn/application'
\ No newline at end of file
+require 'decko/engine'
+require 'wagn/application'
```
ethn: this is wrong, no?

GerryG: I didn't thinks so.  This is the engine that is now in the wagn gem, but in Decko namespace.

I think we almost always need it because it configures (and probably causes to load) Card.  Are you saying it shouldn't always load with wagn/all.  I change the generator back, so 'compatibility mode' is the new mode too.

wagn/lib/wagn/generators/wagn/templates/config/application.rb
```diff
@@ -1,6 +1,6 @@
 require File.expand_path('../boot', __FILE__)
 
-require 'wagn/application'
+require 'wagn/all'
```
ethn: why?

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -1,4 +1,5 @@
 
+require 'wagn/all'
 require 'rails/application'
```
ethn: why not just `require 'wagn/application'`?

wagn/lib/wagn/all.rb
ethn: I thought we were getting rid of wagn/all in new code but just supporting it still to keep old wagns working.  So (1) why the change? and (2) it definitely has to require the application (when I added my first comment it didn't)

I'm not worried about namespaces, I'm worried about dependencies. For the sake of argument, let's just use "decko" here and talk about dependencies as they will be in the future. 

1. decko/application will depend on decko/engine.  So there's no need to require both like this unless there's a load order issue.  There shouldn't be.
2. In general, I would expect that most high-level things in the decko gem (specs, scripts, etc) would require the application and most things in decko-rails would require just the engine.   

I know that, the application is now an extremely thin wrapper (no controllers, no routes, etc), so we might be able to get away with requiring the engine instead, but that would run the risk of ignoring application-level configuration options in decks.  So we may want to set up some situations like that for testing purposes.

wagn/lib/decko/engine.rb
```diff
@@ -26,8 +26,8 @@ class Engine < ::Rails::Engine
       ActiveSupport.on_load(:active_record) do
         if defined? Wagn
           #this code should all be in Wagn somewhere, I suspect.
-          Wagn::Engine.paths['request_log'] = Wagn.paths['request_log']
-          Wagn::Engine.paths['log']         = Wagn.paths['log']
+          Engine.paths['request_log'] = Wagn.paths['request_log']
+          Engine.paths['log']         = Wagn.paths['log']
         else
           Cardio.card_config ::Rails.application.config
         end
```
ethn: another way to think about this:

engine apps will obviously need to require Decko::Engine and NOT Decko::Application.  This should give them access to the controller, routes, and of course, the world of cards.  What else do they need?

It seems like the main other thing is the world of rake tasks (and eventually the CLI wrappers around them, which we think will use a separate "decko-rails" method.  some more comments on that below.

In any case, I'm worried that we're losing a lot of the meat here, because so many of these good discussions are in PR's.  I would suggest we add two new places for bigger issues:
1. a ticket on Wagn.org
2. a README on the decko-rails gem.  That would be a great place to hash out some API stuff. 

In general, I hope that README will be the main document for learning how to add a decko engine to your rails app, so it's a good place to write up the interface!

wagn/lib/wagn/tasks/wagn.rake
ethn: so these rake tasks are probably the trickiest part here, because:
1. we only want to represent most of them in one place (somewhere in the decko gem), BUT
2. when called from an app deck we want to require decko/application (currently wagn/application), AND
3. when called from an engine deck we probably just want decko/engine

ALSO, our rake files are just generally messy and confusing.  So it's awesome that this work will inspire a major cleanup!!!

This may or may not be helpful, but there's the added complication that some of these tasks are really only intended for core developers (bootstrap stuff, for example) and really shouldn't show up at all.

I would offer these rough ideas:
1. it's ok with me if we go ahead and start namespacing these as "decko" rather than "wagn".  The word "rake" doesn't even appear on our standard README; all of these things are basically hidden API at this point.
2. one idea is that the require here is determined by an ENV variable.  eg, if ENV['ENGINE_DECK'] is true, then only decko/engine (and not decko/application) is required.  then we could set it up so that any time you do anything like `decko-rails seed`, it effectively calls `env ENGINE_DECK=true rake decko:seed`.
3. it's possible that this handling is done in decko/all, in which case I would understand why we want that file around!

ethn:
I added a bunch of (overly) long comments here.  I'll go ahead and merge to get the renaming in, but please give them a little attention (and perhaps help me get them represented in a wagn.org ticket).

btw, Philipp pointed out a case earlier today where I'd used "deck app" to mean "engine deck".  It was a screw-up, but now I understand why you thought I wanted that definition before, and it blows a hole in my whole "I never used it that way" defense ;)

ethn:
(we fixed it, btw, but it was there)

GerryG:
I'm working on an issue with tmp paths.  May have been broken since your merge before.  It seems to be putting them in the gem path.

ethn:
the application class is defined in the deck (inherits from the gem, but that doesn't matter), so its root is the deck root.  The engine is (currently) defined in the gem, so its root is the gem root.  If you're initializing Cardio with the non-application engine config, then it will not know where the deck root is.

GerryG:
So, the config object is the same everywhere because of railtie logic, but paths are different because the paths object has a root for relative paths, and that is the deck for the application (in both cases), and the wagn gem for the Decko::Engine.  Now I think I get why you might want to have an engine inheriting from Decko::Engine because it will have another paths, with it's own root (probably the deck for an engine deck).

So, do config objects inherit so that it can hit a particular config object for any attribute?  The first one it finds?  Does the paths object also do that when things are missing in one context?  I just noticed an example that suggests that paths['db/migrate'] is the equivalent of paths.db.migrate, so I guess if these are all inheritable things in this way it all would work.

That seems the right way to connect to Card for config and paths as well.  Is it ok to eventually have a card/lib/card/railtie ?  As long as it isn't automatically included when including the gem, right?

ethn:
are you sure the config is the same everywhere?  not sure that's true.  There may be some connection that happens through the mounting process?

> Now I think I get why you might want to have an engine inheriting from Decko::Engine because it will have another paths, with it's own root (probably the deck for an engine deck).

Right, that's one of many reasons.

> That seems the right way to connect to Card for config and paths as well. Is it ok to eventually have a card/lib/card/railtie ? As long as it isn't automatically included when including the gem, right?

So far I haven't heard any rationale for why that's needed.  I really really really really wish you'd not do any more restructuring until you've built an engine.

GerryG:
Not *the same everywhere*, but connected via railties.  I think the config and paths objects are associated with a Railtie, and there is an inheritance pattern there,  So, the rails application (a deck app, or a rails app hosting an engine deck) is a Railtie, and the engine (Decko::Engine) is a Railtie, and nothing in Card is a Railtie and instead we've hacked up something.  Each of these things is a distinct instance of a Railtie subclass, but it must do some method-missing logic when config.x is missing on mounted engines to go to the application.  I don't really know what that logic is, but you'd think that a random railtie (say for Card at some point) would inherit from engines and then from the application. but maybe there's just a bag of ties and it finds any of them.  Docs are not particularly clear and I've had trouble getting it to make sense.

I wasn't planning on trying it now, I've had enough of that for now.  Had I been successful, it would have been compelling, but I wasn't.  The why is as described above, which, I think, is what you had in mind when you first suggested I look into it.  I think the pattern where you use super to get the config or paths in one Railtie instance is fairly key in getting the pattern to work.

PR #338, opened 2015-02-23
select distinct

xithan:
Is there a case where we want the same card to appear several times in a search result? Or in other words: Do I have to extend the API for unique search results for example like  "return":"unique"?


xithan:
remark: this is needed for wikirate where we use searches like 
````
{
  "type":"_",
  "any":{
    "match":"$keyword",
    "right_plus":[ {},
      { "match":"$keyword" }
    ]
  }
}
````
that return the same card multiple times

ethn:
trying again in ethn:distinct

PR #339, opened 2015-02-23
Engine

GerryG:
Right now I just want this to build again.  I'm investigating whether the tmp stuff may be broken upstream and be writing in the gem.

### DIFF
--- a/card/lib/cardio.rb
+++ b/card/lib/cardio.rb
@@ -50,8 +50,8 @@ module Cardio
   
     def set_paths paths
       @@paths = paths
-      paths.add 'tmp/set',         :with => "#{root}/set"
-      paths.add 'tmp/set_pattern', :with => "#{root}/set_pattern"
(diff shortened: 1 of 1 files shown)


card/lib/cardio.rb
```diff
@@ -50,9 +50,8 @@ def set_config config
   
     def set_paths paths
       @@paths = paths
-      paths.add 'tmp/lib'
-      paths.add 'tmp/set'
-      paths.add 'tmp/set_pattern'
+      paths.add 'tmp/set',         :with => "#{root}/tmp/set"
+      paths.add 'tmp/set_pattern', :with => "#{root}/tmp/set_pattern"
```
ethn: I think this is synonymous to the previous, just noisier.

GerryG: Are you sure?  I guess it depends on the paths object's root, which I guess is the same thing as the deck/app root.  Well the tmp/lib one isn't needed.

PR #340, opened 2015-02-23
another stab at distinct subclauses

ethn:
basically just adds "DISTINCT" on any non-root card clause


card/lib/card/query/card_clause.rb
```diff
@@ -38,6 +38,8 @@ def initialize query
         @vars.symbolize_keys!
         @query = clean(@query)
         @rawclause = @query.deep_clone
+        
+        @sql.distinct = 'DISTINCT' if @parent
```
ethn: not beautiful, I know

PR #342, opened 2015-02-24
wikirate encoding issue

xithan:
this handles all these "invalid byte sequence in UTF-8" errors we are getting on wikirate

### DIFF
--- a/wagn/rails/controllers/card_controller.rb
+++ b/wagn/rails/controllers/card_controller.rb
@@ -73,6 +73,11 @@ class CardController < ActionController::Base
     Card::Cache.renew
     Card::Env.reset :controller=>self
     Card::Auth.set_current_from_session
+    
+    if params[:id] && !params[:id].valid_encoding?  # slightly better way to handle encoding issues (than the rescue in load_id)
+                                                    # we should find the place where we produce these bad urls
+      params[:id] = params[:id].force_encoding('ISO-8859-1').encode('UTF-8')
+    end
   end
 
 


wagn/rails/controllers/card_controller.rb
```diff
@@ -73,6 +73,11 @@ def per_request_setup
     Card::Cache.renew
     Card::Env.reset :controller=>self
     Card::Auth.set_current_from_session
+    
+    if params[:id] && !params[:id].valid_encoding?  # slightly better way to handle encoding issues (than the rescue in load_id)
+                                                    # we should find the place where we produce these bad urls
```
GerryG: We produce them or they are external?  What do we know about where they come from?

ethn: good question. we'll have to do some research on airbrake, I suppose.

xithan: Unfortunately airbrake can't tell us where link is coming from. The last errors are all produced by +link_and_compare addresses, for example http://wikirate.org/Amalgamated_Plantations_Pvt_Ltd_+Human_Rights+L_Or%e9al+link_and_compare. The remote ip belongs Hetzner Online, that's Wikirate's server host, right?

GerryG: I bet they are internal links.  Any ideas about where the third part of that name comes from in the code?  Is it a real cardname when you apply this patched code to it?  The percent encoding decoded should get you a UTF-8 codepoint, right?  UTF encodings are required for URLs, I would think.  I'm guessing it is actually the latin encoding in the patch.

GerryG:
You need to catch it when it is generating the bad links.

In some way, it seems like a risky patch.  The URLs are probably incorrectly encoded.  Workaround isn't bad, but if true, will cause trouble with other sites (potentially).

wagn/rails/controllers/card_controller.rb
xithan: the problem are characters like é as in "L'oréal". they are part of card names but I can't find the place where the bad links are build. And it has nothing to do with recent changes. These errors occasionally occurred for months.

GerryG: Could it be inside the smartname gem?

PR #343, opened 2015-02-25
Take some of the fixes from generator

GerryG:
MIgrations were still a bit broken on 1_15, but I'd been fixing it on generator already.  Took some stuff from that.
There's still an issue with on of the newer card migrations, so I just skip it for the full run.


wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -90,33 +90,31 @@ namespace :wagn do
...
 
   desc 'insert existing card migrations into schema_migrations_cards to avoid re-migrating'
   task :assume_card_migrations do
-    require 'card/migration'
-    Card::CoreMigration.schema_mode do
-      ActiveRecord::Schema.assume_migrated_upto_version Cardio.schema(:core_cards), Card::CoreMigration.migration_paths
-    end
+    require 'decko/engine'
+
+    Cardio.assume_migrated_upto_version :cord_cards
```
ethn: typo

ethn: could do:
````
rake wagn:assume_migrated:structure
rake wagn:assume_migrated:core_cards
rake wagn:assume_migrated:deck_cards
````
(doesn't have to be this PR)

File: wagn/lib/wagn/tasks/wagn.rake [hunk 2]
```diff
@@ -139,24 +137,22 @@ namespace :wagn do
...
       prepare_migration
-      paths = ActiveRecord::Migrator.migrations_paths = Card::Migration.migration_paths
+      paths = ActiveRecord::Migrator.migrations_paths = Cardio.migration_paths(:deck_cards)
     
-      Card::Migration.schema_mode do
+      Cardio.schema_mode(:deck_cards) do
         ActiveRecord::Migration.verbose = ENV["VERBOSE"] ? ENV["VERBOSE"] == "true" : true
         ActiveRecord::Migrator.migrate paths, ENV["VERSION"] ? ENV["VERSION"].to_i : nil
       end
     end
```
ethn: also not necessarily this PR, but it seems like we should be able to do this with less copy/paste

ethn:
looks like good stuff.  We don't have tests, of course.  Have you been testing the changes on the command line?

GerryG:
As I have commented a number or times, I have a script that is checked in (it should be in wagn/test/script, but I made a mistake initially and I'm not sure the change that puts it there is merged).  There are a couple support files, not checked in, but you can re-created pretty easily.  I note there is a 1.10 db dump in the repo, although I created my own and added a few more table drops so it won't leave any tables around that are created in the 1.10 -> current migrations.

In any case, I ran all the migrations and I have reported before that one of the migrations introduced recently is still broken, but when that one is skipped, they all run.  This work flow tests a bunch of the command line cases.  I should add a console and a server test.

ethn:
cool.  I knew about the script; I was just meaning to confirm that you'd run those tests with the above round of changes.  Merging (will fix typo)

PR #344, opened 2015-02-25
bootstrap css and js cards

xithan:
replaces also ui-icons with glyphicons
the migration changes are already merged into the fixtures and bootstrap files (the db bootstrap not the css bootstrap)


GerryG:
So then, glyphicons are a bootstrap resource?  As opposed to ui-icons that come from jQuery?

card/db/migrate_core_cards/data/more_bootstrap.json
```diff
@@ -0,0 +1,51 @@
+{
+  "url": "http://localhost:3000/export.json",
+  "timestamp": "2015-02-25 14:55:17 +0100",
+  "card": {
+    "name": "export",
+    "type": "Pointer",
+    "content": "[[style: glyphicons]]\n[[simple skin]]\n[[simple bootstrap skin]]\n[[Default Layout]]\n[[*all+*script]]\n[[*page header]]\n[[style: bootstrap compatible]]\n[[Old Layout]]",
+    "value": [
```
GerryG: Do we have a definition of this json cards format somewhere on http://wagn.org/ ?  Can we use contextual names in *subcards*?  I'm guessing on a pointer, this actually creates the card content from the pointer names, but presumably you can put children here, or not?

ethn: I don't think there's anything on wagn.org.  Agree that we need docs. IMO this kind of documentation we should mostly write inline in the code and get serious about generating docs from the code like rails does.  And of course we'll want to refer to it online.

I think the format's pretty solid.  content = raw db content.  value we use for processing the content in this case (generating a list of items from that). Can be done recursively.  We may add some extra detail, but I think most of what's there is exactly as we need it.

ethn: (but if you have other ideas, please add to idea/ticket)

ethn: in other words:
what we have: let's document in code
what we want: let's address in ideas/tickets

ethn: btw, this is the default json format.  if you have a pointer card named "export" and go to "export.json", this is what you get.

card/mod/05_standard/set/self/account_links.rb
```diff
@@ -20,8 +20,73 @@ def account_links
...
+      end.join "\n"
+    end
+  end
+  
+  view :list_navbar_right do |args|
+    render_list args.merge(:class=>"nav navbar-nav navbar-right")
+  end
+  
+  view :list_navbar_left do |args|
+    render_list args.merge(:class=>"nav navbar-nav navbar-left")
```
ethn: I really like the idea of separating these into separate views!  I want to ponder a bit whether they should be on _\*account links_, just because writing {{\*account links|sign_out}} is a lot to remember, but this feels like it's clearly the right direction. 

Even if we want to attach these views to a different card, we will definitely need to make sure *account_links still works (for backwards compatibility).  And we don't really have to announce these new views to anyone right away if we haven't decided where we want them yet (for forwards compatibility).

ethn: we could consider handling some of the if's within the view syntax, eg:
````
view :sign_out, :perms=>lambda { |r| Auth.signed_in? }, :denial=>:blank do |args|
...
````
if you do that you want `optional_render` rather than `_optional_render`, which skips the permissions. (hoping to get rid of that dangerous syntax some day!)

File: card/db/migrate_core_cards/data/more_bootstrap.json [hunk 3]
```diff
@@ -0,0 +1,51 @@
...
+      {
+        "name": "style: glyphicons",
+        "type": "SCSS",
```
ethn: this is normal CSS right?  Any reason not to do it as type CSS?  Not a big deal, probably, other than the hundredths of a second spent processing it when there are style changes...

card/db/migrate_core_cards/data/more_bootstrap.json
```diff
@@ -0,0 +1,51 @@
...
+      },
+      {
+        "name": "simple skin",
+        "type": "Skin",
+        "content": "[[style: jquery-ui-smoothness]]\n[[style: functional]]\n[[style: standard]]\n[[style: right sidebar]]\n[[style: common]]\n[[style: traditional]]\n[[style: glyphicons]]\n[[style: bootstrap compatible]]"
+      },
```
ethn: fwiw, I was kind of reworking "style: traditional" to be, effectively "bootstrap compatible".  This might be a better approach, though...

card/db/migrate_core_cards/data/more_bootstrap.json
```diff
@@ -0,0 +1,51 @@
...
+      {
+        "name": "simple bootstrap skin",
+        "type": "Skin",
+        "content": "[[style: jquery-ui-smoothness]]\n[[style: functional]]\n[[style: standard]]\n[[style: right sidebar]]\n[[style: common]]\n[[style: bootstrap]]\n[[style: bootstrap theme]]\n[[style: bootstrap cards]]"
+      },
+      {
+        "name": "Default Layout",
+        "type": "Layout",
+        "content": "<!DOCTYPE HTML>\r\n<html>\r\n  <head>\r\n    {{*head}}\r\n  </head>\r\n  \r\n  <body>\r\n    <div id=\"menu\">\r\n      {{*page header}}\r\n    </div>\r\n    \r\n    <div id=\"primary\">\r\n      {{_main}}\r\n    </div>\r\n    \r\n    <div id=\"secondary\">\r\n      {{*sidebar}}\r\n    </div>\r\n  </body>\r\n</html>"
+      },
```
ethn: I love this.  worth discussing how much it would break to make the names and classes line up, but this is wonderfully simple!

card/db/migrate_core_cards/data/more_bootstrap.json
```diff
@@ -0,0 +1,51 @@
...
+      {
+        "name": "style: bootstrap compatible",
+        "type": "SCSS",
+        "content": "#menu {\r\n  height: 30px;\r\n  div.container-fluid{\r\n    display: block;\r\n  }\r\n  div.navbar-header {\r\n    display: none;\r\n  }\r\n  ul {\r\n    font-size: 14px;\r\n    margin-top: 5px;\r\n  }\r\n  li {\r\n    display: inline;\r\n  }\r\n  .navbar-left {\r\n    float: left;\r\n  }\r\n  .navbar-right {\r\n    float: right;\r\n    margin-right: 15px;\r\n  }\r\n  .navbar-form {\r\n    width: 30%;\r\n    margin-top: 5px;\r\n    margin-left: 10px;\r\n    input {\r\n      width: 100%;\r\n    }\r\n  }\r\n}\r\n\r\n#primary {\r\n  clear: both;\r\n}"
+      },
+      {
+        "name": "Old Layout",
+        "type": "Layout",
```
ethn: For this one I was thinking to actually take their default layout and rename it (so that they still have whatever they've changed).

I'm pretty nervous that we're going to break pretty much any site that has done a lot of customizations.

xithan: :heavy_check_mark:

card/db/migrate_core_cards/data/more_bootstrap.json
xithan: :heavy_check_mark:

card/mod/05_standard/set/self/account_links.rb
xithan: > we could consider handling some of the if's within the view syntax  :heavy_check_mark: 

So you vote for cards *sign_out, *sign_in, *sign_up ... instead of the views?

xithan:
> So then, glyphicons are a bootstrap resource? As opposed to ui-icons that come from jQuery?
correct

GerryG:
If I understand what this branch is doing, we are moving from our own classes to more *standard* ones from bootstrap.

My question is, don't we want to make it relatively easy to support other frameworks too?  I only ask because Julien mentions angular.js in his email.  I'm just guessing, but angular and bootstrap are competing to be that standard to some extent.

card/mod/05_standard/set/self/account_links.rb
ethn: I'm kind of thinking they should be views on *account but rendered on *account links.

(but this is fine for now)

ethn:
Short answer: we should definitely be able to support lots of styling
variants through mods.  We'll probably need to make a bit deeper commitment
to a js infrastructure, but it wouldn't be that hard to build variants
there either.

The main motivation for bootstrap is to improve our styling rapidly in time
for Decko 1.0. Bootstrap will give us lots of clean options and a ton of
alternate skins right away. And it's mostly organized around jquery, Wagn's
current js system, which means there's no major js overhaul involved.

With the exception of our (previously disorganized) top menu, we're not
really moving away any our core classes; we're mostly just adding bootstrap
classes.  The main pattern is making it so that our card frames are panels.
The mapping is pretty direct:

.card-frame += .panel
.card-header += .panel-heading
.card-title +=~ .panel-title
.card-body += .panel-body

In the process, we're moving from the jquery-ui icons (implemented via
image maps) to glyphicons (implemented via fonts/svg).  The latter is a
much better fit, because it handles dynamic sizing and coloring.

Re angular.js, at its core it's more of an alternative to jquery than
bootstrap, though, of course, there are a lot of angular ui things that get
into bootstrap's core space.

Philipp recently found this angular/bootstrap integration:
http://angular-ui.github.io/bootstrap/  That may indeed be a great route
for Decko in the future.  As I say, we'd probably explore it first through
mods but at any point would need to commit to one system for default
installations.



On Thu, Feb 26, 2015 at 10:21 AM, Gerry Gleason <notifications@github.com>
wrote:

> If I understand what this branch is doing, we are moving from our own
> classes to more *standard* ones from bootstrap.
>
> My question is, don't we want to make it relatively easy to support other
> frameworks too? I only ask because Julien mentions angular.js in his email.
> I'm just guessing, but angular and bootstrap are competing to be that
> standard to some extent.
>
> —
> <https://github.com/wagn/wagn/pull/344#issuecomment-76222275>.
>



-- 
Ethan McCutchen
One of the Wagneers, Wagn.org

Wagn. How pioneers roll.

s: ethan.mccutchen
t: @intogreater

GerryG:
Thanks, I don't really have a preferred framework, just wondering what the direction is.  The bootstrap/angular thing seems like a good thing.  These frameworks often cover different subsets of the whole space, and using one styles framework with a different widgets toolkit will be a common use case.

I think I mentioned that we are starting to use ReactJS: http://facebook.github.io/react/
That one might be worth learning more deeply and applying at some point.

Some of your answer probably should be somewhere at http://wagn.org

PR #346, opened 2015-02-26
more migration and generator fixes

ethn:


card/lib/card/migration.rb
```diff
@@ -81,18 +76,6 @@ def contentedly &block
...
-        path_to_file = File.join path, 'data', filename
-        return path_to_file if File.exists? path_to_file
-      end
-    else
-      File.join migration_paths.first, 'data'
-    end
-  end
-
-
   def import_json filename
```
ethn: this stuff just gets overwritten a couple of lines down.

card/lib/card/migration.rb
```diff
@@ -53,17 +53,12 @@ def assume_migrated_upto_version
...
-          path_to_file = File.join path, 'data', filename
-          return path_to_file if File.exists? path_to_file
-        end
-      else
-        File.join self.migration_paths.first, 'data'
-      end
+      path = migration_paths.first
+      File.join( [ migration_paths.first, 'data', filename ].compact )
     end
+
```
ethn: needs to return path even when file doesn't exist (eg in card migrations)

PR #349, opened 2015-02-28
Getting Engine Decks to Work a Little

GerryG:
I will probably remove the 'card' generator.  I still think it is a better idea for the generators to be in the card namespace.  We'll see if this still works for: rails/wagn g --help

At least now we have only a few smaller issues to argue about.


card/lib/card/migration.rb
```diff
@@ -1,7 +1,5 @@
 # -*- encoding : utf-8 -*-
-
-require 'card/version'
```
GerryG: The schema support stuff isn't in version any more.

card/lib/generators/card.rb
```diff
@@ -0,0 +1,43 @@
+# -*- encoding : utf-8 -*-
```
GerryG: I will remove this if we really don't need it.  It does seem like mostly we just need to have the documentation that tells them what they need to do.  At the very least, I've agreed that it doesn't make sense to think of the step to add cards to an app be a migration generation, but it will need to run something that acts like a migration in that it adds the necessary models to db/schema.rb.

I think the rake tasks for loading the seeds should work to initialize the wagn data after adding to the schema.

card/lib/card/cache.rb
```diff
@@ -40,6 +40,11 @@ def renew
         reset_local
       end
 
+      def prefix_root
```
GerryG: This was done to help with some issues where Cardio.config wasn't initialized when this was loaded.  I'm not sure what use case it was, but it's just a lazy load so it doesn't have to be initialize until it is used vs. when defined.

GerryG:
The broader testing worked.  The migration script ran through to completion and even the one that used to break didn't pop up.  That, and the problem the cropped up with the deliver error on setup did not occur.  I noticed the messaging on setup, I take it that you fixed that by disabling deliveries on generation of the app in order to fix this?

Console worked too.  I will also run the tests on the migrated database because I've seen problems with that when the regular test process worked.

GerryG:
There are some problems.  The is the test process, run the migration test so that you have a test database loaded from 1.10 and migrated to the current version.  Run specs or cucs.

Expected result is that they will run correctly, but they don't.

A second issue that I have noticed before, but just explicitly tested.  When I run "wagn seed -d" it drops the test database too.  I suppose that could be related to the change in the rake tasks, so maybe I should check this on master too.  I'm still a bit unclear about when you need to depend on :environment.  My current understanding is that it causes the app to be loaded, but in confusion I was thinking it related to getting the right RAILS_ENV or something.

GerryG:
Output of: wagn g --help

Card:
  card

CardMigration:
  card_migration

Cucumber:
  cucumber:install

EmailSpec:
  email_spec:steps

Format:
  format

Jasmine:
  jasmine:examples
  jasmine:install

JasmineRails:
  jasmine_rails
  jasmine_rails:install

Jquery:
  jquery:install

Js:
  js:assets

Kaminari:
  kaminari:config
  kaminari:views

Paperclip:
  paperclip

Rspec:
  rspec:controller
  rspec:feature
  rspec:helper
  rspec:install
  rspec:integration
  rspec:job
  rspec:mailer
  rspec:model
  rspec:observer
  rspec:scaffold
  rspec:view

Set:
  set

Note that the card related migrations are not together.  Paperclip, even with only one generator, has a namespace.  They need to be in a namespace.   This branch is merged with the arrangement Ethan created in fixing the migrations before merging to master, but I have kept the namespaced branch so I could re-merge master into that one.

Ethan, you owe me a response on this question.  I think the output above shows why my version is the correct solution.

GerryG:
This is what you see in the namespaced version:

Card:
  card
  card:format
  card:migration
  card:set

card/lib/card/cache.rb
```diff
@@ -18,10 +18,10 @@ class Cache
     TEST_ENVS = %w{test cucumber}
     @@prepopulating     = TEST_ENVS.include? Rails.env
     @@using_rails_cache = TEST_ENVS.include? Rails.env
-    @@prefix_root       = Cardio.config.database_configuration[Rails.env]['database']
+    @@prefix_root       = nil
```
ethn: this line isn't needed then, right?

card/lib/cardio.rb
```diff
@@ -50,26 +48,19 @@ def set_config config
...
       @@paths = paths
-      paths.add 'tmp/lib'
-      paths.add 'tmp/set'
-      paths.add 'tmp/set_pattern'
-      paths.add 'db/migrate_deck_cards', :with=>'db/migrate_cards'
 
-      add_gem_path 'mod',      :with => 'mod'
-      add_gem_path "db"
+      add_gem_path 'tmp/set',               :root => root, :with => '/tmp/set'
+      add_gem_path 'tmp/set_pattern',       :root => root, :with => '/tmp/set_pattern'
```
ethn: ??  why the heck would you call add_gem_path to add a deck path?  please restore

card/lib/cardio.rb
```diff
@@ -50,26 +48,19 @@ def set_config config
...
+      add_gem_path 'mod'
+      add_gem_path 'db'
       add_gem_path 'db/migrate'
-      add_gem_path "db/migrate_core_cards"
-      add_gem_path "db/seeds", :with => "db/seeds.rb"
+      add_gem_path 'db/migrate_core_cards'
+      add_gem_path 'db/migrate_deck_cards', :root => root, :with => 'db/migrate_cards'
```
ethn: yuck.

card/lib/cardio.rb
```diff
@@ -50,26 +48,19 @@ def set_config config
...
-      paths.add 'tmp/set'
-      paths.add 'tmp/set_pattern'
-      paths.add 'db/migrate_deck_cards', :with=>'db/migrate_cards'
 
-      add_gem_path 'mod',      :with => 'mod'
-      add_gem_path "db"
+      add_gem_path 'tmp/set',               :root => root, :with => '/tmp/set'
+      add_gem_path 'tmp/set_pattern',       :root => root, :with => '/tmp/set_pattern'
+      add_gem_path 'mod'
+      add_gem_path 'db'
```
ethn: is paths['db'] ever called?

card/lib/generators/card/USAGE
```diff
@@ -0,0 +1,21 @@
...
+  The Decko::Engine routes will be mounted at /cards/
+
+  wagn generate card base
+
+Creates:
+  config/new_routes.rb
+  db/migration/2015010112000000_create_cards_tables.rb
+
+  The table names will not have a prefix (i.e. cards, card_references ...)
+  The Decko::Engine routes will be mounted at /
```
ethn: > wagn generate card new cards pre_fix

That's inscrutable.  The syntax is subject, verb, object. `rails generate migration` tells *rails* to *generate* a *migration*.  `wagn generate card` tells wagn to generate a card.  That's not what's happening. here.

more importantly, this is all decko-rails stuff and should be there.

card/lib/generators/card.rb
ethn: generators (and all code) for the decko-rails scenario should go in the decko-rails gem unless there's a solution that involves deeper code sharing.

card/lib/generators/card/templates/create_cards_tables.erb
```diff
@@ -0,0 +1,106 @@
...
+
+    add_index "<%= prefix %>schema_migrations_core_cards", ["version"], :name => "<%= prefix %>unique_schema_migrations_cards", :unique => true
+
+    create_table "schema_migrations_deck_cards", :id => false, :force => true do |t|
+      t.string "version", :null => false
+    end
+
+    add_index "<%= prefix %>schema_migrations_deck_cards", ["version"], :name => "<%= prefix %>unique_schema_migrations_deck_cards", :unique => true
+  end
+end
```
ethn: we can delete this file, right?

card/lib/generators/card_migration/USAGE
```diff
@@ -15,7 +15,7 @@ Example:
     ...which is set up to use methods in lib/card/migration.rb
     
     
-    wagn generate card_migration import_better_search_cards
+    wagn generate card:migration import_better_search_cards
 
     This will create:
```
ethn: this makes sense.

decko-rails/lib/decko/rails.rb
```diff
@@ -1,11 +1,33 @@
...
     class << self
       def gem_root
         DECKO_RAILS_GEM_ROOT
       end
     end
   end
-end
\ No newline at end of file
+
+  if defined? ::Rails::Railtie
```
ethn: This is the check to see if it's being called from another app?

wagn/lib/decko/engine.rb
```diff
@@ -12,27 +12,39 @@
...
       ActiveSupport.on_load(:active_record) do
-        if defined? Wagn
-          #this code should all be in Wagn somewhere, I suspect.
-          Engine.paths['request_log'] = Wagn.paths['request_log']
-          Engine.paths['log']         = Wagn.paths['log']
-        else
-          Cardio.card_config ::Rails.application.config
-        end
-        
+        Cardio.cache == ::Rails.cache
```
ethn: Do we need this twice in three lines?

wagn/lib/wagn/commands.rb
```diff
@@ -36,19 +36,27 @@ def find_spec_file filename, base_dir
   end
 end
 
+LOAD_TASKS= {
+  'reseed' => 'reseed',
+  'seed' => 'create',
+  'load' => 'seed'
+}
```
ethn: these will be gone soon, right?

ethn:
Have you gotten an engine working yet?

card/lib/cardio.rb
ethn: oh, this is old.  I still wonder, but not your problem ;)

wagn/lib/decko/engine.rb
```diff
@@ -12,27 +12,39 @@
...
-    
+    paths.add 'lib/tasks',       :with => "#{::Wagn.gem_root}/lib/wagn/tasks", :glob => '**/*.rake'
+    paths.add 'lib/wagn/config/initializers',
+              :with => File.join( Wagn.gem_root, 'lib/wagn/config/initializers' ), :glob => "**/*.rb"
+
+    initializer 'decko.engine.load_config_initializers',  :after => :load_config_initializers do
+      paths['lib/wagn/config/initializers'].existent.sort.each do |initializer|
+        load(initializer)
+      end
+    end
```
ethn: are you sure engine decks will want all the application initializers?

It's very hard to get enthusiastic about merging this if none of it has been tested in an engine context.

card/lib/generators/card/USAGE
GerryG: I won't argue it now, I thought I commented that this one will very likely be just deleted.

card/lib/generators/card/templates/create_cards_tables.erb
GerryG: I think the whole migration will be removed.  I more or less agree with you, but don't have the solution yet.  I don't need it that bad at this point anyway.

decko-rails/lib/decko/rails.rb
GerryG: I think this file won't even be loaded in the regular wagn app case.  This file has to trigger any glue logic that is needed for the engine deck case.  In that case, I think this better be defined by now or it won't work, so maybe it isn't even needed.

ethn: Yeah, I thought I realized my question made no sense and withdrew it.

I think there may be cases where you run "bundle exec blah" on a rails app and don't load the application, in which case this kind of check may be necessary.

perhaps the file itself should be at decko/rails/railtie (at some point?)

wagn/lib/decko/engine.rb
ethn: I take that back.  I'm enthusiastic but fuzzy. It feels like the default engine setup will be the easiest point of agreement.  Then, after a closer look at the engine docs, I suspect I may have a clearer idea of where we may be thinking differently.

With that reading it looks to me like the standard pattern would be to name the engine Decko::Rails::Engine, no?  Perhaps we define that in decko-rails and have it inherit from Decko::Engine?

One major question I have is where we'll be overriding engine configuration.  In initializers?

GerryG: I'm testing it.  And I appreciate the feedback.

You may be right about these initializers.  I think there are only like two on the wagn side now.

wagn/lib/decko/engine.rb
GerryG: No, one of them is the right one, I'll delete the other.  And don't worry, I'll retest all the major scenarios once it is all stable.

wagn/lib/wagn/commands.rb
GerryG: Sure, but what should we use?  Removing it completely implies that we need to change the rake tasks around a bit, is that ok with you?  (what is now wagn:create becomes wagn:seed and current wagn:seed becomes wagn:load).  I think that will be fine.  As long as I make sure the descriptions make sense too.

decko-rails/lib/decko/rails.rb
GerryG: Sure, if either of them get any more complex.  With this little logic, it hardly seems worth it.   

And yes about why this might be needed and when.

wagn/lib/wagn/commands.rb
ethn: yes, we can change the task names so long as we support the same core CLI commands (wagn new, wagn update, etc).  you can even use decko for the task namespace afaic.

GerryG: That sort of opens a bigger can of worms.  Is there a justification for tasks being defined in wagn/lib/wagn/tasks/X.rake vs. wagn/lib/tasks/X.rake ?  Can you make the case for any of them being there besides the obvious wagn.rake?

To change the namespace from wagn to decko, I would move wagn/lib/wagn/tasks/wagn.rake to wagn/lib/tasks/decko.rake along with the rest of the *.rake in the directory (same names).  Would that work for you?  I looked around in the gems directory too and there are really no examples of doing it like we currently have it.

I think it would be good to move this stuff to the new name.  I have no attachment to doing it now, but I think it goes well with what we'll have left  on this crazy branch.

GerryG:
The branch status is 'still being tested on the new use cases, otherwise basically working'.

The test failures on the migrated test database are likely in master too, and maybe not even related to any of this work.  Do you want me to verify that theory and run the test on master?

card/lib/card/cache.rb
GerryG: I think otherwise you get not-initialized errors for @@ vars

card/lib/generators/card.rb
GerryG: Actually, this file is used.  It just supplies some generator classes that are used in the various generators.

decko-rails/lib/decko/rails.rb
GerryG: Actually, the reason for it landing here even if there's nothing in Decko::Rails is that this file is the standard require for the name.  We don't have to even define that constant, but since this is the included file I thought it fine to put Railtie stuff here.

wagn/lib/wagn/commands.rb
```diff
@@ -9,15 +9,15 @@ def load_rake_tasks
...
-  "r"  => "runner"
+  'rs' => 'rspec',
+  'cc' => 'cucumber',
+  'jm' => 'jasmine',
+  'g'  => 'generate',
+  'd'  => 'destroy',
+  'c'  => 'console',
+  's'  => 'server',
+  'db' => 'dbconsole',
+  'r'  => 'runner'
```
GerryG: In case you're wondering.  The idea is 'strength reduction' for string constants.  If you don't have any double quote syntax, why incur any costs calling methods to parse.  Not that it is a good idea to be rigid about it or go through and change them all, but I think it's good as a standard style.

card/lib/card/cache.rb
ethn: not for `||=`, I don't think.  no big deal

card/lib/cardio.rb
```diff
@@ -1,6 +1,5 @@
 # -*- encoding : utf-8 -*-
 
-require 'rails'
```
ethn: impressive.  not needed?

card/lib/cardio.rb
```diff
@@ -50,26 +48,19 @@ def set_config config
   
     def set_paths paths
       @@paths = paths
-      paths.add 'tmp/lib'
-      paths.add 'tmp/set'
-      paths.add 'tmp/set_pattern'
-      paths.add 'db/migrate_deck_cards', :with=>'db/migrate_cards'
+      add_gem_path 'tmp/set', :root => root
+      add_gem_path 'tmp/set_pattern', :root => root
```
ethn: I get why you're doing this, but it renders the name `add_gem_path` very odd.

The issue is that if you only define the engine in the gem, then the paths root is in the gem for engine decks and in the app for app decks.  This means we really need to be explicit with every root.  (I had been thinking the engine would have a class for configuration in the deck, which would make the root there)

Perhaps we need to change the method to add_path and handle the engine v deck ambiguities there.

wagn/lib/decko/engine.rb
ethn: == probably not intended

wagn/test/script/run_engine_deck.sh
```diff
@@ -0,0 +1,47 @@
...
+#echo "Load seed db (enter the pw)"
+#mysql -u decko_user -p mdummy_test < ${SEEDSQL}
+# copy db config (populate .local with you user/pw)
+cp ${DBCFG} config/database.yml
+RAILS_ENV=test bundle exec rake db:create --trace
+echo "seed database"
+RAILS_ENV=test bundle exec rake wagn:seed --trace
+#echo "migrating"
+#RAILS_ENV=test bundle exec rake db:migrate --trace
+#RAILS_ENV=test bundle exec rake wagn:migrate --trace
```
ethn: does this work?

card/lib/card/cache.rb
GerryG: Now that I think about it, it does make sense.  All the assignment operators will create the lvalue (a data container) first.  Then it doesn't throw the error because even though it is unassigned, it exists.

wagn/lib/decko/engine.rb
GerryG: Good point.  That must break something.

card/lib/cardio.rb
GerryG: That seems about right.  I didn't think I added the handling of the root option.

You are saying rename add_gem_path because it is more general, not having two methods, right?

wagn/test/script/run_engine_deck.sh
GerryG: I'll get back to you on that as I retest both this and the standard migration scenario.   Parts of it have all worked sporatically.

card/lib/cardio.rb
ethn: I added root so that it could handle different gem_roots (at the time there were places where we were referring to different gem roots in at least one of these).

but my point is that there are gem paths (paths to something in the gem dirs) and deck paths (paths to something in the deck dirs).  And using add_gem_path for adding deck paths makes no sense.

GerryG: That part I understood, the question is what did you want in the fix.  I think you just want the rename, not two methods, so that's what I'll do.

Actually, doesn't this method name pre-date the split?  Oh, you added the root option, right, then I abused it :)

GerryG: It seems the only use cases for the option are referring to the deck root.  I'm sure there were other cases as one point, but no longer.  Is it worth making the add_path call the same in Cardio and Wagn?  In the latter, there is a paths argument, but not in the former.  I have an idea of how to do that even though it is called in the paths method.

GerryG:
Once the other two are merged, I will restart this one under a different branch name and without the hundreds of mostly merged commits.  Then this will really be focussed on getting the engine application case working and with a clean separation from the main case.

It won't be very big anymore.

wagn/lib/wagn/commands.rb
ethn: yeah, I like for us to do this where we can; thanks for doing it here.  (I also like that single quotes are precisely half as visually noisy!)

ethn:
Do I understand correctly that you're planning to redo this PR?

GerryG:
Yeah, that's the plan.  I'll close this one for now.  Mostly I'm just going to compress it into one commit, then work on it some more until it is working mostly.

PR #350, opened 2015-03-02
basic profiling support with ruby-prof

ethn:
basic process to get profiling results:
1. regenerate core-dev deck, eg `wagn new core-dev --core-dev; cd core-dev; bundle install`
2. start server in profile environment, eg `env RAILS_ENV=profile wagn s`
3. make a request to that server, eg `curl http://localhost:3000`
4. check out the log files in tmp/profile

I found the html results useful.  I haven't gotten into the kcachegrind stuff yet (see ruby-prof docs), but I did see somewhere that you can install a mac variant call qcachegrind with homebrew.


ethn:
I'd really like to get semaphore tests down under 10min total.  That's not the main reason I want to improve performance, of course, but for lack of other specific goals, it's a nice one to start with to help speed up dev cycles.

wagn/lib/wagn/generators/wagn/templates/config.ru
```diff
@@ -1,5 +1,11 @@
 # This file is used by Rack-based servers to start the application.
 
 require ::File.expand_path('../config/environment',  __FILE__)
+<% if options['core-dev'] -%>
 
+if Rails.env.profile?
+  use Rack::RubyProf, :path => 'tmp/profile'
+end
```
ethn: I suspect we can find a better place for this...

wagn/lib/wagn/generators/wagn/templates/config/databases/mysql.yml
```diff
@@ -1,11 +1,12 @@
...
+
+production: 
   adapter: mysql2
   encoding: utf8
   reconnect: false
-  database: <%= app_name %>_development
+  database: <%= app_name %>_production
   pool: 5
   username: root
   password:
```
ethn: moved production to the top, because it's the main db for folks creating new wagn apps (unlike those creating new rails apps).  Would like to simplify this further...

PR #351, opened 2015-03-02
make mod tests working again

xithan:


wagn/lib/wagn/mods_spec_helper.rb
```diff
@@ -1,6 +1,5 @@
 # -*- encoding : utf-8 -*-
 require 'spork'
-raise "is this loaded anywhere?"
```
xithan: > raise "is this loaded anywhere?"
yes, this is loaded in the spec_helper.rb file that is generated for mod-dev decks (for example like wikirate - this was the reason why the tests on semaphore failed for wikirate)

GerryG: D'oh.

Should we be removing that for the future?

ethn: removing what?  the raise?  it's gone now.

GerryG: The place where this is required.  I think it may have gotten forked into a file in card and wagn gems.  Probably we don't need two, except for some legacy cases maybe?  Or is it really needed as a distinct thing for mod-dev decks.  Maybe the template would change not to need it.

ethn: it's required by specs on decks.

GerryG: Yes, I realize that.  The question I'm asking is whether the spec template could refer to the probably almost identical file in the card gem.  There may be a good reason for the split, but more likely it just happened with the gem split.

PR #352, opened 2015-03-02
rename follow option cards

xithan:


card/card.gemspec
```diff
@@ -27,6 +27,7 @@ Gem::Specification.new do |s|
     [ 'smartname',    '0.2.3'   ],
 
     [ 'uuid',         '~> 2.3'  ],
+    [ 'cocaine',      '~> 0.5.5'], # used by paperclip, version 0.5.6 is broken
```
ethn: shall we take this back out now that 0.5.7 is out?

PR #355, opened 2015-03-03
pagination with bootstrap

xithan:


wagn/lib/kaminari/helpers/tags.rb
```diff
@@ -0,0 +1,130 @@
...
+
+      def to_s(locals = {}) #:nodoc:
+        @template.render :partial => partial_path, :locals => @options.merge(locals), :formats => [:html]
+      end
+
+      def page_url_for(page)
+        p = params_for(page).merge(:only_path => true)
+        #view"=>"history", "controller"=>"card", "action"=>"read", "format"=>"html", "id"=>"Home", "page"=>"1",
+        "#{p['id']}?view=#{p['view']}&page=#{p['page']}"
+      end
```
xithan: This file is a copy from the kaminari gem. The url_for method doesn't work in the original code. Maybe it has something to do with our new engine routes and we can get rid of this when the PR https://github.com/amatsuda/kaminari/pull/636 gets merged.

ethn: Any chance we can just override the one method?  Perhaps we just stick something like this in card/core_ext.rb (or similar)

````
class Kaminari::Helpers::Tag
  def page_url_for page
     p = params_for(page)
     p.delete :controller
     p.delete :action
     # can these just get left out?
     card = Card[p.delete('id)]
     card.format.path  p
  end
end
````
(sorry for the late response on slack)

Comment (xithan) [inline wagn/lib/kaminari/helpers/tags.rb]: > Any chance we can just override the one method?
tried that first but then all other methods of Kaminari::Helpers::Tag are missing.

Comment (xithan) [inline wagn/lib/kaminari/helpers/tags.rb]: I had to override params_for too. Now it's working

File: wagn/lib/decko/engine.rb [hunk 2]
```diff
@@ -34,6 +35,13 @@ class Engine < ::Rails::Engine
...
         ActiveRecord::Base.establish_connection(::Rails.env)
       end
+      ActiveSupport.on_load(:after_initialize) do
+          begin
+           require_dependency 'card' unless defined?(Card)
+          rescue ActiveRecord::StatementInvalid => e
+            ::Rails.logger.warn "database not available[#{::Rails.env}] #{e}"
+          end
+      end
     end
```
xithan: put that back because rake tasks break without it

PR #356, opened 2015-03-04
fixes to caching, reloading, following...

ethn:


card/lib/cardio.rb
```diff
@@ -18,6 +18,10 @@ module Cardio
   mattr_reader :paths, :config, :cache
 
   class << self
+    def cache
+      @@cache ||= ::Rails.cache
+    end
+
```
ethn: I get that we don't really want this dependency long term, but for now this avoids problems.

card/mod/03_machines/lib/javascript/wagn_menu.js
```diff
@@ -64,7 +64,8 @@ wagn.menu_template = [
       { "text":"account", "related":{"name":"+*account", "view":"edit"}, "text":"details" },
       { "related":"roles"   },
       { "related":"created" },
-      { "related":"edited"  }
+      { "related":"edited"  },
+      { "related":"follows" }
```
ethn: adds an account > follows item to menu (on cards with accounts)

card/mod/05_email/set/right/following.rb
```diff
@@ -13,21 +13,40 @@ def virtual?; true end
...
+       rcard.item_cards.map do |item|
+         %{<div class="alert alert-success" role="alert">
+           <strong>#{item.title}:</strong> #{rcard.rule_set.follow_label}
+          </div>}
+       end * ', '
+     else
+       "No following preference"
+     end
+   end
+
```
ethn: basic idea is to make it so that there's a simple view where you can see what you're following.  Made it so that you go there after editing the rule.

ethn: probably want something a bit different for "cancel", but at least it's not recursing now.

card/mod/05_email/set/self/edited.rb
```diff
@@ -11,7 +11,7 @@ def title
 end
 
 def label
-  "follow what I've edited"
+  "follow if I edited"
 end
```
ethn: changed my mind about the wording after looking at it a bit.  may change again after feedback :)

wagn/lib/decko/engine.rb
```diff
@@ -23,7 +23,6 @@ class Engine < ::Rails::Engine
     
 
     initializer :connect_on_load do
-      Cardio.cache == ::Rails.cache
```
ethn: perhaps the problem was just this == vs =

wagn/lib/decko/engine.rb
```diff
@@ -35,13 +34,6 @@ class Engine < ::Rails::Engine
         
         ActiveRecord::Base.establish_connection(::Rails.env)
       end
-      ActiveSupport.on_load(:after_initialize) do
-        begin
-          require_dependency 'card' unless defined?(Card)
-        rescue ActiveRecord::StatementInvalid => e
-          ::Rails.logger.warn "database not available[#{::Rails.env}] #{e}"
-        end
-      end
```
ethn: I don't really like this code.  Gerry, I'm guessing there was some reason for it, but:
a) this is an odd case to catch a missing database, and
b) card will soon be required if a controller action is triggered.

I would suggest that we move this kind of check into cardio.

wagn/lib/wagn/config/environments/development.rb
```diff
@@ -1,3 +1,9 @@
+
+
+Decko::Engine.configure do
+  config.cache_classes = false
+end
```
ethn: this makes it so that card.rb reloads with each request.

card/mod/05_email/set/self/edited.rb
GerryG: That seems better.

wagn/lib/wagn/config/environments/development.rb
GerryG: Ok, but why is it this one rather than the application?

ethn: It was already in the application (further down the same file).  I left it there, but if it's there alone, it doesn't work.

card/mod/03_machines/lib/stylesheets/functional.scss
```diff
@@ -55,7 +55,8 @@ in your skin.  Choosing (b) will mean your CSS will not be affected by automated
 
 .card-header {
   .card-header-title {
-    overflow:hidden;
+    display: inline-block;
+    overflow: hidden;
```
ethn: this was an inline style attribute before (on a div with a new class).

card/mod/05_email/set/all/follow.rb
```diff
@@ -49,12 +49,12 @@
...
-        html_options[:hover_content] = "unfollow #{args[:label]}" 
-        html_options[:text]          = "following #{args[:label]}"
-      else
+#      if args[:hover]
+#        html_options[:hover_content] = "unfollow #{args[:label]}" 
+#        html_options[:text]          = "following #{args[:label]}"
+#      else
         html_options[:text]          = "unfollow #{args[:label]}"
-      end
+#      end
```
ethn: the hovering is making the submenu jump around.  If that's the case, I'd rather just take out the hover behavior until we update the menu handling.

card/mod/05_standard/set/all/error.rb
```diff
@@ -92,6 +92,7 @@ def unsupported_view view
     return '' unless card.ok? :create  # should this be moved into ok_view?
 
     opts = { :remote=>true, :class=>"slotter missing-#{ args[:denied_view] || args[:home_view]}" }
+    opts[:path_opts] = { :type=> args[:type] } if args[:type]
```
ethn: I think this got lost somehow.  The idea was to make it so that the "add whatever" links to missing nested cards set the type correctly.

card/mod/05_standard/set/rstar/rules.rb
```diff
@@ -282,8 +283,9 @@ def find_current_rule_card
     # This generates a prototypical member of the POTENTIAL rule's set
     # and returns that member's ACTUAL rule for the POTENTIAL rule's setting
     if card.new_card?
-       ((setting = card.right) && card.set_prototype.rule_card(setting.codename, :user=>card.rule_user)) ||
-            Card.new(:name=> "#{Card[:all].name}+#{card.rule_user_setting_name}")
+      if setting = card.right
+        card.set_prototype.rule_card setting.codename, :user=>card.rule_user
+      end
```
ethn: current closed rule handling expects nil if there is no current rule.

wagn/lib/wagn/config/environments/development.rb
GerryG: Ah.  So this is the one that needs to be in both.

card/mod/06_bootstrap/set/all/bootstrap/wrapper.rb
```diff
@@ -2,8 +2,10 @@
 
   def frame args={}, &block
     args.reverse_merge!(
-      :slot_class=>'card-frame panel panel-default',
-      :body_class=>'panel-body'
+      :slot_class   => 'panel panel-default',
+      :header_class => 'panel-heading',
+      :title_class  => 'panel-title',
+      :body_class   => 'panel-body'
     )
```
ethn: this is very satisfying ;)

xithan:
tasks like 'wagn update' don't work anymore: "uninitialized constant Card::Cache"

wagn/lib/decko/engine.rb
xithan: removing the `require_dependency 'card'` killed the rake tasks

PR #357, opened 2015-03-04
Followlinks

ethn:
this is a new approach to getting the follow links to sync up.  Basically, a lot of the logic moves to javascript.

After playing with the functionality a bit, I realized this was actually kind of vital for the demo; it's just an awkward user experience when it doesn't work.  

Not well tested yet.


ethn:
whoops, didn't push master's latest first.

PR #358, opened 2015-03-04
Followlinks

ethn:
this is a new approach to getting the follow links to sync up. Basically, a lot of the logic moves to javascript.

After playing with the functionality a bit, I realized this was actually kind of vital for the demo; it's just an awkward user experience when it doesn't work.

Not well tested yet.


card/mod/03_machines/lib/javascript/wagn_mod.js.coffee
```diff
@@ -304,6 +303,27 @@ $(window).ready ->
...
+      success : (data)->
+        tags = anchor.closest('.card-menu').find('.follow-toggle')
+        tags.find('.follow-verb').html data.verb
+        tags.attr 'title', data.title
+        tags.attr 'class', data.class
+        tags.data 'follow', data
+    }
+    event.preventDefault() # Prevent link from following its href
+    
+
```
ethn: this does all the work of making the request, interpreting the json results, and updating the links

card/mod/05_email/set/all/follow.rb
```diff
@@ -18,57 +18,50 @@
...
+      hash[:content] = '*never'
+      hash[:title]   = "stop sending emails about changes to #{card.follow_label}"
+      hash[:verb]    = 'unfollow'
+    when :on
+      hash[:content] = '*always'
+      hash[:title]   = "send emails about changes to #{card.follow_label}"
+      hash[:verb]    = 'follow'
     end
+    hash
+
```
ethn: this same hash is used in the follow_status json and in the follow_link view

card/mod/05_standard/set/all/links.rb
```diff
@@ -5,6 +5,8 @@
   # link is called by web_link, card_link, and view_link
   # (and is overridden in other formats)
   def link_to text, href, opts={}
+    href = interpret_href href
```
ethn: this isn't deeply related; I just noticed this inconsistency while planning my approach.

PR #360, opened 2015-03-04
fix follow link for self sets

xithan:


card/mod/05_standard/set/rstar/rules.rb
```diff
@@ -246,7 +246,7 @@ def default_follow_item_args args
            else
              card.rule_set.follow_label
            end
-    link_target = if card.rule_set.codename == 'self'
+    link_target = if card.rule_set.tag.codename == 'self'
```
ethn: this is good; will merge.

Realized that this would be faster: `card.rule_set.right_id == SelfID`

xithan: Wouldn't work: `Card.fetch('Home+*self').right_id` is nil.  I guess because the card is virtual.

ethn: oh. well not working is a bit of a downside :(

GerryG: Interesting.  It will always be a real card if there are any rules on the set, but maybe this is a case when the rule may be about to be created, or potentially created.  Certainly card.rule_set.tag is a real card (it has a codename), but if A+*self is new, it might not have a right_id set on the object.

PR #364, opened 2015-03-07
Add missing autoload paths, rename add_gem_paths

GerryG:


card/lib/cardio.rb
```diff
@@ -23,10 +22,11 @@ def cache
...
     
     def set_config config
-      @@config = config
-      @@root = @@config.root
-        
-      config.autoload_paths += Dir["#{Cardio.gem_root}/mod/*/lib/**/"]
+      @@config, @@root = config, config.root
+
+      config.autoload_paths += Dir["#{gem_root}/mod/*/lib/**/"]
+      config.autoload_paths += Dir["#{gem_root}/lib/**/"]
```
GerryG: I think this fixes the reload! issue.

ethn: cool. I'm tempted to limit this to lib/* and lib/card/**/ so that we don't look through generators for autoload stuff.  But cool that this solves issues.

Does this make it so that we don't really need the require_dependency in engine.rb?

GerryG: I suppose one of us should investigate that.

One thing that I'm not clear on.  This doesn't mean that it loads all the matching files, but that it can load them with the auto-resolve constants stuff, right?

PR #365, opened 2015-03-07
Card generator namespace, test scripts, README to markdown with badge

GerryG:
I think these changes are less controversial and more complete.


README.md
```diff
@@ -1,82 +1,82 @@
-=Wagn: how pioneers roll
```
ethn: what's the argument for md over rdoc?

README.md
```diff
@@ -1,82 +1,82 @@
...
+#Wagn: how pioneers roll
 
-==Basics
+[![Build Status](https://semaphoreapp.com/api/v1/projects/0d61c1f9-5ef0-4c5c-89c6-7664e247e4be/294670/badge.png)](https://semaphoreapp.com/ethan/wagn)
```
GerryG: Not a huge one, but I was trying out the Semaphore badge, and they don't have one for RDOC, but they do for markdown.  If there is a way to do this in RDOC, then maybe we could try that.

ethn: Oh, I don't think we want a Semaphore badge on the README.

That badge is for core developers, but README users are generally new people trying to install.  Seems like it has a downside of advertising broken builds and no real upside.

ethn: I suppose we could put in on the contributing page...

ethn:
have you tested all the generators?

card/lib/generators/card.rb
```diff
@@ -0,0 +1,43 @@
...
+          else
+            @_card_source_root ||= File.expand_path(File.join(File.dirname(__FILE__), 'card', generator_name, 'templates'))
+          end
+        end
+
+        # Override Rails default banner.
+        def banner
+          "wagn generate #{namespace} #{self.arguments.map(&:usage)*' '} [options]".gsub(/\s+/, ' ')
+        end
+      end
```
ethn: if this is the same code twice, shouldn't we stick it in a module and extend twice?

README.md
GerryG: I'm fine if you don't want this or want it somewhere else.  Putting it in README does make it show on the github page for any branch with the badge.

People trying to install a version of wagn/decko will want this information.  They won't want to install code when this isn't green.  It should always be green on master, and if it isn't a core dev better fix it soon.

card/lib/generators/card.rb
GerryG: I didn't notice that, but sure, that would work.  I guess the larger question is whether we need/want both of these classes, and if this file is really the best place for it.  Suggestions are welcome, and if no more are forthcoming I'll just do as you suggest above.

GerryG:
I have tested the generators, and I will do it again just to make sure we are good.

I will try to test all of the command line scenarios that we often miss.  I'm not sure if you saw me report that the tests do not all pass on a migrated database (i.e. run the migration from 1.10 test, then run specs and cucs).  I'll also try to verify by suspicion that this has been broken in master for a while now.

README.md
ethn: No man, that's my point.  Anyone who is not a core developer is going to install from the gem; that's what the instructions on the README page tell them to do.  If the master branch is red, it doesn't mean the gem is broken.  (We should never push out a gem branch based on broken code.)

**So putting the badge on this page will often tell users "don't install this now" even when it's perfectly safe to install.**

We need a strong separation between communication to core developers and deck managers.  This is putting a core developer message at the top of the deck manager communication space.

ethn: > It should always be green on master, and if it isn't a core dev better fix it soon.

Mostly true.  But master is the new develop, and the gem is the new master.  If master gets broken, we need to fix it in a hurry.   But whereas master needs to be green, say 98% of the time, the current gem version needs to be working 99.99% of the time.  (And we don't want to risk losing the users who want to install during the 2% red time.)

card/lib/generators/card.rb
ethn: I don't really know what these classes are buying us.  The banner mostly?

ethn:
I'm not sure I would expect the tests to pass on a migrated database.  There are lots of things we've changed about the test seeding in the meantime.  It's mostly just:

1. do the migrations all run?, and
2. does the site roughly work?

If we were going to automate that, we'd have to figure out some way to codify #2, but I doubt it would ever pass all our tests, which require a very specific data set.

In any case, the traditional stumbling block has been #1: usually after we make enough changes our old migrations start breaking.

README.md
GerryG: I think this is wrong.  README should direct them to INSTALL or README_INSTALL, if at all.  The REAME and in particular the badges isn't part of the gem, but the repo.  I don't think they'll even see it if they install the gem.  You can put another README (INSTALL might be better) in the wagn gem for that purpose.  We may need a more general overhaul of these doc files.

Like I said, I'm fine if you don't want this at all, but I don't think you have good arguments for why it is a bad idea.  It may also point out we have deficiencies here with nothing to do with the badge or markdown.

GerryG:
I agree with the basic scenario.  I think #2 should more than roughly work.  We need to explain why/what doesn't.  The theory is that anything in the test seeds ought to migrate correctly and work afterwards.  Anything short of that means there can be manual fixing after the upgrade.  We do need that sometimes, but we want to specifics to be completely clear.  Even if this does work, site admins may have to deal with un-migratable stuff they added (or local wagneers added) that doesn't have equivalents in the test seeds and test cases.  If it doesn't work then people with generic wagns will have troubles.

card/lib/generators/card.rb
GerryG: I'll double check and drop this if we don't need it now.  We might only use one of them now.  The one banner case was only for the generator we don't have anymore (rails g card), and that I didn't include on this branch in any case.

GerryG:
It looks like this got merged somehow.  This actual PR isn't merged, but the commits seem to be in master already.  I'll keep adding the changes we need to this branch and complete the testing.

ethn:
> You can put another README (INSTALL might be better) in the wagn gem for that purpose. We may need a more general overhaul of these doc files.

Yeah, that's not a thing.  There's no documentation space on rubygems.  We link from there to wagn.org and to github; we need to have good docs whichever route they choose.

The README is intended for normal users.  Read it.  It describes gem installation.  It's the same copy (or should be) as on wagn.org.  Web searches might take them to one or the other, but the truth is that MANY people find this README, and the only ones that are wanting information specific to core developers are, well, people who are already core developers. 

CONTRIBUTING.rdoc is for developers. If we're going markdown, we can convert that, too, and a badge makes sense there: that's a great place to advertise that we're using semaphore.

> We may need a more general overhaul of these doc files.

Perhaps, but we definitely want the landing page of the git repo to speak to people new to wagn.  That's a no-brainer.  If you're feeling stubborn about this, look at https://github.com/rails/rails.

GerryG:
I'll revert the README change.

I'm testing the generators and there is some inconsistency that I'm not sure the origin of.  The first argument of the format and set generators should be the name of the directory in the mod/ dir, right?  Also, we do want to support the -c option (--core) on each of these, right?

ethn:
I think that is all correct.  To be honest, I haven't been using this very much, and I couldn't swear that this option worked before.  But that would rock!

card/lib/generators/card.rb
```diff
@@ -5,38 +5,27 @@
...
-            @_card_source_root ||= File.expand_path(File.join(File.dirname(__FILE__), 'card', generator_name, 'templates'))
-          end
-        end
-
-        # Override Rails default banner.
-        def banner
-          "wagn generate #{namespace} #{self.arguments.map(&:usage)*' '} [options]".gsub(/\s+/, ' ')
-        end
-      end
+      extend ClassMethods
```
ethn: thanks for doing this.

PR #366, opened 2015-03-07
Fastfollow

ethn:
the following functionality was proving to have a strong negative impact on performance in production, especially when there were lots of users with default "created" or "edited" rules.  It slowed down all card actions (as in, not simple read requests, but all the others) even in cases where there were notifications to send. 

The main change here is to move most of the organization around the following tests into load time (as opposed to test time).  (see follow_option.rb)

### DIFF
--- a/card/lib/card/name.rb
+++ b/card/lib/card/name.rb
@@ -37,5 +37,9 @@ class Card
       name = trait_name( tag_code )
       name ? name.s : ( raise Card::NotFound, "unknown codename: #{tag_code}" )
     end
+    
+    def code
+      Codename[ Card.fetch_id self ]
+    end
   end
 end
--- a/card/mod/01_core/set/all/rules.rb
+++ b/card/mod/01_core/set/all/rules.rb
@@ -23,19 +23,20 @@ def is_rule?
 end
 
 def is_standard_rule?
-  (l = left( :skip_modules=>true ))  &&
-  l.type_id == Card::SetID           &&
   (r = right( :skip_modules=>true )) &&
-  r.type_id == Card::SettingID           
+  r.type_id == Card::SettingID       &&
+  (l = left( :skip_modules=>true ))  &&
+  l.type_id == Card::SetID
 end
 
 def is_user_rule?
+  cardname.parts.length > 2                                  &&
+  (r = right( :skip_modules=>true ))                         &&
+   r.type_id == Card::SettingID                              &&
   (set = self[0..-3, :skip_modules=>true])                   &&
    set.type_id == Card::SetID                                && 
   (user = self[-2, :skip_modules=>true] )                    &&
-  (user.type_id == Card::UserID  || user.codename == 'all' ) &&
-  (r = right( :skip_modules=>true ))                         &&
-   r.type_id == Card::SettingID
+  (user.type_id == Card::UserID  || user.codename == 'all' )
 end
 
 
--- a/card/mod/05_email/lib/card/follow_option.rb
+++ b/card/mod/05_email/lib/card/follow_option.rb
@@ -3,9 +3,10 @@
 
 class Card
   module FollowOption
-    mattr_reader :test
+    mattr_reader :test, :test_option
+    @@test, @@test_option = {}, {}
+    
     @@options = { :all=>[], :main=>[], :restrictive=>[] }
-    @@test = {}
     
     def self.included(host_class)     
        host_class.extend ClassMethods
@@ -52,6 +53,10 @@ class Card
       def follow_test opts={}, &block
         codename = get_codename opts
         Card::FollowOption.test[codename] = block
+      end
+      
+      def follow_test_option option, &block
+        Card::FollowOption.test_option[option] = block
       end
       
       private
--- a/card/mod/05_email/lib/card/follow_option.rb
+++ b/card/mod/05_email/lib/card/follow_option.rb
@@ -3,8 +3,9 @@
 
 class Card
   module FollowOption
-    #mattr_reader :codenames
+    mattr_reader :test
     @@options = { :all=>[], :main=>[], :restrictive=>[] }
+    @@test = {}
     
     def self.included(host_class)     
        host_class.extend ClassMethods
@@ -48,6 +49,11 @@ class Card
         add_option args, :main
       end
       
+      def follow_test opts={}, &block
+        codename = get_codename opts
+        Card::FollowOption.test[codename] = block
+      end
+      
       private
       
       def insert_option pos, item, type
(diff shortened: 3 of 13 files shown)


card/lib/card/name.rb
```diff
@@ -37,5 +37,9 @@ def trait tag_code
       name = trait_name( tag_code )
       name ? name.s : ( raise Card::NotFound, "unknown codename: #{tag_code}" )
     end
+    
+    def code
+      Codename[ Card.fetch_id self ]
+    end
```
ethn: this could be optimized further.

card/mod/01_core/set/all/rules.rb
```diff
@@ -23,19 +23,20 @@ def is_rule?
...
+  cardname.parts.length > 2                                  &&
+  (r = right( :skip_modules=>true ))                         &&
+   r.type_id == Card::SettingID                              &&
   (set = self[0..-3, :skip_modules=>true])                   &&
    set.type_id == Card::SetID                                && 
   (user = self[-2, :skip_modules=>true] )                    &&
-  (user.type_id == Card::UserID  || user.codename == 'all' ) &&
-  (r = right( :skip_modules=>true ))                         &&
-   r.type_id == Card::SettingID
+  (user.type_id == Card::UserID  || user.codename == 'all' )
```
ethn: no real logic change here.  trying to speed it up a bit.  no real evidence yet that this is working...

card/mod/05_email/lib/card/follow_option.rb
```diff
@@ -48,6 +50,15 @@ def follow_opts args
...
       
+      def follow_test opts={}, &block
+        codename = get_codename opts
+        Card::FollowOption.test[codename] = block
+      end
+      
+      def follow_test_option option, &block
+        Card::FollowOption.test_option[option] = block
+      end
+
```
ethn: could stand for some naming improvements.  the follow test gets a block that is the actual test.  The test_option could be called _argument instead -- it's about sending arguments to the test.

This was done this way to avoid processing the same test options over and over for the same card for each user, as was happening before (on wikirate, which had tons of users)

ethn: well, not tons, but they all had follow rules.

card/mod/05_email/set/all/follow.rb
```diff
@@ -118,16 +120,30 @@ def followed?
...
+      if test = FollowOption.test[ value.to_name.code ]
+        if test.call @follow_test_options.merge( :user_id => user_id )
+          return value.gsub( /[\[\]]/, '' )
+        end
+      else
+        Rails.logger.debug "didn't find follow test for #{ value }"
       end
     end
   end 
   return false
```
ethn: this is the pivotal test code.  Much faster now.  no cards instantiated, no duplicate queries (eg in edited).  But more hard code than is optimal!

card/mod/05_email/set/self/edited.rb
```diff
@@ -2,10 +2,15 @@
...
 self.restrictive_follow_opts :position=>2
 
-def applies_to? card, user_id
-  Card.search(:editor_of=>card.name, :return=>:id).find { |editor_id| editor_id.to_i == user_id }
+self.follow_test do |opts|
+  opts[:editor_ids].find { |editor_id| editor_id == opts[:user_id] }
 end
 
+self.follow_test_option :editor_ids do |card|
+  Card.search( :editor_of=>card.name, :return=>:id ).map &:to_i
```
ethn: really, this doesn't even have to involve the cards table.

card/mod/01_core/set/all/rules.rb
ethn: I do think we need to take the "User" part out of this.  the requirement is that there be an account, not that there be a user card.  We could probably get away without confirming this cardtype anyway ;)

card/lib/card/name.rb
GerryG: I think you do have to do a fetch, to change from name to id, but you could just get the codename from the fetched card (no need for modules or virtuals).  This can often be a cache hit in any case.  But you don't save much either, just a hash table lookup of the id for the code.

ethn: yeah, this is pretty much exactly how I was thinking of it.  Probably a good idea, not a huge deal.

card/mod/05_email/lib/card/follow_option.rb
xithan: the wording here obscures the functionality. `follow_test_option` is not about some optional options it's the core of finding the followers (took me a while to get that). I don't think we need in `follow_test_options` anything else but user ids. Maybe something in this direction:
````ruby
def select_follower_ids &block
  Card::FollowOption.select_follower_ids[get_codename({})] = block
end
````
and then you use it as follows in edited.rb for example:
````ruby
self.select_follower_ids do |card|
  Card.search( :editor_of=>card.name, :return=>:id ).map &:to_i
end
self.follow_test do |user_id, approved_user_ids|  
  approved_user_ids.include? user_id
  # this could be the default and we override it for *always and *never
end

File: card/mod/05_email/set/all/follow.rb [hunk 6]
```diff
@@ -167,13 +179,15 @@ def direct_followers
...
 # doesn't include users that follow this card because they are following parent cards or other cards that include this card
-def direct_follower_ids args={}  
+def direct_follower_ids args={}
   result = ::Set.new
-  set_names.each do |set_name| 
-    set_card = Card.fetch(set_name)
-    set_card.all_user_ids_with_rule_for(:follow).each do |user_id|
-      if (!result.include? user_id) and self.follow_rule_applies?(user_id)
-        result << user_id
+  set_follow_test_options do
```
xithan: this was very confusing to me. "set" appears six times in various names in this method but in all other cases it means card set.

card/mod/05_email/lib/card/follow_option.rb
ethn: yeah, the naming was poor.  Sorry.

Actually, if we go your route, we might be able to make it so that *always and *never are the only ones that typically override the follow_test.

Because if *create uses "select_follower_ids" (or maybe "find_follower_ids")?, then it will presumably need to return an array of one id rather than just the id itself.  And then the "follow_test" will be the same as *edited (and both of the voted ones, too).  I like that.

card/mod/05_email/set/all/follow.rb
ethn: so we have "set" the verb, "Set" the core ruby class, and "set" as in card set.  Yeah, that's bad, too.  Will change.

PR #367, opened 2015-03-08
History mod

GerryG:
Consolidates history related code in a single module (most of it, there are some history views and such that probably should be here too)


card/mod/01_history/set/all/content.rb
```diff
@@ -0,0 +1,132 @@
...
+def content_with_selected
+  if @selected_action_id
+    @selected_content ||= begin
+      (change = last_change_on( :db_content, :not_after=> @selected_action_id ) and change.value) || db_content
+    end
+  else
+    content_without_selected
+  end
+end
+alias_method_chain :content, :selected
```
ethn: I'm not a fan of the alias_method_chain stuff. could use super...

wagn/rails/controllers/card_controller.rb
```diff
@@ -119,9 +119,7 @@ def load_card
       end
     raise Card::NotFound unless @card
     
-    if action = @card.find_action_by_params( params )
-      @card.selected_action_id = action.id
-    end
+    card.select_action_by_params params
```
ethn: yay for moving code out of the controller!

card/mod/01_history/set/all/content.rb
GerryG: I thought about that.  I guess I'll give it a try.

GerryG:
I think the code is cleaner this way.  That's the best argument for this way of doing things.

I'm not sure I totally get the argument about tweaks being harder.  With the aliasing, I'd think it gives you a clearer way to tweak things on either side of this alias, but with super you can't really do that without serious voodoo.  Load order could still effect which level your tweaks would effect in much the same way, no?

GerryG:
I'm going to try an confirm this, but I think this is not a use case for super.  My suspicion is that because both definitions are in all sets and loaded into the main Card class, they are just replacing the methods from one module to the next.  I have an idea I will try next.

card/mod/01_history/set/all/history.rb
```diff
@@ -7,17 +7,10 @@ def create_act_and_action
   else
     acts.build(:ip_address=>Env.ip)
   end
-  
+
   @current_action = actions.build(:action_type=>@action, :draft=>(Env.params['draft'] == 'true') )
   @current_action.act = @current_act
-  
-  # @current_act = (@supercard ? @supercard.current_act : Card::Act.new(:ip_address=>Env.ip))
```
GerryG: This didn't look like something we were likely to revert to using ...

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -201,6 +201,7 @@ namespace :wagn do
       Card::Cache.reset_global
       conn =  ActiveRecord::Base.connection
       # Correct time and user stamps
+      # USER related
```
GerryG: Next I want to create a clear separation of user stamps and authentication so we can plug this into an authentication system in the environment.

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -215,6 +216,7 @@ namespace :wagn do
           end
         end
         Card::Cache.reset_global
+        # FIXME: can this be associated with the machine module somehow?
```
GerryG: If it is only stuff like this one, and it is not common this may be ok, but what is the right solution to move this sort of logic to a module?  At least maybe move the code to the module and do something here like: if defined?(Card::Set::MachineSomething) { Card::Set::MachineSomething.cleanup }  

Better would be a more standard way of adding stuff to the config and iterating through railties to find the modules with a particular hook.

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -225,6 +227,8 @@ namespace :wagn do
 
       Card.empty_trash
 
+      # FIXME: move this to history
```
GerryG: This should use a similar way of registering things to do with particular tasks, rather than having modules needing to customize rake tasks.

GerryG:
It mostly worked.  For some reason it isn't working right for one of the call.  Maybe the other cases are not well tested.  It seems to have all the right modules included in Card, but it doesn't ever enter the one in Content::ContentHistory

GerryG:
D'oh.  Didn't realize that when the set path and target is the same for a set module, it really is the same module and that's what caused the overloading.

card/mod/01_history/set/all/content_history.rb
```diff
@@ -0,0 +1,131 @@
+
+# if these aren't in a nested module, the methods just overwrite the base methods,
+# but we need a distict module so that super will be able to refer to the base methods.
```
ethn: I don't think I understand this.  There is no base method #content; just methods defined in modules.  In ruby, super doesn't only apply to super classes, it traverses the entire ancestry list, which includes modules.

card/mod/01_history/set/all/actions.rb
```diff
@@ -0,0 +1,47 @@
+# -*- encoding : utf-8 -*-
```
ethn: this is not an issue for this PR, but I really want to get rid of this file naming pattern (01_whatever) and start representing dependencies in the module.  It really shouldn't be that hard, because the *all modules are all loaded independently first and then later included.  It's just a matter of resorting them first.

(I started the naming pattern with formats; it's my fault.  I just meant it as a temporary hack, not a core pattern.)

wagn/lib/wagn/tasks/wagn.rake
ethn: It might be helpful to split out tasks that are just for core developers.  We could even put them in the repo root (as in, not in the gems), because this is at best noise for non-core-developers, and at worst it could be really confusing.

`bootstrap:clean` and `bootstrap:dump`

wagn/lib/wagn/tasks/wagn.rake
ethn: this task isn't the right place for that; this just cleans up the template database from which the bootstrap data is generated.

card/mod/01_history/set/all/actions.rb
GerryG: Ok, so the ordering is purely for getting the load order right when they are included into the Card class?

My request is that we address the selection of modules at the same time we handle ordering.  I wan't to be able to skip some of the modules in the gem (card) with configuration overrides, and I think it would be fine to just override a default configuration the also gets us the load ordering.

ethn: Right, it's about load order.

> My request is that we address the selection of modules at the same time we handle ordering. 

that's a little different.

You're talking about picking and choosing mods (analagous to a Gemfile), we're talking about dependency tracking, which needs to be represented in the mods themselves (analagous to a gemspec).

Why should we conflate those two?

GerryG: They are not the same, but connected.  The full dependency feature would have to handle lots of stuff eventually, including mods from different containers.  You are talking the analogy to a gemspec for mods, which supports Gemfile functionality.  Seems to me that we would do selection and ordering in the same features, and if you have one, you'll want the other pretty soon.

The real issue here is the set of modules included in the core gem are just all loaded, so there is no dependency handling involved, just include all of these is this order.  Once we re-organize these standard and core mods into a handful of dependency chains, we'll want to be able to 1) not include some chains, or 2) replace a dependency with a mod handling the same role.

The history module added in this change is meant to be replaceable in this way.  I might have a gem containing one or more mods that implements externalized history.  One replaces the standard history module with one that sends messages to listeners, and several listeners that can process the messages to store them, send them to networked clients or perform notifications and other post processing.

ethn: good thoughts.

I didn't mean to suggest they weren't related.  If we're talking *design*, then it makes sense to design them together.  But if we're talking implementation, it's best to bite off smaller chunks.

GerryG: That's sort of what I was thinking with my request.  Dependency handling is a lot more complex than just configuring a load order for the core mods, and that would be an opportunity to leave specific items out of the list.  That might be enough for the standard modules.  Maybe you could just have a single array that had all of them, core or deck looking first in the deck, then the gem.

The full solution probably needs for the use cases to develop a bit so that we really need a more complex solution for dependencies.

ethn: I guess I had a picture that was much more oriented towards whitelisting
than blacklisting.

Let's say:

A depends on B,
F depends on G, and
Y depends on Z.

There are several ways we could create this whitelist.  It could be:

A,B,F,G,Y,and Z
A,F,Z
:mypackage (referring to some group of them)
...or whatever.

I don't know where exactly this configuration happens, but it's obviously
somewhere in the deck space.

Like with gems, "A depends on B" means that when you A, you must have B.
Unlike gems, when card mod A depends on card mod B, B needs to be loaded
first.

We could certainly get fancier than this, but it seems like this would be
very powerful already.

The reason I'm thinking of the whitelist and the dependency handling as
separate is that most of what I want to happen is this:

   1. mod whitelists are configured in deck
   2. at run time, mods are loaded (into ruby constant space) in standard
   order (based on mod paths config / alphabet)
   3. adjustments to mod inclusion order is worked out based on dependencies
   4. *all set modules are loaded immediately based on mod inclusion order
   5. other set modules are loaded dynamically based on mod inclusion order


In this model, they would seem to be different phases of the process.


On Mon, Mar 9, 2015 at 5:39 PM, Gerry Gleason <notifications@github.com>
wrote:

> In card/mod/01_history/set/all/actions.rb
> <https://github.com/wagn/wagn/pull/367#discussion_r26087613>:
>
> > @@ -0,0 +1,47 @@
> > +# -*- encoding : utf-8 -*-
>
> That's sort of what I was thinking with my request. Dependency handling is
> a lot more complex than just configuring a load order for the core mods,
> and that would be an opportunity to leave specific items out of the list.
> That might be enough for the standard modules. Maybe you could just have a
> single array that had all of them, core or deck looking first in the deck,
> then the gem.
>
> The full solution probably needs for the use cases to develop a bit so
> that we really need a more complex solution for dependencies.
>
> —
> <https://github.com/wagn/wagn/pull/367/files#r26087613>.
>



-- 
Ethan McCutchen
One of the Wagneers, Wagn.org

Wagn. How pioneers roll.

s: ethan.mccutchen
t: @intogreater

PR #368, opened 2015-03-08
Remove a lot of requires after autoload fix

GerryG:
A bit of oddity about Set constant.  A few things like exceptions don't autoload because the missing constants aren't the filename.


card/lib/card/loader.rb
```diff
@@ -1,5 +1,7 @@
 # -*- encoding : utf-8 -*-
 
+require 'card/set'
```
ethn: should be require_dependency

wagn/features/notifications.feature
```diff
@@ -19,7 +19,8 @@ Feature: Notifications
     Then Joe Admin should be notified that "Joe User created \"Foo\""
     And the card Phrase+*type+*followers should point to "Joe Admin"
     And I should see "was just created by Joe User" in the email body
-    And I should see |You received this email because you're following "all "Phrases""| in the email body  #FIXME these double quotes are ugly
+    #FIXME these double quotes are ugly
+    And I should see |You received this email because you're following "all "Phrases""| in the email body
```
ethn: still there, no?

ethn:
If we're tweaking requires this deeply, I'd want to make sure we've hand-tested a few different load scenarios: wagn new, wagn seed, wagn update, rake tasks, console, etc.  I don't remember why all those requires were there, but many of them predated the autoload removal.

card/lib/card/loader.rb
GerryG: Oops, will fix.

wagn/features/notifications.feature
GerryG: It wasn't matching the step definition.  I don't think cucumber strips comments before matching, so the comment has to be on it's own line.  That appears to be the quoting generated, so you have to fix the test, then fix the code.  For now, these steps actually pass, and I don't really know what you want the fixed behavior to be.

ethn: oh, you just moved the comment.  doh!

GerryG:
Sure, I'll need to run that on this branch too.  It would be nice if someone else was verifying my reports as well.

wagn/features/notifications.feature
GerryG: I think I also fixed a questionable step definition, but yes.

card/lib/card.rb
```diff
@@ -4,20 +4,10 @@
...
-  require_dependency 'card/format'
   require_dependency 'card/exceptions'
-  require_dependency 'card/auth'
   require_dependency 'card/log'
-  require_dependency 'card/loader'
-  require_dependency 'card/content'
   require_dependency 'card/action'
-  require_dependency 'card/act'
-  require_dependency 'card/change'
-  require_dependency 'card/reference'
```
ethn: It occurs to me that one nice thing about having these explicit is that it hedges against cases where users define a root level constant of the same name.

For example, let's say I create a rails app with a module in the root namespace named Change (or Act or Reference or Format or whatever).

Then I add decko-rails to my app.

When there is a contextual reference to Card::Change like this:

````
class Card
  module MyMod
    def all_changes
      Change.all
    end
  end
end
````
...that only works with autoloading so long as there is no known constant "Change".

So, after further thought, I think we should probably keep these explicit and *only* autoload card.rb itself.

File: wagn/features/notifications.feature (hunk 2 above)
Comment (ethn) [inline wagn/features/notifications.feature]: is it definitely matching now?

File: wagn/features/step_definitions/email_steps.rb [hunk 4]
```diff
@@ -95,11 +95,15 @@ def current_email_address
...
-Then /^(?:I|they) should see [\|"](.*)[\|"] in the email body$/ do |text|
+Then /^(?:I|they) should see \|(.*)\| in the email body$/ do |text|
   expect(current_email.text_part.body.raw_source).to include("#{text}")
 end
 
-Then /^(?:I|they) should see \/([^"]*?)\/ in the email body$/ do |text|
+Then /^(?:I|they) should see "(.*)" in the email body$/ do |text|
+  expect(current_email.text_part.body.raw_source).to include("#{text}")
+end
+
```
ethn: Ah, I think I get what you did here now.  Do you think we need all these variants?  If you have a stylistic preference (and we don't use them all in the tests), feel free to pick one.

card/lib/card.rb
ethn: Besides feeling confirmed that I'm annoying and wishy-washy, any reaction to this?

GerryG: I'm a bit wishy-washy on this too.  The fact is that is one of the uglinesses of rails autoloading, and that is why there needed to be a require for Card::Set in the loader.  Otherwise it ends up loading the global Set into Card.

I guess I lean a little towards leaving them out until a problem develops, but I could easily be swayed the other way.  Probably, we shouldn't include things like the history classes (card/act|action|change) because Card shouldn't depend on them directly because history may load a whole other set of classes.

wagn/features/step_definitions/email_steps.rb
GerryG: Don't know if we need or want them, just making them work.  At first I thought I'd use the '/' form, but that delimits a regexp.  The other case allowed double quotes or pipes, but I didn't trust the combined regexp so I made two.

card/lib/card.rb
ethn: I think I'm voting for the explicit route.  Autoloading is for rapid development, but we need a higher premium on reliability.

I would only want NOT to require the history stuff if we really want card to stand without the history stuff. That would be a bit of a leap.  I haven't really pondered that.

card/lib/card/loader.rb
```diff
@@ -94,8 +96,8 @@ def load_formats
       def load_sets
         prepare_tmp_dir 'tmp/set'
         load_sets_by_pattern
-        Set.process_base_modules
-        Set.clean_empty_modules
+        Card::Set.process_base_modules
+        Card::Set.clean_empty_modules
```
ethn: Let's not add these in.  the require_dependency should obviate them.

card/lib/card.rb
GerryG: I think it is essential in the long run that history is optional.  If only to move it out of the critical code paths, but a deeper reason is that I might want to have system components that don't need or want history because they are acting like caches and/or workflow components.  Yes, we want history to be conceptually integrated with the model, but it is awfully expensive when you don't need it.

I think it is fine to have all of them here, it may gain a bit of reliability as you say.  The right answer would make these part of the module dependency system.  Then they always get required according the module rules and dependencies.

card/lib/card/loader.rb
GerryG: Yeah, that was done before I realized I needed the require.  Without it, these throw an error about the fact that Card::Set has an explicit prefix, but it isn't defined there.

GerryG:
Ran various test scenarios to exercise command line stuff.  The only problem I note is that *wagn seed -d* seems to clear the test db and only load the dev db.  I don't think *wagn seed -t* also clears development.

The noted issue with running tests after migrations.  Assuming that is because we actually added new support data to the seeds that probably counts as expected.  Otherwise there are migration issues not related to the split or this branch.

I'll look into the seed thing, but don't think it should hold this up.

ethn:
I think that seed thing (wiping test) has something to do with default rails behavior for db:create (or something similar).  I don't like it for wagn.

Re the 1.10 migration thing, it doesn't look to me like it was ever test data in the first place (I think it was just the dump of a newly created site).  The sql dump doesn't appear to contain the word "joe".

Neither of these should hold this PR up at all.  I'll merge it now.

Thanks for rolling back some of those requires.  In the long run we may not want them there, but I think whatever solution we come up with will need to ensure you can't break wagn by creating constants for common words in the root namespace.

GerryG:
You're probably right about db:create.  On the 1.10, I actually loaded that version, seeded the test db and dumped that fresh.  I'm not using the one in the repo.  I can add that and more of my support files to wagn/test/script or somewhere near there if you want.  I'd have to look at whether the seed.rb for the fixtures got any new stuff, I suspect it is something like that.

PR #370, opened 2015-03-09
follow_rule_applies? reorganization

ethn:
a little further cleanup based on @xithan's suggestions.

### DIFF
--- a/card/mod/05_email/set/all/follow.rb
+++ b/card/mod/05_email/set/all/follow.rb
@@ -99,7 +99,7 @@ end
 # for sets and cardtypes it doesn't check whether the users is following the card itself
 # instead it checks whether he is following the complete set
 def followed_by? user_id
-  set_follow_test_options do
+  with_follower_candidate_ids do
     if follow_rule_applies? user_id
       return true
     end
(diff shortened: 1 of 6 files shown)


card/mod/05_email/set/all/follow.rb
```diff
@@ -115,29 +115,38 @@ def followed?
...
-      if test = FollowOption.test[ value.to_name.code ]
-        if test.call @follow_test_options.merge( :user_id => user_id )
-          return value.gsub( /[\[\]]/, '' )
+           
+      value_code = value.to_name.code
+      accounted_ids = @follower_candidate_ids[ value_code ] ||= begin
+        if block = FollowOption.follower_candidate_ids[ value_code ]
+          block.call self
+        else
+          []
```
ethn: this lazy loading means we don't pre-process, eg, *created and *edited candidates when no rules with those values apply

card/mod/05_email/set/self/never.rb
```diff
@@ -2,7 +2,7 @@
 
 self.follow_opts :position=>3
 
-self.follow_test { |opts| false }
+self.follow_test { |follower_id, accounted_ids| false }
```
ethn: not sure whether these args are necessary to track here, but I would want to check old rubies before removing them, because some of that handling has changed.

card/mod/05_email/set/all/follow.rb
```diff
@@ -115,29 +115,38 @@ def followed?
...
 end
 
-def set_follow_test_options
-  @follow_test_options = {}
-  FollowOption.test_option.each do |key, val|
-    @follow_test_options[key] = val.call(self)
-  end
+
+def with_follower_candidate_ids
+  @follower_candidate_ids = {}
```
ethn: much happier with this nomenclature!

xithan: me too :+1:

PR #371, opened 2015-03-09
Coverage

ethn:
minor tweaks to get more accurate coverage numbers.

Even with these fixes, it looks like we've had significant slippage and are now down around 84% (from either 90% or very close to that)


ethn:
whoops.

PR #372, opened 2015-03-11
fixed type changes on rule cards

ethn:


card/mod/05_standard/set/rstar/rules.rb
```diff
@@ -166,31 +166,32 @@ def set_fieldset args
...
-                            #{ card_link set_name, :text=> Card.fetch(set_name).label, :target=>'wagn_set' }
-                            #{'<em>(current)</em>' if is_current }
-                            #{"<em> #{card_link "#{set_name}+#{card.rule_user_setting_name}", :text=>"(overwritten)"}</em>" if state == :overwritten }
-                          </span>
-                         }.html_safe
-                     end
-
-                   end
+    option_list = 
+      wrap_each_with :li do
```
ethn: is this still within the style guidelines?  made for much more readable code given how deep it got.

card/mod/05_standard/set/rstar/rules.rb
```diff
@@ -147,7 +147,7 @@ def editor args
 
   def type_fieldset args
     fieldset 'type', type_field(
-      :href         => path(:card=>args[:success][:card], :view=>args[:success][:view], :type_reload=>true),
+      :href         => path(:name=>args[:success][:card].name, :view=>args[:success][:view], :type_reload=>true),
       :class        => 'type-field rule-type-field live-type-field',
```
ethn: further evidence for the need for better path docs/api.

card/mod/05_standard/set/rstar/rules.rb
GerryG: I would still join these two lines, but keep the indentation below the same.  You get one indent for the incomplete statement (=) and one for the open do block.  So, I would say it is well within them.

ethn: Philipp's code complies with the style guide, and it looks like mine does, too.  What you're suggesting does not, though:

https://github.com/bbatsov/ruby-style-guide#indent-conditional-assignment

ethn: (and, for what it's worth, I think the style guide makes sense there)

ethn: nice to have that guide as a referee ;)

GerryG: It is after all a guide, but it doesn't actually cover what I'm suggesting as good or bad.  The bad form would be not to indent for the hanging expression (the rest of the body of the do ... end).

Then there is the style disagreement on keywords that don't begin or end a block, but just divide it.  That is when in a case, protected, private, etc. in modules, rescue, ensure in begin ... end.  Arguably else and elsif in if ... end as well.  Same rule should probably be for all of this, and I think that is our style even if we are not explicit about it.  I sort of like indenting when and else (and the blocks) in case, but when I seem people doing that with the storage classes in modules, I never like it.

ethn: I thought you were suggesting
````
option_list = wrap_each_with :li do
  args[:set_options].map do |set_name, state|
    ...
  end
end
````

the guide says that is "bad - pretty convoluted"

ethn: I'm not sure about your other point or how it relates.  Any specific code in mind?  The guide is pretty helpful.

The point isn't to suggest that the guide is "right" about any of this.  The point is that if we just agree to follow the guide, then we don't have to debate these style issues point by point.

ethn: oh, are you saying the guide doesn't apply here because it's not a conditional but a block? I suppose that's true, but it seems like the same logic applies, no?

GerryG: Close, I'm proposing more like this:

````
option_list = wrap_each_with :li do
    args[:set_options].map do |set_name, state|
      ...
    end
  end
````

ethn: that's not ok by the style guide, and I agree.  it means you have to do a difficult visual search to match the "wrap_with_each" to the end.  much easier to read in its current form.

GerryG: For me, it still pops out fine this way.  The queues I need are there.  And I'm particularly challenged in the vision department.  

Agreed about the guide in general.  The only reason we are debating the point is you wondered if what you did was within or not.  I'm saying that you should just apply the guide when you write the code and beyond that use your judgement.  I don't think we should critique each other's code based on the style guide, but on the actual merits of the case in point.  Most of the time, it just plain doesn't matter enough and we should just accept the code as committed.  If you're actually having trouble following, or offended by some ugliness, then carry on.

PR #373, opened 2015-03-11
context name clearing

ethn:
needed because of several cases where context names were too inclusive.

Consider: Cardtype Project includes lists of tasks (A, B, C) each of which has a +Project card.

currently, the context_name is getting passed down so that A+Project has both Project and A in its context names.  When all of a card's name parts are in the context, we give the full name as title, because we don't know what to strip.  But clearly we are really in the "A" context here (not the "Project" context).

Solution is to make it so that only parts of a card's name can be in its context.

### DIFF
--- a/card/mod/03_machines/lib/stylesheets/standard.scss
+++ b/card/mod/03_machines/lib/stylesheets/standard.scss
@@ -646,7 +646,7 @@ td.missing-rule {
 .rule-setting {
   padding-top: 4px;
 }
-.open-rule .rule-setting {
+.open-rule > .rule-setting {
   width: 28%;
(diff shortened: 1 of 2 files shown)


card/mod/03_machines/lib/stylesheets/standard.scss
```diff
@@ -646,7 +646,7 @@ td.missing-rule {
 .rule-setting {
   padding-top: 4px;
 }
-.open-rule .rule-setting {
+.open-rule > .rule-setting {
```
ethn: not relevant, but fixes styling issue in nested rule editors.

PR #374, opened 2015-03-11
fix not showing ace editor in setting page

chuenlok:
1. card type is stated in textarea instead of slot now as rows in setting page don't have slot
2. include layout type to show html editor

### DIFF
--- a/card/mod/03_machines/lib/javascript/wagn_mod.js.coffee
+++ b/card/mod/03_machines/lib/javascript/wagn_mod.js.coffee
@@ -37,7 +37,7 @@ $.extend wagn,
     wagn.tinyMCEConfig = setter()
   
   initAce: (textarea) ->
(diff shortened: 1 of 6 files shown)


card/mod/03_machines/lib/javascript/wagn_mod.js.coffee
```diff
@@ -37,7 +37,7 @@ $.extend wagn,
     wagn.tinyMCEConfig = setter()
   
   initAce: (textarea) ->
-    type_code = textarea.slot().attr "data-card-type-code"
+    type_code = textarea.attr "data_card_type_code"
```
ethn: if I understand correctly, it's actually kind of important that it be "data-" rather than "data_".  See http://www.w3schools.com/tags/att_global_data.asp

Either data-type_code or data-type-code is fine.

ethn:
I will merge this in with the requested fix.

PR #375, opened 2015-03-11
Compress generator branch to one commit

GerryG:
New and Improved.

### DIFF
--- a/card/lib/card/cache.rb
+++ b/card/lib/card/cache.rb
@@ -18,10 +18,9 @@ class Card
     TEST_ENVS = %w{test cucumber}
     @@prepopulating     = TEST_ENVS.include? Rails.env
     @@using_rails_cache = TEST_ENVS.include? Rails.env
-    @@prefix_root       = Cardio.config.database_configuration[Rails.env]['database']
     @@cache_by_class    = {}
 
-    cattr_reader :cache_by_class, :prefix_root
+    cattr_reader :cache_by_class
 
     class << self
       def [] klass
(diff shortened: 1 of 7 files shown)


card/lib/card/cache.rb
```diff
@@ -40,6 +39,11 @@ def renew
         reset_local
       end
 
+      def prefix_root
+        @@prefix_root ||= ( cfg = Cardio.config and cfg = cfg.database_configuration and
+                            cfg[Rails.env]['database'] )
+      end
+
```
ethn: fyi: looks like there are still about 60 results for "Rails" in card.  We got rid of the require, but autonomy is a ways off yet.

wagn/lib/decko/engine.rb
```diff
@@ -13,26 +13,36 @@
...
-    
+    paths.add 'lib/tasks',       :with => "#{::Wagn.gem_root}/lib/wagn/tasks", :glob => '**/*.rake'
+    paths.add 'lib/wagn/config/initializers',
+              :with => File.join( Wagn.gem_root, 'lib/wagn/config/initializers' ), :glob => "**/*.rb"
+
+    initializer 'decko.engine.load_config_initializers',  :after => :load_config_initializers do
+      paths['lib/wagn/config/initializers'].existent.sort.each do |initializer|
+        load(initializer)
+      end
+    end
```
ethn: so now this runs before we even define Wagn::Application?  and that doesn't cause any problems?

GerryG: Is that true?  The tricky part is that we can't depend on Wagn::Application getting defined, because it doesn't.  Some other MyApp < Rails::Application does, and sometimes it does, but these have to load either way.  I would think all that would be defined by this time.  Well, now I'm not sure about :loadOconfig_initializers, need to look for that now.

card/lib/card/cache.rb
GerryG: I guess I was mostly looking for Wagn at first.  Need to survey those a bit more.

PR #377, opened 2015-03-18
Titled default

ethn:
The idea was to make it so that titled was the default view throughout.  I haven't actually flipped that switch yet, but:

New setting = _\*default html view_
Old sites have rule: _\*all+\*default html view_ set to content


card/db/migrate_core_cards/20150202143810_import_bootstrap_layout.rb
```diff
@@ -2,16 +2,54 @@
...
+      end
+#    else
+#      unmerged.map! do |row|
+#        puts "didn't merge #{row['name']}"
+#        row['name'] += '+alternate'
+#        row
+#      end
+#      Card.merge_list unmerged
+#    end
+
```
ethn: at first I wanted to make it so that *any* non-pristine style/skin would not get updated, but I kept just ending up with bizarrely bad pages.  This part is still a WIP (though no worse than before).

card/mod/01_core/spec/format/html_format_spec.rb
```diff
@@ -86,22 +86,22 @@
...
+
+#      it "should default to open view for main card" do
+#        @layout_card.content='Open up {{_main}}'
+#        Card::Auth.as_bot { @layout_card.save }
+#
+#        result = @main_card.format.render_layout
+#        expect(result).to match(/Open up/)
+#        expect(result).to match(/card-header/)
+#        expect(result).to match(/Joe User/)
+#      end
```
ethn: these were breaking, as well they should!

wagn/lib/wagn/tasks/db.rake
```diff
@@ -47,7 +47,7 @@ namespace :db do
     desc 'Prepare the test database and load the schema'
     Rake::Task.redefine_task( :prepare => :environment ) do
       if ENV['RELOAD_TEST_DATA'] == 'true' || ENV['RUN_CODE_RUN']
-        puts `env RAILS_ENV=test rake wagn:create`
+        puts `env RAILS_ENV=test rake wagn:seed`
```
ethn: Following through on a rename Gerry started.

card/mod/01_core/format/html_format.rb
```diff
@@ -45,15 +40,21 @@ def layout_from_card_or_code name
...
     def default_item_view
       :closed
     end
 
+    def nest nested_card, opts={}
+      unless opts[:view].present?
+        opts[:view] = nested_card.rule :default_html_view
+      end
+      super nested_card, opts
+    end
```
xithan: This means default_html_view overrides the view defined explicitly in the inclusion (because this is handled in super if there is no view in opts). I don't think that's a good idea.

PR #379, opened 2015-03-19
further bootstrapification

ethn:
a bunch of further changes to improve bootstrap adaptability.

Not ready for merge yet (need to do some more dumps / migration work)


card/db/migrate_core_cards/data/bootstrap_layout.json
```diff
@@ -54,7 +54,7 @@
...
...gin: 0 20px;\r\n}\r\n\r\n\r\n/*---------- comments -------------------------------------- */\r\n\r\n.comment-box {\r\n  padding: 8px;\r\n  margin: 0px;\r\n  input {\r\n    padding: 0px;\r\n  }\r\n  textarea {\r\n    margin-bottom:5px;\r\n  }\r\n}\r\n.w-comment-author {\r\n  font-style: italic;\r\n  margin-left: 10px;\r\n}\r\n\r\n\r\n/* the logo and credit will not be in the sidebar in most new layouts */\r\n\r\naside {\r\n  > div {\r\n    > a > .SELF-Xlogo.content-view,\r\n    > .SELF-Xcredit.content-view {\r\n      display: block;\r\n      text-align: center;\r\n      padding: 10px;\r\n      background: #fff;\r\n      border: 1px solid #bbb;\r\n      -moz-border-radius: 8px;\r\n      border-radius: 8px;\r\n      img {\r\n        max-width: 100%;\r\n      }\r\n      a {\r\n        color: #186886; /* NOTGREY - Wagn blue */\r\n      }\r\n    }\r\n  }  \r\n}\r\n\r\n\r\n\r\n.closed-content {\r\n  font-size: 1em !important;\r\n}\r\n\r\n\r\n/* english content in css.  ugh */\r\n.search-no-results:after {\r\n  content:'0 results';\r\n}\r\n.closed-content .search-no-results:after {\r\n  content:'';\r\n}\r\n\r\nlegend {\r\n  h2 {\r\n    font-size: 0.7em;\r\n  }\r\n  border: 0;\r\n}\r\n\r\n"
```
ethn: certainly a downside of this json approach that it makes this stuff unreadable...

card/mod/02_basic_types/set/type/pointer.rb
```diff
@@ -41,14 +41,16 @@ def wrap_item item, args={}
     _render_core args
   end
 
-  view :edit do |args|
-    super(args.merge(:pointer_item_class=>'form-control'))
-  end
+#  view :edit do |args|
+#    super(args.merge(:pointer_item_class=>'form-control'))
+#  end
```
ethn: we actually only want the form-control class on controls like input, password, textarea, etc.  We *don't* want it on checkboxes and radios (it was causing a lot of headaches).  this is much cleaner.

card/db/migrate_core_cards/data/bootstrap_layout.json
ethn: a lot of the not-yet-submitted changes are to this card (which is good, because it only affects bootstrap layouts)

card/mod/03_machines/lib/stylesheets/functional.scss
```diff
@@ -136,20 +136,6 @@ in your skin.  Choosing (b) will mean your CSS will not be affected by automated
...
-  > .button-area,
-  > .editor {
-    float: right;
-  }
-  > legend {
-    white-space: inherit;
-    float:left;
-  }
-}
-
```
ethn: need to make sure forms still look reasonable in old styles.

card/mod/03_machines/lib/stylesheets/standard.scss
```diff
@@ -277,24 +244,6 @@ textarea {
...
-        width: auto;
-        margin: 0;
-    }
-    label {
-        float: left;
-        width: auto;
-        margin: 6px 8px;
-     }
-}
-
```
ethn: possible that this needs to be restored in "traditional" sheet.

card/mod/05_standard/set/type/search_type.rb
```diff
@@ -231,13 +250,6 @@ def default_search_params
     set_default_search_params :default_limit=>20
   end
   
-  
 end
```
ethn: lots of these changes are from my moving stuff in from bootstrap mod.

In general, if it calls "super" I leave it, because it's actually separating things out.  But if it's a hard override, I usually just copy it back in; no need to keep two copies of the same code.

GerryG: Presumably there is also a standard of generality, so that if there is stuff that should be in the other mod, we split the base functionality from the customization, with super, if we ca.

ethn: Right.  In many cases, all we were doing to support bootstrap was adding CSS classes.  Sometimes it was hard to do that without overriding an entire view, so we temporarily did that, but it was pretty ugly to have both, with the overridden one never used.  I do hope we improve the api to make it so that adding in CSS classes with super is a very easy thing to do; it would be a very common use case for many integrations.

There were, however, some cases where we changed the HTML, and our strategy there has been for that HTML to become the new default.  Examples of that are the new card-header, the new page header, and the new handling of form fields.  We kept the changes as minimal as possible, but I do suspect there will be some old skins that will need to be updated if folks choose to update to the new bootstrap layout.

PR #380, opened 2015-03-19
search results and signin success event

xithan:
1) moved search results out of search_vars to its own method
2) added a signin_success event so that it is possible to have an event after a successful signin. The only drawback in the api is that it only works as `:before=>:signin_success` event  and not with `:after=>:signin_success`


card/mod/05_standard/set/type/search_type.rb
```diff
@@ -63,25 +63,34 @@ def get_query params={}
...
-    rescue =>e
-      { :error => e }
-    end
+    @search_vars ||=
+      begin
+        v = {}
+        v[:query] = card.query(search_params)
+        v[:item]  = set_inclusion_opts args.merge( :query_view=>v[:query][:view] )
+        v
+      rescue =>e
```
xithan: Should I keep the @search_vars[:results] value for compatibility reasons? How likely is it that there are decks that use this method. I override the new search_results method in a wikirate mod. That's why I isolated it.

card/mod/05_standard/set/self/signin.rb
```diff
@@ -96,6 +95,10 @@ def default_core_args args={}
   end  
 end
 
+event :signin_success, :after=>:signin do
+  abort :success
+end
+
```
xithan: The signin event always aborted the action so it wasn't possible to do something after a signin. I moved the `abort :success` to a separate event.

ethn: clever!

card/mod/05_standard/set/type/search_type.rb
ethn: unlikely.

I *think* that the reason we did things this way was that the old cloning-based format approached every instance variable had to be clearer by hand.  Since that no longer applies, I'd guess this is safe.

card/mod/05_standard/set/self/signin.rb
GerryG: Wondering if you would use :after => :signin of :before => :signin_success to add something after signin?  As long as it is loaded after this code, I think either works, but is one better?

ethn: Oog, I can't remember how the ordering works.  I've really got to get on the documentation work (planning to send something out about that soon. I have a rough plan in mind, but this is not a great place to discuss that)

I guess the latter has the advantage that you don't have to remember the ordering logic ;)

GerryG: Yeah, the latter is a little more odd looking, but the independence of ordering is a plus.

PR #381, opened 2015-03-20
fix titled default

xithan:


card/lib/card/format.rb
```diff
@@ -114,15 +114,15 @@ def include_set_format_modules
...
     end
   
-    def inclusion_defaults
+    def inclusion_defaults nested_card
       @inclusion_defaults ||= begin
-        defaults = get_inclusion_defaults.clone
+        defaults = get_inclusion_defaults(nested_card).clone
         defaults.merge! @inclusion_opts if @inclusion_opts
         defaults
       end
```
ethn: this is a little bit of a problem, conceptually at least (perhaps not in practice yet)

the @inclusion_defaults here is intended to be a little cache of this value *for the format object*.  but the inclusion defaults can differ by nested card, theoretically.

so if a given object has several nests, and the nested cards have different rules, this will break.

since that's not the case today on wikirate, we can deploy, but we should probably fix before 1.15

xithan:
merged to get wikirate tests running

card/lib/card/format.rb
ethn: *I* should probably fix.  (and add tests!)

ethn: I think I see how to fix.  in HtmlFormat:
````
def inclusion_defaults nested_card, opts={}  #send opts from nest
  defaults = super
  if !opts[:view] && !defaults[:view]
    defaults.merge :view=>(nested_card.rule(:default_html_view) || :titled )
  else
    defaults
  end
end
````
then you don't actually have to give an arg to get_inclusion_defaults

PR #382, opened 2015-03-20
fix titled default

xithan:


card/mod/05_standard/set/type/search_type.rb
```diff
@@ -217,7 +216,7 @@ def default_search_params
     total = card.count search_params
     return '' if limit >= total # should only happen if limit exactly equals the total
 
-    @paging_path_args = { :limit => limit, :item=> inclusion_defaults[:view] }
+    @paging_path_args = { :limit => limit, :item=> inclusion_defaults(card)[:view] }
```
xithan: I'm not sure how we want to handle inclusion defaults in this case. Above in line 174 I use the default of every item card but here in the page link we define a default for all items.

ethn: good point.  in pointers and searches we probably don't want anything card-specific

ethn: I don't think you want a list of closed, closed, titled, closed, closed... :)

GerryG: Interesting.  I would think that in general when you are doing options on nested contexts (item_view, and now the more general syntax of more pipes (|) to provide options on each nested context), that you never want the value of the options to vary item to item in a list (search/pointer).  I guess it can when we access the default value and it is overridden in card sets.

ethn: Right, I was trying to say that the latest code introduces this, and I think it's a bug.  As a rule, regardless of what kind of setting we're talking about, nest should override rule should override code.  But, as written, I think we have a case where the rule overrides the nest.

We had to get this out quickly for Wikirate purposes, so we merged it in before really perfecting it.  I know that's not good form; will try to avoid that in the future.

PR #387, opened 2015-03-26
Hardcode

ethn:
cleaned up handling of hard-coded files a bit.


GerryG:
The changes look a bit odd on the PR.  Some of the diffs seem to show file movement that should have been done with the gem re-org.  Probably just git history oddities, but I can't tell.

PR #388, opened 2015-03-26
fix history and themes

xithan:


card/mod/03_machines/set/type/scss.rb
```diff
@@ -4,12 +4,11 @@ def diff_args
   {:format=>:text}
 end  
 
-
 format do
   include Css::Format
   
   view :core do |args|
-    process_content compile_scss(_render_raw)
+    compile_scss(process_content _render_raw)
```
xithan: make inclusions work in css and scss cards

card/mod/01_history/set/all/history.rb
```diff
@@ -40,14 +40,13 @@ def create_act_and_action
...
     Action.fetch(a_id) || nil
   end
-  rollback_actions.each do |action|  
+  rollback_actions.each do |action|
     if action.card_id == id
-      revision.merge!(revision(action)) 
+      revision.merge!(revision(action))
     else
-      revision[:subcards].merge!(revision(action))
+      revision[:subcards][action.card.name] = revision(action)
```
xithan: this was the reason that for every db field a card was created

xithan:
maybe useful to review this PR: add ?w=1 to the url to hide whitespace changes

card/mod/03_machines/set/type/scss.rb
ethn: I feel like there was a rationale for the previous order, but I can't remember what now.  Perhaps it was something like wanting to be able to click "add +image" in the middle of an html rendering of a css file?

At any rate, I think for now this is better.

ethn: And if not, it's my fault for not writing a test way back when ;)

card/mod/06_bootstrap/set/self/bootswatch_mixins.rb
```diff
@@ -0,0 +1,9 @@
+view :raw do |args|
+  File.read "#{Cardio.gem_root}/mod/06_bootstrap/lib/stylesheets/#{card.codename}.css"
+end
+
+format :html do
+  view :editor do |args|
+    "Content is stored in file and can't be edited."
+  end
+end
```
ethn: I gather this isn't used yet here (but is in the other PR)?

card/mod/06_bootstrap/set/all/rich_bootstrap.rb
```diff
@@ -1,7 +1,7 @@
 format :html do
   
-  def glyphicon icon_type
-    content_tag(:span, '', :class=>"glyphicon glyphicon-#{icon_type}", 'aria-hidden'=>true)
+  def glyphicon icon_type, extra_class=''
+    content_tag(:span, '', :class=>"glyphicon glyphicon-#{icon_type} #{extra_class}", 'aria-hidden'=>true)
   end
```
ethn: we really need a more general solution for classes!

PR #389, opened 2015-03-26
wagneerable bootswatch themes

xithan:
this PR removes 106,187 lines. I guess that's a new record :)


card/db/migrate_core_cards/20150326205655_bootswatch_themes.rb
```diff
@@ -0,0 +1,26 @@
...
+        theme.update_attributes! :type_id=>Card::ScssID, :codename=>nil, :content => "{{+variables}}{{bootswatch shared}}{{+style}}", :subcards=> {
+          "+variables" => {:type_code=>:scss, :content=>File.read(File.join path, '_variables.scss')},
+          "+style"     => {:type_code=>:scss, :content=>File.read(File.join path, '_bootswatch.scss')}
+        }
+      else
+        Card.create! :name=>"theme: #{theme}", :type_code=>:scss, :content => "{{+variables}}{{bootswatch shared}}{{+style}}", :subcards=> {
+          "+variables" => {:type_code=>:scss, :content=>File.read(File.join path, '_variables.scss')},
+          "+style"     => {:type_code=>:scss, :content=>File.read(File.join path, '_bootswatch.scss')}
+        }
+      end
```
xithan: I don't need a code rule anymore for these theme so the "bootswatch theme" cardtype is not necessary. But the question is whether we want to have a structure rule for these themes instead of setting the content to be `{{+variables}}{{bootswatch shared}}{{+style}}`. A `bootswatch+*right+*structure` rule and using names like "theme: darkly+bootswatch" is an alternative

ethn:
I do feel safer waiting on this one, but from what I can see so far I'm enthusiastic about 1.15.1!

card/db/migrate_core_cards/20150326205655_bootswatch_themes.rb
GerryG: So, the actual theme card would be virtual?  You could just have a regular cardtype (i.e. no codename or code for it) that inherits from Theme (or whatever type theme cards are now).  In theory inheritable created cardtypes is working.

ethn: Not sure the inheritability matters here, Gerry.  The question is not about rules (inheritable) but about nest patterns (not inheritable).

I'm tempted to start a pattern like this:

Cosmo Skin (Skin) -> 
- themeless bootstrap skin
- +bootswatch theme

This is very close to what you're suggesting, Philipp, but the plus card is based off the skin, which means there's actually only one simple card with the actual skin/theme name (eg "Cosmo", "Darkly", etc).  Thoughts?

xithan: sounds good, I'll change the migration

GerryG: I mean inherit in the ruby class for the model sense.  What we recently fixed so that it uses the *default rule's Cardtype for user created cardtypes (without *structure rule, with *default rule).  It used to always be Basic unless you have a module.

PR #391, opened 2015-03-27
added primary button css class

ethn:
...here and there.

### DIFF
--- a/card/mod/06_bootstrap/set/all/bootstrap/form.rb
+++ b/card/mod/06_bootstrap/set/all/bootstrap/form.rb
@@ -1,13 +1,12 @@
 format :html do
   def button_tag content_or_options = nil, options = {}, &block
-    if block_given?
-      content_or_options[:class] ||= ''
-      content_or_options[:class] += ' btn btn-default'
-    else
-      options[:class] ||= ''
-      options[:class] += ' btn btn-default'
-    end
+    bootstrapify_button( block_given? ? content_or_options : options )
     super(content_or_options, options, &block)
+  end
+  
+  def bootstrapify_button options
+    situation = options.delete(:situation) || 'default'
+    options[:class] = [ options[:class], 'btn', "btn-#{situation}" ].compact*' '
   end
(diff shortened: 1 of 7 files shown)


card/mod/06_bootstrap/set/all/bootstrap/form.rb
```diff
@@ -1,15 +1,14 @@
...
-      options[:class] += ' btn btn-default'
-    end
+    bootstrapify_button( block_given? ? content_or_options : options )
     super(content_or_options, options, &block)
   end
   
+  def bootstrapify_button options
+    situation = options.delete(:situation) || 'default'
+    options[:class] = [ options[:class], 'btn', "btn-#{situation}" ].compact*' '
+  end
```
ethn: not sure "situation" is perfect, but "context" already has lots of meaning in wagn, and "state" seems wrong too (those are the two bootstrap words for this).

default, primary, info, success, warning, danger.... hard to find a word that feels really natural for all of those.

GerryG: Maybe something like :button_context ?  What does class="btn-primary" do?

ethn: look into bootstrap components.  what I'm calling "situation" they call "context" or "state", and it's a much broader concept than just buttons.

GerryG: Ok, that makes more sense.  More of a modifier on the container classes, so more general.  We don't overload 'state' as much as 'context', maybe that would be better.  Is it bootstrap related?  I mean, it's their class semantics, so maybe :bootstrap_state, but just :state might be clear enough.

ethn: I don't think state makes much sense here.  state = primary?  state = warning?  Situation isn't beautiful, but it seems like state has a lot of misleading connotations that situation doesn't have, no?

PR #392, opened 2015-03-27
fixed encoding so Philipp can have a proper umlaut in his name

ethn:


GerryG:
Ok, why does encoding matter for the gemspec files?  What context does this fix?

ethn:
umlaut in PK's name.  see PR description.

PR #396, opened 2015-04-07
new gear menu

xithan:
almost ready - I have to adjust some cucumber tests


card/db/bootstrap/card_actions.yml
```diff
@@ -2933,1269 +2933,1514 @@ card_actions_419:
   action_type: 0
   draft: false
```
ethn: starting to think we should put the bootstrap data in a subrepo and maybe a separate gem.  really becoming a nuisance for code review :(

card/mod/01_core/chunk/include.rb
```diff
@@ -5,7 +5,7 @@
 module Card::Chunk
   class Include < Reference
     cattr_reader :options
-    @@options = ::Set.new [ :inc_name, :inc_syntax, :view, :items, :type, :size, :title, :hide, :show, :structure ]
+    @@options = ::Set.new [ :inc_name, :inc_syntax, :view, :items, :type, :size, :title, :hide, :show, :structure, :unlabeled ]
```
ethn: could we do this as hide: set_label?  This seems like a fairly narrow case for new nest syntax.  As you've probably noticed, I try to keep the bar for new nest syntax high, because we're soon going to introduce a gui nest editor, and I want it to be very streamlined.

card/mod/03_machines/lib/javascript/wagn.js.coffee
```diff
@@ -23,19 +23,23 @@ $.extend wagn,
...
 jQuery.fn.extend {
-  slot: -> 
-    if @attr('slotSelector')
-      @closest @attr('slotSelector')
+  slot: ->
+    if @data('slot-selector')
+      close = @closest(@data('slot-selector'))
+      if close.length == 0
+        $('body').find(@data('slot-selector'))
+      else
```
ethn: this makes me nervous; it could find tons of slots.  Why do we need this?

card/mod/03_machines/lib/javascript/wagn_mod.js.coffee
```diff
@@ -126,113 +126,6 @@ $.extend wagn,
...
-        path += vars['linkname']
-      else
-        params['card[name]'] = vars['self']
-      
-      params['view'] = i.view unless i.view == 'home'      
-      path += '?' + $.param(params) unless $.isEmptyObject params
-      
-      text = i.text || i.view
-      '<a href="' + path + '" data-remote="true" class="slotter">' + text + '</a>'
-
```
ethn: auf wiedersehen

card/mod/03_machines/lib/javascript/wagn_mod.js.coffee
```diff
@@ -253,31 +146,6 @@ $(window).ready ->
...
-  $(document).on 'tap', 'body', (event) ->
-    unless $(event.target).closest('.card-header')[0] or $(event.target).closest('.card-menu-link')[0]
-      $('.card-menu').hide()
-      # this and mouseleave should use a close menu method that handles collapsing. (though not seeing bad behavior...)
-
-  $(document).on 'tap', '.ui-menu-icon', (event)->
-    $(this).closest('li').trigger('mouseenter')
-    event.preventDefault()
-
-
```
ethn: Lebe wohl

card/mod/03_machines/lib/javascript/script_card_menu.js
```diff
@@ -1,72 +0,0 @@
...
-  { "text":"account", "related":{"name":"+*account", "view":"edit"}, "if":"account", "sub":
-    [
-      { "text":"account", "related":{"name":"+*account", "view":"edit"}, "text":"details" },
-      { "related":"roles"   },
-      { "related":"created" },
-      { "related":"edited"  },
-      { "related":"follows" }
-    ]
-  }
-
```
ethn: Abschied für immer

card/mod/03_machines/lib/stylesheets/style_cards.scss
```diff
@@ -49,28 +49,92 @@ Choosing (b) will mean your CSS will not be affected by automated updates.
 /* -- card-headers and menus--*/
 
 .card-header {
+  position: relative;
   .card-header-title {
     display: inline-block;
     overflow: hidden;
+    span.glyphicon {
```
ethn: since these are the non-bootstrap rules, perhaps this should be a non-bootstrap selector?

card/mod/05_email/set/right/follow.rb
```diff
@@ -1,9 +1,87 @@
...
+      tag :span, :class=>"glyphicon glyphicon-ok", 'aria-hidden'=>"true"
+    end
+
+  end
+
+  view :add_follow_rule_button do |args|
+    button_tag :type=>:submit, :class=>'btn-xs btn-item-add', 'aria-label'=>'Left Align' do
+      tag :span, :class=>"glyphicon glyphicon-plus", 'aria-hidden'=>"true"
+    end
+  end
```
ethn: can you say in a sentence or two what all this new code is for?

ethn:
Why does this show 64.9K additions?  Is there some merge confusion re bootswatch themes?

card/mod/01_core/chunk/include.rb
xithan: A case of too much respect for existing code :)  The label used to be handled with args so I needed that but makes sense to turn that into an optional view

card/mod/03_machines/lib/javascript/wagn.js.coffee
xithan: When you click on a toolbar button the target slot is a sibling and not a parent. If you choose your slot-selector carefully it won't find tons.

card/mod/05_email/set/right/follow.rb
xithan: the modal stuff is for the dialog that appears when you choose follow in the card menu. The rest moved here from all/rules.rb. We had the pattern `set+setting+user` for user-specific rules and because we don't support `*type plus *right plus *type` set patterns I had to use an *all rule. Now with the pattern `set+user+setting`  I can use right/follow for this.

card/mod/03_machines/lib/stylesheets/style_cards.scss
xithan: hm, but we don't have non-bootstrap icons

xithan:
this includes also the bootswatch themes PR

card/mod/03_machines/lib/stylesheets/style_cards.scss
ethn: OK to ignore this for now.  Fwiw, I was just thinking that if we added a class like "card-header-icon" or something, then we could have the margins work, so if some day someone adds a module to use font-awesome rather than glyphicon, this kind of margin styling would work.

Not worth sweating if you want to move on :)

ethn:
but the bootswatch PR mostly deletes files and this mostly adds them. that's why I got confused.

xithan:
I don't know why the deletes don't appear here but the commit for that is part of this PR. The additional adds are the db seed changes

card/mod/03_machines/lib/javascript/wagn.js.coffee
GerryG: For the case where close.length == 0 ?

$('body') is the whole doc, can't you at least start at the root of the menu?  Or are these card-slots?

ethn: Is it safe to assume that the the slot-selector is always a sibling?  If so, you could go back to the parent slot rather than to body, as Gerry suggests.  It does seem like it would be nice if we had some other relative alternatives.

PR #400, opened 2015-04-16
modal and navbar fix

xithan:


card/mod/05_standard/set/all/rich_html/modal.rb
```diff
@@ -43,7 +45,7 @@ def default_modal_content_args args
   end
 
   view :modal_title do |args|
-    _render_title args.merge(:title_class=>'modal-title')
+    "<h4>#{_render_title args.merge(:title_class=>'modal-title')}</h4>"
```
wagn: do we want to consider giving it a "header"?

xithan: What do you mean by "header"? A <header> tag? There is a "modal_header" view with a "modal-header" css class that calls "modal_title"

ethn: oh, then why do we need a modal_title view?  can't it just render the normal title with the args?

I was actually asking because conceptually it seemed like the card-header context would give you whatever you needed instead of using the h4, but maybe the h4 brings in more bootstrappiness?

PR #406, opened 2015-04-20
move menu in closed view to the right side

xithan:
fixes the problem that the menu is not visible in a list of closed views

### DIFF
--- a/card/mod/03_machines/lib/javascript/wagn_mod.js.coffee
+++ b/card/mod/03_machines/lib/javascript/wagn_mod.js.coffee
@@ -186,7 +186,7 @@ $(window).ready ->
       type : 'GET'
       dataType : 'json'
       success : (data) ->
-        tags = modal.closest('.card-menu').find('.follow-link')
+        tags = $(modal).parent().find('.follow-link')
         tags.find('.follow-verb').html data.verb
         tags.attr 'href', data.path
         tags.attr 'title', data.title
@@ -205,7 +205,7 @@ $(window).ready ->
         'success[id]'   : anchor.data('card_key')
       }
       success : (data) ->
-        tags = anchor.closest('.card-menu').find('.follow-toggle')
+        tags = anchor.closest('.modal').parent().find('.follow-toggle')
         tags.find('.follow-verb').html data.verb
         tags.attr 'title', data.title
         tags.removeClass( 'follow-toggle-on follow-toggle-off').addClass data.class
--- a/card/mod/05_standard/set/all/rich_html/menu.rb
+++ b/card/mod/05_standard/set/all/rich_html/menu.rb
@@ -2,7 +2,8 @@ format :html do
   view :menu, :denial=>:blank, :tags=>:unknown_ok do |args|
     return _render_template_closer if args[:menu_hack] == :template_closer
     return '' if card.unknown?
-    _optional_render(:horizontal_menu, args, :hide) || _render_menu_link(args)
+    (_optional_render(:horizontal_menu, args, :hide) || _render_menu_link(args)) +
+      _render_modal_slot(args)
   end
 
   view :menu_link do |args|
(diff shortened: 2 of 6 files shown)


card/mod/05_standard/set/all/rich_html/content.rb
```diff
@@ -111,8 +111,10 @@ def show view, args
   end
 
   view :closed do |args|
-    frame args.reverse_merge(:content=>true, :body_class=>'closed-content', :toggle_mode=>:close, :optional_toggle=>:show ) do
-      _optional_render :closed_content, args
+    wrap_with :div, :class=>'closed-wrapper' do
+      frame args.reverse_merge(:content=>true, :body_class=>'closed-content', :toggle_mode=>:close, :optional_toggle=>:show ) do
```
xithan: the frame has `overflow: hidden` so we have to position the menu relative to something outside of the frame

card/mod/03_machines/lib/javascript/wagn_mod.js.coffee
```diff
@@ -241,6 +241,14 @@ $(window).ready ->
...
 
 
+  $('body').on 'click', '.submit-modal', ->
+    $(this).closest('.modal-content').find('form').submit()
+
+  $('body').on 'mouseover', '.closed-view', ->
+    $(this).find('.card-menu').show()
+  $('body').on 'mouseleave', '.closed-view', ->
+    $(this).find('.card-menu').hide()
+
```
ethn: I think this can be done with just CSS.  

Actually, we may only have to delete a line:

````
.card-slot > .card-header .card-menu-link {
  display: none;
}
.card-slot.closed-view,
.card-slot:hover {
  > .card-header .card-menu-link {
    display: inline-block;
  }
}
````

(this is from bootstrap_cards.scss, and wikirate doesn't appear to have the latest version)

Comment (ethn) [inline card/mod/03_machines/lib/javascript/wagn_mod.js.coffee]: but if we get rid of the closed-view part, it may work.

File: card/mod/05_standard/set/all/rich_html/content.rb (hunk 1 above)
Comment (ethn) [inline card/mod/05_standard/set/all/rich_html/content.rb]: doesn't this mean that if we open and close and open and close, we'll keep adding more and more of these wrappers?

Comment (ethn) [inline card/mod/05_standard/set/all/rich_html/content.rb]: one option: we take the "overflow:hidden" off the slot itself and put it on: (a) the title and (b) the body.

Comment (ethn) [inline card/mod/05_standard/set/all/rich_html/content.rb]: this may require putting max-width: 100% on the card-header (and something similar on the card-header-title)

Comment (xithan) [inline card/mod/05_standard/set/all/rich_html/content.rb]: if the slot the has no `overflow: hidden` then the body adjusts its width to its content

Comment (xithan) [inline card/mod/05_standard/set/all/rich_html/content.rb]: we can put the card slot class on the wrapper

File: card/mod/05_standard/set/all/rich_html/menu.rb [hunk 3]
```diff
@@ -118,17 +118,17 @@ def show_menu_items
       :structure  => show_structure?,
       :discuss    => disc_card && disc_card.ok?( disc_card.new_card? ? :comment : :read ),
       :page       => !main?,
-      :more       => true
+      :more       => !card.new_card?
```
ethn: this means no more menu on virtual cards.  fine for wikirate deployment, but let's give some specific attention to virtual card handling before putting out a gem release.

card/mod/05_standard/set/all/rich_html/menu.rb
```diff
@@ -118,17 +118,17 @@ def show_menu_items
...
     }
     if card.real?
       res.merge!(
-        :edit      => card.ok?(:update) || show_structure?,
+        :edit      => !card.new_card? && (card.ok?(:update) || show_structure?),
```
ethn: is this necessary?  we're already in "card.real?"  if it's real, it's not new, right?

card/mod/05_standard/set/all/rich_html/menu.rb
```diff
@@ -118,17 +118,17 @@ def show_menu_items
...
         :account   => card.account && card.ok?(:update),
         :follow    => show_follow?,
         :delete    => card.ok?(:delete)
       )
     else
-      res[:edit] = res[:structure]
+      res[:edit] = !card.new_card? && res[:structure]
```
ethn: meanwhile, this will never be true.  if it's not real, it's always new.

card/mod/05_standard/set/all/rich_html/wrapper.rb
```diff
@@ -46,15 +45,22 @@ def wrap_body args={}
...
-        #{ wrap_body args do output( yield args ) end }
-      }
+      panel args do
+        [
+          _optional_render( :header, args, :show),
+          (%{ <div class="card-subheader">#{ args[:subheader] }</div> } if args[:subheader]),
+          _optional_render( :help, args.merge(:help_class=>'alert alert-info'), :hide),
+          wrap_body(args) { output( yield args ) } ,
+        ]
+      end
```
ethn: elegant solution.

PR #408, opened 2015-04-22
use core view instead of content view for modal body

xithan:


card/mod/05_standard/set/all/rich_html/modal.rb
```diff
@@ -37,7 +37,7 @@ def default_modal_content_args args
   end
 
   view :modal_body do |args|
-    _render_content(args)
+    _render_core(args)
```
ethn: will merge.  why is there a separate view for this?

Also, I think you can just do
````
  view :modal_body, :core
````
...for the same effect

xithan: in case you want something else in the modal body than the core for some set

GerryG: In that case, use the alias form (view :model_body, :core), it will do what you want with one fewer block defined.

ethn: He tried it but it didn't work.  The difference is that the alias attaches the modal_body view to a specific core view at load time, whereas the long definition attaches to it at render time.  We should probably get clear (elsewhere) on whether that's what we want, but that's what we got!

GerryG: That actually makes some sense, it's a ruby thing to some extent. run-time vs. load time.  Not the only place this difference might matter.

PR #409, opened 2015-04-22
marginal tweaks

xithan:


xithan:
doesn't work as expected

PR #412, opened 2015-04-29
make _self refer to self in plus card searches

xithan:
http://wagn.org/make_self_refer_to_self_in_plus_card_Searches

Anything I missed in the migration?
A test produced the following result: 
````

"_self"       => "_left"
"_self+about" => "_left+about"
"_left"       => "_LL"
"_right"      => "_LR"
"_l"          => "_ll"
"_r"          => "_lr"
"_lr"         => "_llr"
"_LRLR"       => "_LLRLR"
"_RllLl"      => "_lRllLl"
"_lRlLl"      => "_llRlLl"
"_RllLlno"    => "_RllLlno"
"_lava"       => "_lava"
"left"        => "left"
"do_not_replace_self", => "do_not_replace_self",
"_selfaware"  => "_selfaware"
````

### CODE REVIEW

File: card/mod/05_standard/set/type/search_type.rb [hunk 1]
```diff
@@ -38,7 +38,7 @@ def get_query params={}
   if default_limit = query.delete(:default_limit) and !query[:limit]
     query[:limit] = default_limit
   end
-  query[:context] ||= (cardname.junction? ? cardname.left_name : cardname)
+  query[:context] ||= cardname
```
GerryG: If I understand, this gets rid of a great ugliness.  With the typical *right rule virtual card, the context card used to be left, not self.  I guess also when I create a card A+my_tag -> a Search card.  Seems like it may have been a bit broken on *self rules.
The migration looks a little scary, hope it works for all the cases ;)

ethn: Re ugliness: that's right.  See http://wagn.org/make_self_refer_to_self_in_plus_card_Searches

Re *self, I wasn't aware of any problem, but then again I probably haven't had cause to do mysearch+*self+*structure as a Search card very often; usually I'd just do mysearch.  But there could be use cases (overrides, for example).

card/db/migrate_core_cards/20150429090551_search_card_context.rb
```diff
@@ -0,0 +1,25 @@
...
+    replace = [
+      ['[lr]+','l\\1'],
+      ['[LR]+','L\\1'],
+      ["(?=[lrLR]+#{sep})(?=[LR]*[lr]+)(?=[lr]*[LR]+).*",'l\\1'],   # mix of lowercase and uppercase l's and r's
+      ['left',   'LL'],
+      ['right',  'LR'],
+      ['self',   'left'],
+      ['',       'left'],
+    ]
+    Card.search(:type=>'search').each do |card|
```
ethn: probably fine, but let's do type_id=>SearchID

card/db/migrate_core_cards/20150429090551_search_card_context.rb
```diff
@@ -0,0 +1,25 @@
...
+      if card.cardname.junction?
+        content = card.content
+        replace.each do |key, val|
+          content.gsub!(/(?<=#{sep})_(#{key})(?=#{sep})/, "_#{val}")
```
ethn: ok, what do `?<=` and `?=` mean?  something to do with greediness?

I follow and like everything but lines 9 and 19.  Don't dislike those, but I can't say I totally get them yet.

xithan: `(?<=...)` and `(?=...)` are the lookbehind and lookahead operators. They don't change the position of the regex engine and the pattern is not a part of the match.
In line 19 they make sure that there is no word character before or after the contextual name.
In line 9 the first lookahead finds contextual names containing [lrLR], the second assures that there is at least one uppercase letter and the third that there is at least one lowercase letter. 
This is necessary because I have to make sure that I don't process the completely lowercase and uppercase contextual names that I handled already in line 7 and 8.
I'm not sure if we support mixed cases though. I couldn't find the place where we process contextual names.

xithan: Good that you asked. While explaining it I noticed that `.*` in line 9 is too greedy. It should say `[lrLR]+` instead

card/db/migrate_core_cards/20150429090551_search_card_context.rb
xithan: ok, I'll change that. When I wrote that I wasn't sure if ID uses the cardname or the codename, if it's SearchID or SearchTypeID so I used type instead of type_id.

xithan: checked it: codename not cardname. makes sense. So SearchID is wrong (that's the id of *search not of the type). Let's do type_id=>SearchTypeID :)

xithan: there is a wikirate migration (update_system_emails) where we used `type_id = SearchID` :see_no_evil:

card/db/migrate_core_cards/20150429090551_search_card_context.rb
```diff
@@ -0,0 +1,29 @@
...
+        end
+        card.update_attributes! :content=>content
+      end
+    end
+    Card["*self+*right+*structure"].update_attributes! :content=>'{"name":"_left"}'
+    Card["*right+*right+*structure"].update_attributes! :content=>'{"right":"_left"}'
+    Card["*type+*right+*structure"].update_attributes! :content=>'{"type":"_left"}'
+    Card["*type plus right+*right+*structure"].update_attributes! :content=>'{"left":{"type":"_LL"}, "right":"_LR"}'
```
ethn: this looks really good.

A few questions:
1. do we care that we're not updating old versions in the history?
2. should we consider update_column to be faster?
3. is there a reason we're doing sets individually rather than adding Set cards to the Search loop?

xithan: 1. probably we should
2. yes
3. Are there other set cards that we have to change?

ethn: Re #3, now that I think about it, yes and (I hope) yes.

Any Set with a (card) rule has an actual set card associated with it, eg http://wikirate.org/Company+*type?view=raw, where the db_content is `{"type":"_self"}`.  That would, of course, be incorrect WQL for that set with the new syntax.

I *think* it should be safe, because the WQL in those cards is basically used like any WQL in a search context, eg http://wikirate.org/Company+*type+by_name

At any rate, I do suspect it's unsafe *not* to do it, because use the above pattern quite a lot.

ethn:
do we think this is ready?  ie, should be included in 1.16.5?

If so, let's run migrations on wikirate dev later this week.  If it doesn't break things on wikirate, it's probably pretty solid :)

xithan:
yes, we do :)

PR #413, opened 2015-04-30
Gemspec

ethn:
also some menu-related cleanup


card/mod/05_standard/set/all/error.rb
```diff
@@ -58,10 +58,7 @@ def rendering_error exception, view
                       %{
                         <h3>Error message (visible to admin only)</h3>
                         <p><strong>#{ exception.message }</strong></p>
-                        <div>
-                          #{exception.backtrace * "<br>\n"}
-                        </div>
-                        </div>
```
ethn: this extra </div> causes weirdness whenever there's an error.

card/card.gemspec
```diff
@@ -17,37 +13,49 @@ Gem::Specification.new do |s|
...
+  # add submodule files (seed data)
+  `git submodule --quiet foreach pwd`.split($\).each do |submod_path|
+    gem_root = File.expand_path File.dirname(__FILE__)
+    relative_submod_path = submod_path.gsub "#{gem_root}/", ''
+    Dir.chdir(submod_path) do
+      s.files += `git ls-files`.split($\).map do |filename|
+        "#{relative_submod_path}/#{filename}"
+      end
+    end
+  end
```
GerryG: Sure would be nice if there were some git helpers for gemspecs that would do this with a simpler interface.  Ugly but effective.

ethn: I couldn't swear there aren't.  Since this is mostly just used at gem build time, it's probably not the wartiest of our warts.

PR #416, opened 2015-05-06
fix coverage report that doesn't include the folder/file.rb

chuenlok:
Example: `source/source_preview.rb` will make a file `xxxx-source-source_preview.rb`. In the old filter function, it will get the name after `-` as file name which doesn't correctly point to the real files.


ethn:
please merge (sorry so slow)

PR #418, opened 2015-05-10
fixed skin images

ethn:
They were broken because they were hardcoded, but the skins no longer had codenames.


card/db/migrate_core_cards/20150510031118_fix_skin_images.rb
```diff
@@ -0,0 +1,13 @@
...
+class FixSkinImages < Card::CoreMigration
+  def up
+
+    # This is needed because the bootswatch_themes migration removed codenames.
+    # They were no longer needed for style handling but are still needed for images
+    %w{bootstrap_default cerulean cosmo cyborg darkly flatly journal lumen paper readable sandstone simplex slate spacelab superhero united yeti }.each do |theme_name|
+      theme_name = "#{theme_name}_skin"
+      Card.fetch( theme_name ).update_attributes! :codename=>theme_name
+    end
+  end
```
ethn: comment above mostly explains it.

we may be able to get rid of this migration before putting out the gem, but we have to have it to fix the broken images on cloudstore

card/db/seed
```diff
@@ -1 +1 @@
-Subproject commit 360634412ce8976570fcc2cbc9f2225223dd26c8
+Subproject commit 43c5349364a5d212a04f71d89e054f2a49cfe859
```
ethn: (reseeded data.  AWESOME that it just looks like this!!)

PR #419, opened 2015-05-15
add titles to toolbar links

ethn:


card/mod/05_standard/set/all/rich_html/toolbar.rb
```diff
@@ -162,20 +162,20 @@ def close_link css_class
...
 
-  def toolbar_button text, symbol, hide, target
-    btn_class = 'btn btn-default'
+  def toolbar_button text, symbol, hide, tag_args
+    tag_args[:class] = [ tag_args[:class], 'btn btn-default' ].compact * ' '
+    tag_args[:title] ||= text
     link_text = "#{glyphicon symbol}<span class='menu-item-label #{hide}'>#{text}</span>"
 
-    if target[:page]
-      card_link target[:page], :class=>btn_class, :text=>link_text, :path_opts=>{:slot=>{:show=>:toolbar}}
```
ethn: unless I'm mistaken, we don't ever really want the "show toolbar" here.

card/mod/05_standard/set/all/rich_html/menu.rb
```diff
@@ -2,7 +2,7 @@
   view :menu, :denial=>:blank, :tags=>:unknown_ok do |args|
     return _render_template_closer if args[:menu_hack] == :template_closer
     return '' if card.unknown?
-    wrap_with :div, :class=>'menu-slot' do
+    wrap_with :div, :class=>'menu-slot nodblclick' do
```
ethn: was finding double clicks on menu led to edit view

GerryG: Isn't that a good thing?  That would be selecting the first menu item, no?  Seems like what you would want that action to do.

ethn: makes rational sense, but it was confusing, even for me.   If I want to double-click, I can usually do it pretty much anywhere else.

PR #420, opened 2015-05-25
Add space/nbsp reordering on clean

GerryG:
I turned the feature on by default (it has a config to turn it off).  Can't test it on and off in one run, but the specs are there.

### DIFF
--- a/card/config/initializers/recaptcha.rb
+++ b/card/config/initializers/recaptcha.rb
@@ -3,5 +3,5 @@ Recaptcha.configure do |config|
   config.public_key  = Cardio.config.recaptcha_public_key  || nil
   config.private_key = Cardio.config.recaptcha_private_key || nil
   config.proxy       = Cardio.config.recaptcha_proxy       || nil
-  config.api_version = 'v1'
+  config.api_version = 'v1' if config.respond_to?(:api_version=)
 end
--- a/card/lib/card/content.rb
+++ b/card/lib/card/content.rb
@@ -179,6 +179,12 @@ class Card
         end.gsub(/<\!--.*?-->/, '')
       end
 
+      if Card.config.space_last_in_multispace
+        def clean_with_space_last! string, tags = ALLOWED_TAGS
+          clean_without_space_last!(string, tags).gsub(/(?:^|\b) ((?:&nbsp;)+)/, '\1 ')
+        end
+        alias_method_chain :clean!, :space_last
+      end
       def truncatewords_with_closing_tags(input, words = 25, truncate_string = "...")
         if input.nil? then return end
         wordlist = input.to_s.split
--- a/card/lib/cardio.rb
+++ b/card/lib/cardio.rb
@@ -34,6 +34,7 @@ module Cardio
       config.recaptcha_public_key  = nil
       config.recaptcha_private_key = nil
       config.recaptcha_proxy       = nil
+      config.api_version           = 'v1'
 
       config.cache_store           = :file_store, 'tmp/cache'
       config.override_host         = nil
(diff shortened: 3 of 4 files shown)


card/config/initializers/recaptcha.rb
```diff
@@ -3,5 +3,5 @@
   config.public_key  = Cardio.config.recaptcha_public_key  || nil
   config.private_key = Cardio.config.recaptcha_private_key || nil
   config.proxy       = Cardio.config.recaptcha_proxy       || nil
-  config.api_version = 'v1'
+  config.api_version = 'v1' if config.respond_to?(:api_version=)
```
GerryG: This was breaking for me.  Probably because of gem versions, but this should allow it to work either way.

card/lib/card/content.rb
```diff
@@ -179,6 +179,12 @@ def clean!( string, tags = ALLOWED_TAGS )
         end.gsub(/<\!--.*?-->/, '')
       end
 
+      if Card.config.space_last_in_multispace
```
GerryG: It seemed best to have this happen at the same time as tag cleaning.  It doesn't make sense except when you have entity encoding, which is generally HTML contexts where we do tag cleaning.

card/lib/cardio.rb
```diff
@@ -34,6 +34,7 @@ def set_config config
       config.recaptcha_public_key  = nil
       config.recaptcha_private_key = nil
       config.recaptcha_proxy       = nil
+      config.api_version           = 'v1'
```
GerryG: This was a mistake, removing it.

GerryG:
Implements this: http://wagn.org/Better_Handling_of_Sentence_Spacing

card/spec/lib/card/content_spec.rb
```diff
@@ -306,6 +306,18 @@
...
+      it 'fixes regular nbsp order by default' do
+        assert_equal 'space&nbsp; test&nbsp; two&nbsp;&nbsp; space',
+                      Card::Content.clean!('space&nbsp; test &nbsp;two &nbsp;&nbsp;space')
+      end
+
+      it 'doesn\'t fix regular nbsp order with setting' do
+        # manually configure this setting, then make this one live (test above will then fail)
+        pending "Can't set Card.config.space_last_in_multispace= false for one test"
+        assert_equal 'space&nbsp; test &nbsp;two &nbsp;&nbsp;space',
+                      Card::Content.clean!('space&nbsp; test &nbsp;two &nbsp;&nbsp;space')
```
ethn: Can't you just set the config value right before the test and set it back afterwards (probably in an ensure block)?

card/config/initializers/recaptcha.rb
ethn: would rather fix the minimum gem version but can let this through for now.

card/spec/lib/card/content_spec.rb
GerryG: I'm not remembering exactly what was happening, but it was a problem changing it.  I think it might be more how it was implemented than how the config setting works.  Once it is used, the code can't change back to the other way of processing.  I think a method is overloaded or something like that.

GerryG: Maybe we could force it to re-load the module or something.

ethn: Are you sure?  If the code works, then turning it on and off should work, no?

GerryG:
Turning it on/off works, just not after it has been tested.  Either you have to do it in a separate rspec run (i.e. change the setting, enable the other test case and run again), or do some re-load.  The test situation never comes up in real life 'cause you can't change it after startup.
BTW, is this on or off on GC sites?  Can we even control this other than globally on the virtual host?

PR #421, opened 2015-05-26
some fixes and a "insert item" event for pointers

xithan:


card/mod/02_basic_types/set/type/pointer.rb
```diff
@@ -4,6 +4,11 @@
   self.drop_item Env.params['drop_item'] if Env.params['drop_item']
 end
 
+event :insert_item_event, :before=>:approve, :on=>:save, :when=> proc {|c| Env.params['insert_item']} do
+  index = Env.params['item_index'] || 0
+  self.insert_item index.to_i, Env.params['insert_item']
+end
+
```
xithan: This is the only siginificant change in this PR. I need this for wikirate's profile pages to change position of pointer items. Maybe arguable whether we want to support this in wagn.

ethn: it occurs to me (a bit late) that we might want to mimic Array methods pretty closely here.  That would just be insert(index, card_or_name [, card_or_name2])

ethn: (obviously very close)

GerryG: Seems like a good idea.  Note the 'self.' on line 9 and similar on line 4 (which didn't change) isn't needed.  At least I don't think they are, and generally I think we should avoid self. whenever possible.  Otherwise I start looking for a reason it is needed.

ethn: probably right that it's not needed.

in my experience, the most common time this is needed is when you're doing assigments.  Imagine, for example, that you've got an accessor like "age".  When you do:
````
def gonna_die_in
  death_age - age
end
````
...there's no need for self.  But if you do:
````
def have_birthday
  self.age = age + 1
end
````
...then you need that "self" to avoid creating a local variable "age"

GerryG: Yes, assignment is the main place that it is ambiguous between creating a new local and sending a method.  Language-wise, it is a bit odd that an un-referenced local assignment has higher precedence than an assignment method, but it does.  It might be expensive to test for this case in all the places you would have to in order to get it right.

I can't recall any cases besides assignment and self.class to avoid the class keyword interpretation.

PR #423, opened 2015-05-30
toolbar improvements

xithan:


card/mod/05_standard/set/type/session.rb
```diff
@@ -0,0 +1,26 @@
...
+event :update_in_session, :after=>:approve, :on=>:update do
+  if db_content_changed?
+    Env.session[key] = db_content
+    self.db_content = ''
+  end
+end
+
+event :delete_in_session, :after=>:approve, :on=>:delete do
+  Env.session[key] = nil
+  abort :success
```
xithan: I'm not sure whether deleting a session card should delete only the session value or the card as well.

card/lib/card/view_name.rb
```diff
@@ -0,0 +1,44 @@
...
+          @parts = [str]
+          @simple = true
+          str.empty? ? '' : simple_key
+        end
+      @@name2viewnameobject[str] = self
+    end
+
+    def simple_key
+      decoded.underscore.gsub(/[^#{OK4KEY_RE}]+/,'_').split(/_+/).reject(&:empty?)*'_'
+    end
```
xithan: Card::ViewName doesn't singularize in contrast to Card::Name. I had to copy some code from smartname to introduce a new cache to avoid conflicts  (`@@name2viewnameobject` instead of `@@name2nameobject`).

card/mod/05_standard/set/type/session.rb
ethn: I suppose it begs the question of how you would ever delete the card itself.

I don't feel strongly (and wouldn't want to hold anything up for this), but I guess I'm tempted to say we should go ahead and delete it unless there's a good reason to have the card around.

card/lib/card/view_name.rb
ethn: btw, I think the plan is to rename the smartname gem to card_name and move it into this same repository.  Rationale for card_name vs. cardname here: http://guides.rubygems.org/name-your-gem/

Within Wagn/Decko, Card::Name (and Card::ViewName) should inherit from CardName (the gem class), imo.

ethn: Gerry and I are discussing possible future directions with view names here: http://wagn.org/Link_Override_Rule.  I don't know if that will impact your ideas on singularization.

card/mod/03_machines/lib/javascript/wagn_mod.js.coffee
```diff
@@ -163,6 +163,38 @@ $(window).ready ->
...
+      type : 'PUT'
+      data : 'card[content]=false'
+
+  $('body').on 'click', '.toolbar-pin.inactive > a', (e) ->
+    e.preventDefault()
+    $('.toolbar-pin').removeClass('inactive').addClass('active')
+    $.ajax '/*toolbar_pinned',
+      type : 'PUT'
+      data : 'card[content]=true'
+
```
ethn: I don't see anything problematic here, though I have a hunch we might be able to break some of this out into shared methods, no?

card/lib/card/view_name.rb
GerryG: Singularization is a bit of a wart on names and not easy to internationalize.  I'm not really sure where we should be going on that.  I think that the parts and even words (of a part) are potentially contextualized so that key translation can depend on the "kind of" or type for the name (or name part).
I was messing about with Smartname trying to do something smarter with dates and numbers that are parts of names.  I think it is something we will wan't eventually, but there are some conceptual issues to resolve first.

GerryG: It would be better if we didn't have to clone some of the code in smartname.  Is there a clean way to make that processing optional?  In the end, I think it will be really hard to do something like this without it being part of the way names are handled generally.  What I mean is that if we ever will have the view names actually represented by cards, and it seems that we will,, then we can't use an option to do view name translation to codenames that isn't in effect whenever we are doing name lookup or we won't be able to find cards with plural parts to their keys.
If we process view names the same as cardnames are processed now, what is the downside?  I think it is just that we sometimes have odd looking codenames for some view, right?  That's a lot easier to live with than inconsistent handling of name to key translations.

ethn: Good thoughts.  Philipp, correct me if I'm wrong, but I think he mostly just wanted to change the name handling to get rid of some odd-looking names in code, as you say.  Specifically, he was having to do things like `default_edit_rule_args` when the view was `edit_rules`.

I couldn't immediately come up with (or remember) a strong argument for why view names *had* to be treated like cardnames, so it kind of came down to deciding which was the bigger wart: ugly method names or the ViewName code.

I do think the SmartName code could be made more adaptable, and that we'll probably do a better job of maintaining that once it's in the same repo.

In general it feels like we're losing a little bit of our simplicity in all these upgrades and may need to fight to gain it back.  The good news is that this isn't really affected much by data (unless there are references to different name variants in nest syntax in folks' cards), so we're a bit more pliant here.

GerryG: So, you will move the gem to the wagn repo so we'll have four now?  It shouldn't be that big of a deal to maintain either way, but I think together makes sense and follows that rails platform pattern.  Folks could still use any subset of the 4 like they would will rails gems.
I think there could be some real advantages to using name logic for views.  It's much stronger if we actually have cards for views, but I think it is mostly a good thing.  We are conflating ruby (and rails) inflections for constants vs. identifiers where singular/plural in particular is used differently with our name processing which is about 1) folding equivalence classes (as defined by key identity) and 2) translating between presentation names and internal (code) names.
Oh, that reminds me, codenames really are arbitrary.  Generally we start with the codename being the key for the default name, but there is no logical requirement for this.  We do the singular processing for 1), to fold from all the possible names to the much smaller space of keys, but I think this issue is all about 2), having a stable codename.

ethn: Whoa, that's a great point.  If we do eventually have view cards, then they would be mapped to coded views via codenames.  So, **there's no great reason to support variable view names in code references to views**, is there? It's just extra processing.

The only time you'd really want to do view name processing is in interpreting views as specified in content (nest syntax). And for now, we can probably get away without supporting view name variants in nests.

Fwiw, we do still need to handle (even today) the old aliased/deprecated names (eg "core" used to be "naked" and is still probably used in some old sites).  But I'm not sure there's much need for name objects without supporting name variants, is there?

GerryG: I think we are saying more or less the same thing.  I don't see any need for variants in code, and my point was that if you want the plural for the codename, you can have that, but only that codename (pluralized).  You could actually create two cards with codenames (i.e. edit_rule and edit_rules could be codenames of distinct cards, although that is probably a bad idea to do) that would match as names.

I thought that's why we were using name processing for view names (to get the code/user context translation on view names)?

Can't we implement a migration to convert those old views?  Presumably, these need to be supported as user space names (i.e. in inclusion/nest syntax), and not as codenames?

ethn: We could do a migration, sure.  Right - don't need/want to support them in code.

Nobody seems to know why exactly we were doing name processing on viewnames. It doesn't really handle the "code/user context translation" with any sophistication.

GerryG: Well, I'm not sure, but I'd bet that you added it in the first place for some reason.  Maybe you should make something up :)

GerryG: Oh, and something I didn't notice because I didn't look at the whole context.  Smartname is designed so you can supply your own method for singularize, so all you need to do is configure that method with a method that doesn't change its input argument.

GerryG: Just noticing some cruft in Smartname.  There is a method that isn't used to do some kind of variable substitution (complete with a comment that it shouldn't be there).  The 'var_re' is broken anyway, not sure when it was last in use.  I also note that we should just update the banned list in the gem.  Seems that it is only the 'alternative' joint that's in the list now.  When were the others removed from the list?

I guess you can't configure smartname to use a different "uninflect" for different uses, so that isn't an option after all.

PR #424, opened 2015-06-01
handle error ajax responses, and trigger slot ready while modal is shown

chuenlok:


chuenlok:
In bootstrap modal, sometimes we get the modal content from ajax. If the ajax returns errors ,it won't show the content but a full window grey layer. For example, a user tries to show a forbidden page in the modal, the response http code is not 200 but some other numbers. Then the response body won't be shown in the modal. This fix is to handle this and also trigger the slot ready for the new added content.

card/mod/06_bootstrap/lib/javascript/bootstrap_modal_wagn.js
```diff
@@ -0,0 +1,27 @@
+wagn.slotReady(function(slot) {
+  slot.find('.modal.fade').on('loaded.bs.modal', function(e) {
+    $(this).trigger('slotReady');
+  });
+  slot.find('[data-toggle=\'modal\']').off("click").on('click', function(e) {
```
ethn: why do we need to click the modal off and on again?

card/mod/06_bootstrap/lib/javascript/bootstrap_modal_wagn.js
```diff
@@ -0,0 +1,27 @@
...
+      type: 'GET',
+      success: function(html) {
+        $(modal_selector + ' .modal-content').html(html);
+        return $(modal_selector).trigger('loaded.bs.modal');
+      },
+      error: function(jqXHR, textStatus) {
+        $(modal_selector + ' .modal-content').html(jqXHR.responseText);
+        return $(modal_selector).trigger('loaded.bs.modal');
+      }
+    });
```
ethn: I don't really understand what this is yet.  No comments or anything.

I would think that slotReady usually gets called *after* receiving an AJAX response.  But this looks like it's triggering a new request.  How come?

ethn: Perhaps you could break this out into some thoughtfully named functions?

card/mod/06_bootstrap/lib/javascript/bootstrap_modal_wagn.js
chuenlok: The `off("click")` can remove all binded handler from bootstrap. We need to override the onclick handler on all `'[data-toggle=\'modal\']'` to prevent bootstrap from getting content with `load`.

chuenlok:
Let me explain a little bit about the bootstrap modal.

The structure of a bootstrap modal link is like 
```
<a data-toggle="modal" href="remote.html" data-target="#modal">Click me</a>
```
While the anchor is clicked, bootstrap tries to load the `remote.html` by calling JQuery's `load`. 

```
if (this.options.remote) {
      this.$element
        .find('.modal-content')
        .load(this.options.remote, $.proxy(function () {
          this.$element.trigger('loaded.bs.modal')
        }, this))
    }
```
from bootstrap source code. 

The problem of this code is it is not going to load the content from the `remote.html` if `remote.html` return a non-http-success response, like 403 forbidden. 

In the web interface, you may see a grey layer on top of the webpage. 

The code I suggest is to handle the non-http-success response case. 

```
$(modal_selector).modal('show');
    $.ajax({
      url: href,
      type: 'GET',
      success: function(html) {
        $(modal_selector + ' .modal-content').html(html);
        return $(modal_selector).trigger('loaded.bs.modal');
      },
      error: function(jqXHR, textStatus) {
        $(modal_selector + ' .modal-content').html(jqXHR.responseText);
        return $(modal_selector).trigger('loaded.bs.modal');
      }
    });
```
It will show the modal window and get the html content by a ajax call while the anchor is clicked. If success, the content will be loaded to the modal content. If fail, it will show the responseText in the modal content which the default modal is not going to do. 

I will check the slot ready trigger again. I remember the content in modal is raw while loaded.

ethn:
Can you add all that as a comment in the js file?

This all still feels very hacky to me.  I think perhaps we should keeping use the bootstrap modal handling but not the AJAX handling.  We could handle the AJAX ourselves as part of the standard slot calls but call the modal method after we have the results.

I'll merge this, but it doesn't feel great.

card/mod/06_bootstrap/set/self/bootstrap_js.rb
```diff
@@ -1,5 +1,7 @@
 view :raw do |args|
-  File.read "#{Cardio.gem_root}/mod/06_bootstrap/lib/javascript/bootstrap.js"
+  [ "bootstrap.js", "bootstrap_modal_wagn.js" ].map do |name|
+    File.read "#{Cardio.gem_root}/mod/06_bootstrap/lib/javascript/#{name}"
+  end.join("\n")
```
ethn: seems like we should keep the bootstrap_js pure and add this handling to the wagn_mod js. (The idea has been that we will eventually break out the wagn_mod stuff into different mods)

PR #425, opened 2015-06-01
improve rule navigation

xithan:
deals with 2. - 4. of  http://wagn.org/improve_rule_navigation
fixes also http://wagn.org/Cancel_on_Rule_Edit_is_Broken


card/mod/04_settings/lib/card/setting.rb
```diff
@@ -8,34 +8,36 @@ module Setting
...
+      :permission    => "Permissions",
+      :event         => "Events",
+      :webpage       => "Webpage",
+      :templating    => "Templating",
+      :editing_cue   => "Editing cues",
       :pointer       => "Pointer",
-      :other         => "Other"  
+      :other         => "Other"
     }
+    @@groups = @@group_names.keys.inject({}) { |groups, key| groups[key] = []; groups }
```
GerryG: This is fine, but I've found each_with_object is usually simpler than inject.  Each with object only works as expected when you are mutating the object, which you are often when loading a hash.  So, @@group_names.keys.each_with_object({}) {|key, groups| groups[key] = [] }

xithan:
part of #443

PR #426, opened 2015-06-02
support new recaptcha version

xithan:


card/db/migrate_core_cards/20150914134339_admin_info_card.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    Card[:recaptcha_settings].update_attributes! :type_id=>Card::PointerID,
+      :content=>"[[+public key]]\n[[+private key]]\n[[+proxy]]"
+
+    ['public_key', 'private_key', 'proxy'].each do |name|
+      card = Card["recaptcha_#{name}".to_sym]
+      card.update_attributes! :name=>"#{Card[:recaptcha_settings].name}+#{name.gsub('_',' ')}"
+    end
+    Card.fetch('*recaptcha settings+*self+*structure').delete
+
```
ethn: We need to delete the *script output (if not build a system to detect file updates; probably not yet, given the time crunch...)

card/mod/04_settings/set/self/recaptcha_public_key.rb
```diff
@@ -0,0 +1,9 @@
+event :update_recaptcha_public_key, :after=>:store do
```
ethn: I'm pretty confused about these new configurations.  Are they Settings or not?  In the migration, it looks like they're Phrase cards, but here we're using `extend Card::Setting`.  A recaptcha public key isn't set-specific, is it?

If not, why do we assign them to groups?

wagn/lib/wagn/generators/wagn/templates/config/application.rb
```diff
@@ -8,8 +8,9 @@ class Application < Wagn::Application
 
     # Wagn inherits Rails configuration options.  See http://guides.rubyonrails.org/configuring.html
 
-    config.recaptcha_public_key  = '6LdhRssSAAAAAFfLt1Wkw43hoaA8RTIgso9-tvtc'
-    config.recaptcha_private_key = '6LdhRssSAAAAAGwzl069pJQBdmzCZigm1nV-dmqK'
+    config.recaptcha_public_key  = '6LeoHfESAAAAAN1NdQeYHREq4jTSQhu1foEzv6KC'
+    config.recaptcha_private_key = '6LeoHfESAAAAAHLZpn7ijrO4_KGLEr2nGL4qjjis'
```
ethn: If we're migrating these into cards anyway, shouldn't we just make these the default card content and get rid of the code configuration option?

ethn:
Do I understand correctly that this follows the same interface pattern as our current one; it just uses the new api?

xithan:
The interface pattern didn't change. You click on submit and then you get the recaptcha.

PR #428, opened 2015-06-09
better prepend/append handling

ethn:
hodgepodge of minor cleanups

### DIFF
--- a/card/lib/card/query.rb
+++ b/card/lib/card/query.rb
@@ -3,7 +3,7 @@
 class Card::Query
   require_dependency 'card/query/clause'
   require_dependency 'card/query/card_clause'
-  require_dependency 'card/query/value_clause'  
+  require_dependency 'card/query/value_clause'
   require_dependency 'card/query/ref_clause'
 
   MODIFIERS = {};  %w{ conj return sort sort_as group dir limit offset }.each{|key| MODIFIERS[key.to_sym] = nil }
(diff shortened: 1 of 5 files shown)


card/lib/card/query.rb
```diff
@@ -16,43 +16,53 @@ class Card::Query
...
-            Card.fetch cardname, :new=>{}
-          else
-            Card[ row['name'] ]
-          end
-        card.nil? ? Card.find_by_name_and_trash(row['name'],false).repair_key : card
+    if retrn == 'card'
+      simple_run('name').map do |name|
+        Card.fetch name, :new=>{}
+      end
+    else
```
ethn: easier to see what this is doing now.

card/lib/card/query.rb
```diff
@@ -16,43 +16,53 @@ class Card::Query
...
+
+  def simple_run retrn
+    rows = run_sql
+
+    case retrn
+    when 'name' #common case
+      if query[:prepend] || query[:append]
+        rows.map do |row|
+          [ query[:prepend], row['name'], query[:append] ].compact * '+'
+        end
```
ethn: key difference is that append/prepend works even here.

card/mod/01_core/set/all/permissions.rb
```diff
@@ -1,5 +1,5 @@
 
-Card.error_codes.merge! :permission_denied=>[:denial, 403], :captcha=>[:error,449]
+Card.error_codes.merge! :permission_denied=>[:denial, 403], :captcha=>[:errors,449]
```
ethn: this is a bug introduced by the new view handling

card/mod/01_core/set/all/templating.rb
```diff
@@ -12,29 +12,34 @@ def template
...
-      dup_card.type_id = default_card ? default_card.type_id : Card.default_type_id
-
+      dup_card.type_id = default ? default.type_id : Card.default_type_id
 
-      if content_card = dup_card.structure_rule_card
+      if structure = dup_card.structure_rule_card
         @virtual = true if junction?
-        content_card
+        self.type_id = structure.type_id if assign_type_to?(structure)
+        structure
```
ethn: the above seeks to address a bug that can occur when a structured card is given an explicit type.  That error is caught in validation, but it was screwing up virtual cards.

card/mod/03_machines/set/type/scss.rb
```diff
@@ -2,21 +2,23 @@
...
-  
+
   view :core do |args|
     compile_scss(process_content _render_raw)
   end
-  
+
   def compile_scss scss, style=:expanded
+    Rails.logger.info "----------------\n\ncompress SCSS: #{scss}"
+
```
ethn: oops.  got to get rid of this.

card/mod/01_core/set/all/templating.rb
```diff
@@ -12,29 +12,34 @@ def template
...
       dup_card = dup
-      dup_card.type_id = default_card ? default_card.type_id : Card.default_type_id
-
+      dup_card.type_id = default ? default.type_id : Card.default_type_id
 
-      if content_card = dup_card.structure_rule_card
+      if structure = dup_card.structure_rule_card
         @virtual = true if junction?
-        content_card
+        self.type_id = structure.type_id if assign_type_to?(structure)
```
GerryG: Is this new?  Doesn't seem related to the main ticket.

When does this happen, on a real? card that also has a structure rule?

ethn: no, it's a virtual card thing.  In general, you're not allowed to set the type of a virtual card to be something other than what the structure rule defines.  We had a weird case where the url (or nest or something) was trying to override that type.

That's prevented in validation for real cards.

True that it shouldn't really be in this pr. my bad.

GerryG: You know I'm not against slipping a small fix in with some other changes :)

I know it was about virtual cards, but only when they are real? or !real?   Is it just about the type_id on the card object (in memory), so mainly the !real? case?  Since nests and URLs probably don't effect the type when it is a real card ...

ethn: By definition, virtual cards are never real; this isn't a problem for real cards.  (Virtual cards are a subset of structured cards.)

It's actually not so much about caching/memory as initialization.  We weren't using the type from the structured card until the validation process, but since virtual cards are never saved (and thus never validated), it wasn't getting set correctly when there was an explicit type given.

GerryG: Sorry about the imprecise language again, but you understood.  It just feels to me like it is still pretty virtual when the structure rule is over a real card, but that is the definition we use for coding reasons like this.

Makes sense, for this case we would never do validation, so the type never gets set on the actual card (in memory, since by definition it doesn't get saved).  The case of a virtual card getting saved always worked because it would happen before validation sometime or maybe even in validation.

PR #429, opened 2015-06-09
improved performance logger

xithan:
Would be good to separate this into a developer mod but we need a "read event" for that to replace the before_filter and after_filter in the card controller.

### DIFF
--- a/card/lib/card/log.rb
+++ b/card/lib/card/log.rb
@@ -192,33 +192,6 @@ class Card::Log
         end
       end
 
-
-      def render_html_file content
-        %{
-          <html>
-          <head>
-            <!-- Latest compiled and minified CSS -->
-            <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css">
-
-            <!-- Optional theme -->
-            <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap-theme.min.css">
-
-            <!-- Latest compiled and minified JavaScript -->
-            <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.2/jquery.min.js"></script>
-            <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/js/bootstrap.min.js"></script>
-            <style>
-            .panel-group {
-              margin-bottom: 0;
-            }
-            </style>
-          </head>
-          <body>
-          #{content}
-          </body>
-          </html>
-        }
-      end
-
       def text_log
         @@log.each do |entry|
           Rails.logger.wagn entry.to_s! if entry.valid
--- a/card/lib/card/log.rb
+++ b/card/lib/card/log.rb
@@ -45,7 +45,7 @@ class Card::Log
     # config.performance_logger = {
     #     :min_time => 100,                              # show only method calls that are slower than 100ms
     #     :max_depth => 3,                               # show nested method calls only up to depth 3
-    #     :details=> true                                # show method arguments and sql
+    #     :details=> true,                                # show method arguments and sql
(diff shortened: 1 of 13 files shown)


card/mod/01_core/set/all/collection.rb
```diff
@@ -57,6 +57,27 @@ def include_item? item
...
+    self.content = new_names.empty? ? '' : new_names.join("\n")
+  end
+end
+
+def insert_item index, name
+  new_names = item_names
+  new_names.delete(name)
+  new_names.insert(index,name)
+  self.content =  new_names.join "\n"
+end
```
xithan: We have methods like `item_names` and `include_item?`in the all set.  No reason to restrict the other item methods to pointers.

card/mod/05_standard/set/self/performance_log.rb
```diff
@@ -0,0 +1,92 @@
...
+        if File.exists? path
+          File.read(path)
+        end
+      end.compact.join "\n"
+    browser_data = CSV.parse(File.read(card.csv_path))
+    output [
+      table(browser_data, true),
+      wagn_data
+    ]
+  end
```
xithan: The logs are too big for the cards table. I save them in files and store the file names in this pointer card. The core view renders the content of all these files.

card/mod/05_standard/set/self/performance_log.rb
```diff
@@ -0,0 +1,92 @@
...
+            end
+
+    %{
+      <table class="table">
+        #{thead}
+        #{tbody}
+      </table>
+    }
+  end
+end
```
xithan: One could make a table view out of this that renders CSV data as html tables (columns comma separated, rows newline separated - a 2-dimensional pointer card :)

wagn/rails/controllers/card_controller.rb
```diff
@@ -135,11 +135,29 @@ def request_logger
...
-    Card::Log::Performance.stop
+    if Wagn.config.performance_logger || params[:performance_log]
+      Card::Log::Performance.stop
+    end
+    if params[:perfomance_log]
+      Wagn.config.log_level = @old_log_level
+      if Wagn.config.performance_logger
+        Card::Log::Performance.load_config Wagn.config.performance_logger
+      end
+    end
```
xithan: The performance logger can now be activated and configured for a single request with query arguments in the url. With `performance_log[output]=card` the output will be saved in the `*performance log` card otherwise it will appear in the server log.

card/lib/card/log.rb
```diff
@@ -266,18 +347,22 @@ def initialize( parent, level, args )
...
+          @parent.add_children self
           #@sibling_nr = @parent.children_cnt
         end
       end
 
-      def add_children
+      def add_children child=false
         @children_cnt += 1
+        @children << child if child
       end
```
xithan: For the text output the number of child used to be enough but to generate html output I need a proper tree structure.

card/lib/card/log.rb
```diff
@@ -166,11 +183,74 @@ def enabled_method? method_name
...
+              </div>
+              <div id="#{collapse_id}" class="panel-collapse collapse #{'in' if entry.duration > 100}" role="tabpanel" aria-labelledby="heading-#{collapse_id}">
+                <div class="panel-body">
+                  #{ panel_body }
+                </div>
+              </div>
+            </div>
+          </div>
+          }
+      end
```
xithan: Maybe we should start using a bootstrap rails gem like https://github.com/seyhunak/twitter-bootstrap-rails that offers helper methods for bootstrap components.

xithan: the bh gem seems to be a reasonable option (http://fullscreen.github.io/bh/)

card/mod/01_core/set/all/collection.rb
ethn: I like it.

PR #431, opened 2015-06-10
change from absolute path to relative path

chuenlok:
I am testing it.


chuenlok:
tested. It is good to go!

PR #433, opened 2015-06-13
rails 4

xithan:


card/db/migrate_core_cards/20150611203506_rails_inflection_updates.rb
```diff
@@ -0,0 +1,55 @@
...
+          content_changed = true
+          new_content = new_content.gsub(wrong_sing, correct_sing)
+        end
+      end
+      if content_changed
+        card.update_attributes! :content=>new_content
+      end
+    end
+  end
+end
```
xithan: Rails 4 has some inflection updates (https://github.com/rails/rails/commits/4-2-stable/activesupport/lib/active_support/inflections.rb). The migration updates the keys in the cards tables and in the content of css, scss and html cards.

card/db/migrate_core_cards/20150611203506_rails_inflection_updates.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def up
+    card_names = Card.pluck(:name)
+    apply_to_content = ::Set.new
+    corrections =   [
+             # plural,     wrong singular,  correct singular
+      word([ '(\w+)lice',  '(\w+)louse',    '\1lice' ]),
+      word([ '(\w+)mice',  '(\w+)mouse',    '\1mice' ]),
+      word([ 'kine',       'cow',           'kine']),
```
xithan: Rails 4 dropped support for more than one cow :) And the plural of Moscow is no longer Moskine.

card/lib/card/migration.rb
```diff
@@ -42,8 +42,10 @@ def schema_mode mig_type=type
       original_suffix = ActiveRecord::Base.table_name_suffix
 
       ActiveRecord::Base.table_name_suffix = new_suffix
+      ActiveRecord::SchemaMigration.reset_table_name
       yield
       ActiveRecord::Base.table_name_suffix = original_suffix
+      ActiveRecord::SchemaMigration.reset_table_name
     end
```
xithan: Rails 4 introduced an ActiveRecord model for the migration table. So changing the table name suffix is not sufficient to change the migration table. The SchemaMigration class remains connected to the old table.

card/lib/card/migration.rb
```diff
@@ -98,50 +100,24 @@ def migration_paths
...
-
-    case direction
-    when :up   then announce "migrated (%.4fs)" % time.real; write
-    when :down then announce "reverted (%.4fs)" % time.real; write
+    else
+      contentedly { send(direction) }
     end
+  ensure
+    @connection = nil
   end
```
xithan: Rails 4 moved the code that we have to change for our card migrations from the `migrate` method to the `exec_migration` method

card/mod/01_core/set/all/phases.rb
```diff
@@ -69,24 +69,13 @@ def identify_action
...
-
-
 
 def store
   run_callbacks :store do
     yield #unless @draft
-    store_changes
     @virtual = false
   end
+  run_callbacks :stored
```
xithan: `store changes` is only needed for the history. To move it to the history mod I introduced a new hook `stored`

card/mod/05_standard/set/all/event_viz.rb
```diff
@@ -2,10 +2,14 @@
...
 #  not ready for prime time!
+#  not working with Rails 4
 
 def events action
   @action = action
-  root = _validate_callbacks + _save_callbacks
+
+  # don't know what this line is for but it breaks with Rails 4
+  # root = _validate_callbacks + _save_callbacks
+
```
xithan: not sure whether this is worth to put more effort in to get it working again

card/mod/05_standard/set/all/rich_html/form.rb
```diff
@@ -25,11 +25,14 @@ def edit_slot args={}
...
-    builder = ActionView::Base.default_form_builder
     card.name = card.name.gsub(/^#{Regexp.escape(root.card.name)}\+/, '+') if root.card.new_card?  ##FIXME -- need to match other relative inclusions.
 
-    builder.new("card[subcards][#{card.relative_name}]", card, template, {}, block)
+    # doesn't work anymore in Rails 4
+    # TODO -- check whether forms work with the new instantiate_builder call
+    # block = Proc.new {}
+    # builder = ActionView::Base.default_form_builder
+    # builder.new("card[subcards][#{card.relative_name}]", card, template, {}, block)
+    builder = instantiate_builder("card[subcards][#{card.relative_name}]", card, {})
```
xithan: makes me nervous that I couldn't use all arguments here that we used before. What did the template argument in the old builder initialization?

wagn/lib/decko/engine.rb
```diff
@@ -20,7 +21,7 @@ class Engine < ::Rails::Engine
 
     paths.add "app/controllers", :with => 'rails/controllers', :eager_load => true
     paths.add 'gem-assets',      :with => 'rails/assets'
-    paths.add 'config/routes',   :with => 'rails/engine-routes.rb'
+    paths.add 'config/routes.rb',   :with => 'rails/engine-routes.rb'
```
xithan: the 'config/routes' key is not used anymore in Rails 4 but 'config/routes.rb' seems to work.

wagn/lib/wagn/generators/wagn/templates/Gemfile
```diff
@@ -107,6 +107,7 @@ group :debug do
 	else
 		gem 'debugger'
 	end
+  gem 'better_errors', :git=>"https://github.com/xithan/better_errors.git"
```
xithan: I forked this gem to correct the tmp paths in the errors messages

card/mod/05_standard/spec/set/all/history_spec.rb
```diff
@@ -68,11 +68,41 @@
...
+          end
+        end
+
+        it "doesn't create an action" do
+          Card::Auth.as(:anonymous) do
+            action_count = Card::Action.count
+            Card.create :name=>"create fail"
+            expect(Card::Action.count).to eq action_count
+          end
+        end
```
xithan: you were right, Ethan, if the transaction fails it takes care of other stuff that was created in the same process

wagn/rails/engine-routes.rb
```diff
@@ -1,49 +1,52 @@
...
   # ~~~~~~~~~~~~~~~~~~~~~~~~~~
-  
-  # standard non-RESTful
-  match '(card)/:action(/:id(.:format))' => 'card', :action => /create|read|update|delete|asset/
 
+  # standard non-RESTful
+  get '(card)/:action(/:id(.:format))'  => 'card', :action => /create|read|update|delete|asset/
+  match '(card)/create(/:id(.:format))' => 'card#create', :via=>[:post, :patch]
+  match '(card)/update(/:id(.:format))' => 'card#update', :via=>[:post, :put, :patch]
+  match '(card)/delete(/:id(.:format))' => 'card#delete', :via=>:delete
```
xithan: cucumber doesn't work without those routes

card/lib/card/migration.rb
GerryG: The gist of this relates to us having two different sets of migrations tracked in their own tables, right?  This fixes the table switching for rails 4 ...

card/lib/card/migration.rb
GerryG: What do we actually change?  I'm guessing you have copied/customized a rails method.  Nice that the new method is way simpler.

xithan: We wrap  `send(direction)` in `contentedly`. `send(direction)` is the call of the `up` method in the migration. `contentedly` does some Wagn stuff we need for migrations (clearing the cache, setting permissions, ...)

card/lib/card/migration.rb
xithan: Actually, it's three different tables: schema_migrations, schema_migrations_core_cards and schema_migrations_deck_cards. I thought about adding models for our extra tables but Rails doesn't seem to support this in any way. The `reset_table_name` is the much simpler approach.

card/lib/card/env.rb
```diff
@@ -5,40 +5,40 @@ module Env
     class << self
       def reset args={}
         @@env = { :main_name => nil }
-        
+
         if c = args[:controller]
           self[:controller] = c
           self[:session]    = c.request.session
-          self[:params]     = c.request.params
+          self[:params]     = c.params
```
xithan: I don't know what the difference is between  `request.params` and `params` but changes to `request.params` don't affect `params` anymore. In card_controller we use `params` so we have to use here `params` too.

card/lib/card/log.rb
```diff
@@ -194,7 +196,7 @@ def print_log
 
       def text_log
         @@log.each do |entry|
-          Rails.logger.wagn entry.to_s! if entry.valid
+          Rails.logger.send @log_level, entry.to_s! if entry.valid
```
xithan: `Logger` has changed and I couldn't find a easy way to keep the custom wagn log level. Instead of that there is now a log_level option for the performance logger.

wagn/features/signup.feature
```diff
@@ -19,20 +19,21 @@ Feature: Signing up
...
+
     When I open the email
     And I click the first link in the email
     Then I should see "Wanna B"
-    
-    When I follow "Sign out"
+
+    When I go to the home page
+    And I follow "Sign out"
     Then I should not see "Wanna B"
```
xithan: I'm wondering how this passed before (or why it started to fail with Rails 4 respectively). The link in the email leads to the "Wanna B" user card. When you click on "Sign out" there is still "Wanna B" in the header of the user card so the "I should not see 'Wanna B' " test fails.

ethn:
have this open to study on the plane!

Would be nice to get view caching working before deploying this to wikirate dev, and I'd rather not merge this into master until I've studied the PR a bit more, but if it's helpful to test this on dev by deploying a rails4 branch, go for it!

wagn/rails/engine-routes.rb
ethn: I would have thought line 45 covers the next three.

card/mod/01_core/set/all/content.rb
```diff
@@ -35,6 +34,10 @@ def clean_html?
   true
 end
 
+def history?
+  false
+end
```
ethn: What does this mean?  (Seems like it means that we never store histories, but that can't be right, can it?)

card/mod/01_core/set/all/phases.rb
ethn: why can't we just do :after=>:store ?

card/mod/01_core/set/all/tracked_attributes.rb
```diff
@@ -8,7 +8,9 @@ def assign_attributes args={}, options={}
     @subcards = extract_subcard_args! args
     reset_patterns
   end
-  super args, options
+  params = ActionController::Parameters.new(args)
+  params.permit!
+  super params
```
ethn: we should probably put some comments here; I suspect some rails folks might challenge us on this decision if they dig deeply enough.

I suspect it's the right way to go for now...

card/mod/05_standard/set/all/attach.rb
```diff
@@ -111,13 +113,25 @@ def self.included(base)
...
+# allow empty files
+# without this monkey patch empty files get detected as application/octet-stream
+# and throw a type mismatch exception
+# alternative solution: turn spoof detection always off with option :validate_media_type=>false
+class Paperclip::MediaTypeSpoofDetector
+  alias_method 'content_spoofed?', 'spoofed?'
+  def spoofed?
+    @file.size == 0 ? false : content_spoofed?
+  end
+end
```
ethn: this seems sensible.  (alias_method_chain might be slightly more idiomatic?)

card/mod/05_standard/set/all/event_viz.rb
ethn: It's basically a developer tool.  The idea here is to be able to do mycard.events and see output of all the events that a card goes through.

I have found it occasionally useful in the past and likely will again, but it doesn't have be working for this PR.

card/mod/05_standard/set/all/rich_html/form.rb
ethn: Well, template was/is a wart.  It just called this:

````
    def template
      @template ||= begin
        c = controller
        t = ActionView::Base.new c.class.view_paths, {:_routes=>c._routes}, c
        t.extend c.class._helpers
        t
      end
    end
````

I think we had to do this to get access to all the standard rails form API.

We should probably come up with a cleaner idea of how we want people to create views.  We've got a lot of divergent models now (rails helpers, haml, do-it-yourself...)

File: card/mod/05_standard/set/self/signin.rb [hunk 18]
```diff
@@ -14,34 +14,34 @@ def default_title_args args
...
-  
+
   def default_core_args args={}
     args[:buttons] = button_tag 'Sign in', :situation=>'primary'
     if Card.new(:type_id=>Card::SignupID).ok? :create
       args[:buttons] += link_to( '...or sign up!', card_path("account/signup"))
     end
-    args[:buttons] += raw("<div style='float:right'>#{ view_link 'RESET PASSWORD', :edit }</div>") #FIXME - hardcoded styling
+    args[:buttons] += raw("<div style='float:right'>#{ view_link 'RESET PASSWORD', :edit, :path_opts=>{:slot=>{:hide=>:edit_toolbar}} }</div>") #FIXME - hardcoded styling
     args
```
ethn: I would think the toolbar would only come with the views with frames.  We have to start sticking this on other views now?

wagn/features/signup.feature
ethn: good question.  I might have a look to make sure the email didn't change

wagn/rails/engine-routes.rb
```diff
@@ -1,49 +1,52 @@
...
-  match 'recent(.:format)'      => 'card#read', :via=>:get, :id => ':recent' #obviate by making links use codename
-#  match ':view:(:id(.:format))'          => 'card#read', :via=>:get  
-  match '(/wagn)/:id(.:format)' => 'card#read', :via=>:get  #/wagn is deprecated
-  
+  root 'card#read'
+  get "#{ Decko::Engine.config.files_web_path }/:id(-:size)-:rev_id.:format" =>
+                                 'card#read', :id => /[^-]+/, :explicit_file=>true
+  get "assets/*filename"      => 'card#asset'
+  get "javascripts/*filename" => 'card#asset'
+  get "jasmine/*filename"     => 'card#asset'
```
ethn: side note: I bet we could express these 3 as one route

wagn/rails/engine-routes.rb
xithan: no, line 45 covers only get requests

card/mod/01_core/set/all/phases.rb
xithan: There are a lot of :after=>:store events that have to be completed before that one and :before=>:extend is too late.

card/mod/01_core/set/all/content.rb
xithan: It means there is no history without the history mod. I override this with `true` in the history mod for all cards.

card/mod/05_standard/set/self/signin.rb
xithan: > I would think the toolbar would only come with the views with frames
That's correct.
> We have to start sticking this on other views now? 
No. The edit view is framed but it doesn't make sense to show the toolbar for the signin card.

card/db/schema.rb
```diff
@@ -9,120 +9,120 @@
...
-    t.integer  "account_id",                                                :null => false
+    t.string   "password_reset_code", limit: 40
+    t.string   "status",              limit: 255, default: "request"
+    t.integer  "invite_sender_id",    limit: 4
+    t.string   "identity_url",        limit: 255
+    t.integer  "card_id",             limit: 4,                       null: false
+    t.integer  "account_id",          limit: 4,                       null: false
   end
 
 end
```
ethn: what's the rationale behind all the explicit limits?

card/mod/01_core/set/all/phases.rb
ethn: I kind of want to bang on this a bit more, but I get the rationale.

GerryG: Hmm, in other words, we can't say "last thing after => store", and there isn't a sequential next step that you could do 'before => last_store_event'  It wouldn't be too bad to add such an event (better named).

PR #435, opened 2015-06-17
ActiveJob events

xithan:
use Rails' ActiveJob to execute events after the request
The current syntax is  `event :my_event :after/:before=>:subsequent`


card/card.gemspec
```diff
@@ -34,8 +34,9 @@ Gem::Specification.new do |s|
 
     [ 'smartname',    '0.2.3'   ],
 
+    [ 'activerecord-session_store'],
     [ 'uuid',         '~> 2.3'  ],
-    [ 'paperclip',    '~> 2.8'  ],
+    [ 'paperclip',    '~> 4.2'  ],
```
GerryG: Is this so paperclip can use ActiveJob too?

card/db/schema.rb
```diff
@@ -9,120 +9,120 @@
...
-  create_table "card_revisions", :force => true do |t|
-    t.datetime "created_at", :null => false
-    t.integer  "card_id",    :null => false
-    t.integer  "creator_id", :null => false
-    t.text     "content",    :null => false
+  create_table "card_revisions", force: :cascade do |t|
+    t.datetime "created_at",               null: false
+    t.integer  "card_id",    limit: 4,     null: false
+    t.integer  "creator_id", limit: 4,     null: false
+    t.text     "content",    limit: 65535, null: false
```
GerryG: I'm guessing most of the new limit: values are because of a gem version update, but I do wonder about this one.  Maybe postgres or mysql text is so limited, but I think we want to be careful with any explicit limit on content size.

card/card.gemspec
xithan: no, that's there because ActiveJob needs Rails 4 and Rails 4 needs that paperclip update.

card/db/schema.rb
xithan: 65535 is mysql's limit for text so the content size didn't change but maybe we should increase that to mediumtext (16MB). While working on bootstrap I encountered css cards that were too big for the cards table.

GerryG: Yeah, I don't like arbitrary limits.  We'll always be using some form of large object type in the database for this, so I don't think there is any value in a limit smaller that what the DB can handle.

Of course the tricky part is generalizing this so it works in all DB choices we care about.

card/card.gemspec
GerryG: Ah, didn't realize this is downstream of the Rails 4 update.  I guess this was missed in that branch, or more likely it just isn't merge yet.

PR #437, opened 2015-06-19
remove non related source in groups

chuenlok:
In coverage report, it always shows non mod code in mod groupings which may confused the readers.

I am not sure if this is a ok approach to check if the file is related to the mod.

### DIFF
--- a/card/lib/card/simplecov_helper.rb
+++ b/card/lib/card/simplecov_helper.rb
@@ -1,3 +1,6 @@
+def is_in_mod? file,mod_path
+  file.src.join("") =~ /below pulled from #{Rails.root}\/mod/#{mod_path}\//
+end
 def card_simplecov_filters
   add_filter 'spec/'
   add_filter '/config/'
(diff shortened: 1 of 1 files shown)


card/lib/card/simplecov_helper.rb
```diff
@@ -1,3 +1,6 @@
+def is_in_mod? file,mod_path
+  file.src.join("") =~ /below pulled from #{Rails.root}\/mod/#{mod_path}\//
+end
```
ethn: We can do this for now, but it might make more sense to put a cleaner flag in those files at the time that they're generated.

card/lib/card/simplecov_helper.rb
```diff
@@ -15,7 +18,8 @@ def card_simplecov_filters
       src_file.filename =~ /mod\/#{mod}\// or 
         (
           src_file.filename =~ /tmp\// and
-          /\d+-(.+\.rb)/.match(src_file.filename) { |m| Dir["mod/#{mod}/**/#{m[1].gsub("-","/")}"].present? } 
+          /\d+-(.+\.rb)/.match(src_file.filename) { |m| Dir["mod/#{mod}/**/#{m[1].gsub("-","/")}"].present? } and
+          is_in_mod? src_file,mod
```
ethn: should probably separated this matching out into a separate method as well (and give it a name so that we know what it's doing)

card/lib/card/simplecov_helper.rb
GerryG: What is file.src ?  Is file other than some standard File or related class?  It certainly isn't very clear what this is about.

PR #438, opened 2015-06-20
turn ViewCache on again

xithan:


ethn:
Feel free to deploy to dev when you think it will work there.

PR #440, opened 2015-06-27
new modal layout

ethn:
modal is no longer a view but a layout.

### DIFF
--- a/card/mod/05_email/set/all/follow.rb
+++ b/card/mod/05_email/set/all/follow.rb
@@ -51,27 +51,9 @@ end
 format :html do
 
 
-=begin
+
+
   view :follow_link, :tags=>:unknown_ok, :perms=>:none do |args|
-    hash = follow_link_hash args
-    text = %[<span class="follow-verb">#{hash[:verb]}</span> #{args[:label]}]
-    opts = {
-      :title           => hash[:title],
-      :class           => "follow-toggle #{hash[:class]}",
-      'data-follow'    => JSON(hash),
-      'data-rule_name' => card.default_follow_set_card.follow_rule_name( Auth.current.name ).to_name.url_key,
-      'data-card_key'  => card.key
-    }
-    link_to text, '', opts
-  end
-
-  def default_follow_link_args args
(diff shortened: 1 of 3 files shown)


card/lib/card/format.rb
```diff
@@ -392,9 +392,12 @@ def prepare_nest opts
...
       case
-      when opts.has_key?( :comment )                            ; opts[:comment]   # as in commented code
-      when @mode == :closed && @char_count > Card.config.max_char_count   ; ''     # already out of view
-      when opts[:inc_name]=='_main' && !Env.ajax? && @depth==0  ; expand_main opts
+      when opts.has_key?( :comment )                                      ; opts[:comment]
+        # commented nest
+      when @mode == :closed && @char_count > Card.config.max_char_count   ; ''
+        # already out of view
+      when opts[:inc_name]=='_main' && show_layout? && @depth==0          ; expand_main opts
+        # why do we have these restrictions on main?
```
GerryG: The sense of it seems to be only doing this in a layout rule card.  This is the only place the _main is handled, right?.

Maybe it is for viewing layout cards, and not looping the inclusion?

ethn: this is about when we expand _main *as the main card*.  there are other times we process the inclusion, and there are times within layout cards that we don't process _main.

GerryG: Yes, but the comment asks why we need to test more than just _main, and I think it is because of what you write.  We only want this particular processing when it is the main card, always depth 0 from a layout card.
And I guess you are saying there is another place where inclusion of _main is suppressed (when the layout is the main card, for example).  Subbing in the _main content is a different step when processing the nest at the next level.

ethn: right.  One question is whether to process the nest, the next is whether it's the one slot on the page that gets rendered with the <div id="main"> wrapper.  (It may actually be possible to have two for now, technically, but at least this makes that less likely)

GerryG: In other words, to have two {{_main|...}} in a layout.  That could be handled if we have a spec for what we wanted.  Maybe two are ok, maybe one doesn't get the wrapper.

PR #441, opened 2015-06-28
Fix type changing problem

chuenlok:
I think adding a checking on the event is better than overriding a dummy on every other type.


card/mod/02_basic_types/set/type/pointer.rb
```diff
@@ -211,8 +211,11 @@ def wrap_item item, args
   end
 end
 
-
-event :standardize_items, :before=>:approve, :on=>:save do
+# while a card's card type and content are updated in the same request, 
+# the new module will override the old module's events and functions.
+# this event is only on pointer card. Other type cards do not have this event,
+# so it is not overridden and will be run while updating type and content in the same request. 
+event :standardize_items, :before=>:approve, :on=>:save, :when=>proc{  |c| c.type_id == Card::PointerID  } do
```
GerryG: This is a great catch, although this may not be the best solution.  Ethan probably had a handle on the right way, probably you just have to throw away any in-memory versions of the card object (local cache and other dynamic references active during the life of the request).  If the card is no longer a pointer, it shouldn't be calling this.  This works for a specific workaround, but there are other type specific event handlers.

ethn: Any chance we could reproduce this bug in a test?

chuenlok: Let me try to make a test case for that.

GerryG: Yeah, that will be pretty important in making sure the better way to fix it actually works.

PR #442, opened 2015-06-30
Docker

GerryG:
In the long run, I think this is a way better approach to how to deploy on cloud infrastructure.


docker/db/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM jpetazzo/pgsql
```
GerryG: There are many postgres images.  If there is some 'official' ones from the project maintainers ...  This may be even be one of those.

docker/db/init
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+# customize postgres startup here
+# 
+# Synopsis: /init secret [username]
+#  or starting current state: /init
```
GerryG: Based on the init script in the FROM image.  That one only took the 'secret' for the 'root' user.  We add the ability to specify the user name as a second argument.

docker/db/init
```diff
@@ -0,0 +1,30 @@
...
+
+# start the server (background it)
+su postgres sh -c "/usr/lib/postgresql/9.1/bin/postgres -D /var/lib/postgresql/9.1/main  -c config_file=/etc/postgresql/9.1/main/postgresql.conf -c listen_addresses=*" &
+
+sleep 10  # make sure it is all the way up
+
+# run on db create and remove
+UUE_SCRIPT='/tmp/script/template_uuen.sql'
+if [ -n "${SECRET}" -a -f ${UUE_SCRIPT} ]; then
+  # create rails user
```
GerryG: We also create a database with the same name as the user.  It seems like logins can fail unless they specify the database to connect too and this database doesn't exist.

The UUE_SCRIPT is from some Q&A site, but I can't find the link right now.

docker/files/webapp.conf
```diff
@@ -0,0 +1,8 @@
+server {
+  listen 80;
+  server_name wagn;
```
GerryG: We'll need to use an environment variable, or something like that so we can make this universal.

docker/files/webapp.conf
```diff
@@ -0,0 +1,8 @@
+server {
```
GerryG: wagnapp.tgz is pretty much just the result of running: wagn new wagnapp.

Note that you can't do: wagn new wagn (re-uses the Wagn constant)
The error is not right, it tries to throw an exception, but gets an undefined constant on the exception class.  That might be a leftover from the gem split.

GerryG:
Automatically builds the image on docker hub now:
https://registry.hub.docker.com/u/gerry/wagn/build_id/71780/code/bcnkqzshdnakyun3cswhkcj/

ethn:
This is just a static image for deployment, right?  Doesn't handle updates?  If so, not sure I could agree it's a "way better approach" then Cloudstore, which handles updates very well.

GerryG:
Actually, no.  Docker hub has a build server and git integration like semaphore.  I think you could run several on different branches and update different tags.  At least I hope you can do that with the free services.

What's better about it is the 'one function per container', which I think is the way to go from a scalable deployment perspective.  There is still a role for the cloud store service in providing a UI for managing the  configuration.  The function you'd still need to even think about replacing cldstr would be the management of the containers on a docker host.  The mapping of IP/DNS configuration to containers.  To do something like name based virtual hosts would need to be handled by a proxy configuration on the docker host.  It may be easier to just allocate IP addresses to a containers.  Note that it also removes the constraint of using mysql that is imposed by cldstr.  The database is in another container and not our concern.  If you go with AWS provisioning, they have better ways handling data anyway.  We are about to start deploying internal applications on AWS using Elastic Beanstalk.

Cldstr is right for our use case of deploying many wagns on a single virtual host, or repeating that pattern some small number of times.  If I am deploying a wagn to scale, I don't need what cldstr does, I need the enterprise class services that AWS and the other cloud service providers offer natively.

Just to be clean, I'm not proposing that we even think about doing that now.  That's a sunk investment of time that is working fine right now.  Docker based solutions could work for that and I would encourage JE to change to that approach at the appropriate time.

I provide this as an option for us and others to use.  No reason you couldn't use it for wikirate.  I like the idea of doing the integration of wagn/decko plugins in the building of the image rather than having cldstr manage them as plugins.  Not that they require that say wikirake = wagn + mods1 + mods2 vs. wikirate is at the same level as wagn with no obvious relationship.  It is also a great option for future applications where decks are providing micro-services to a more complex application.

There is no cost or risk to merging this branch.  It won't effect anyone who doesn't use docker.

GerryG:
The URI above is a specific build, but this will track commits to the branch.  When you merge this, I will change the branch to master ...

GerryG:
Ok, merged master and pushed.  It should automatically build.
I wonder if semaphore can run the docker images?

GerryG:
Yes, it updates.  Created this one when I just merged master again: https://registry.hub.docker.com/u/gerry/wagn/build_id/71780/code/barke8jghf9qcxcegvbqtdb/

GerryG:
Re: Slowness in getting back to stuff.  No worries, there are no dependencies on this now.  We should probably invite some people to give this a try before we merge this.  No reason to merge it before there is more demand for it.  On the other hand, it won't hurt to merge it when you do get a chance to check it out a bit more.

docker/db/init
GerryG: Actually, the db image isn't needed anymore, but I have to change the test setup for that.  Now it can use the 'official' postgres images from the public registries.  There already is a way to tell it what user to add, and the default setup is UTF8 encoding in the right template.

docker/files/webapp.conf
GerryG: I figured this out too.  You can pass --env MY_ENV_VAR=my_value in the docker run commands.  Then you just use the server's way of referencing the environment.  So wagh should be replaced by ${SERVER_NAME} (or whatever is correct syntax for nginx config files).

GerryG:
I now have docker capability on my semaphore account, but I need to work on the images.  I think the main issue is that I generated an image to run wagn, not to test it.

Also, wanted to ask whether we need images for both postgres and mysql.  I think we prefer postgres and only switched to mysql to use cldstr.  We can handle this by creating several images and using docker's tags to distinguish which one.  I think I will just create a 'test' label for the semaphore builds, and if you want, I can also create a mysql and mysql_test.

GerryG:
BTW, docker beta is enabled on your semaphore account too.  Not sure if they did that just now or if you asked for it yourself.  I accidentally changed the platform on your wagn project, but I noticed what I had done and switched it back.

ethn:
I'll merge this in.

I don't think it makes much sense to pit Cldstr vs Docker; they're pretty different in terms of audience / value add.

A bit part of what makes Cldstr the better fit for us has been that it's not just geared towards developers; I can send anyone there.  It does actually handle roles (like the one function per container), but that's not really where they're focusing.  You can see by its new branding at http://ubos.net that they probably won't have much interest in rebuilding on Docker.  (And, fwiw, our servers are already on AWS).

I do like the idea of supporting Docker, though.

By updates, I was talking about managing migrations and such (`wagn update`, basically).  I suspect that's possible, too.

So is your new semaphore build just checking that the docker build works, or is it running all the wagn tests?

I would suspect there's no need for both mysql and postgres until we've got some idea of how we want to promote the docker service in the first place (and get some demand for the other one).  If it's running all the tests in semaphore, I suppose we could run them in postgres on the docker account and mysql on the standard build...

GerryG:
Right, not the same function, but I think the better design solution for cldstr would use docker.  The part of the problem that isn't that well solved (automation wise) is managing domain names through to virtual hosts, virtual network configuration and all that.  But that's all the plumbing, the UX of click to add your website is great.  Correct me if I'm wrong, but GC sites really don't have any connection the the UX, right?  No reason to interfere with what is smoothly working, but I think you get the point.  I could implement a card based interface to some configurations scripts for the dockerhost, and we would have something of much higher value in usefulness that the current setup.  Not something to be done overnight, but not a huge project either.

GerryG:
BTW, still working on the Docker based CI config on Semaphore.  It's in beta for Semaphore.  I think they enabled it on your account too.

GerryG:
Ok, I'm going to check out UBOS.  I do agree about the different audiences.  On the other hand, there is no reason that you couldn't use UBOS as the base docker image.  The data isn't in the app container whatever your architecture.  Managing data is one of the big maintenance issues no matter how you slice it, and whatever facilities you use for your data, it's got to be portable.  The fascility needs to export/import in standard (or per DB standard) formats.  You don't want to be stuck in a cloud or hosting provider based on any of your core technologies.

PR #443, opened 2015-07-02
united toolbars, improved rule navigation

xithan:


xithan:
No idea what is causing the trouble on semaphore. The failing cukes passed on my computer and I used the same seed data.

card/lib/card/name.rb
```diff
@@ -41,5 +41,17 @@ def trait tag_code
     def code
       Card::Codename[ Card.fetch_id self ]
     end
+
+
+    def is_a_field_of? context_name
```
GerryG: Ok, I'm not totally sure I get the logic of this.  And, if it works as the name implies, I think we should push it to the general name layer (Smarname for now).  We'll still keep it in a gem, but combine that gem in this repo as I understood it.  Well, never mind about the location, the logic ...

If the to_absolute does any expanding of _self, or _left, etc, or the standard +X expansion, then it can't match the relative name.  So, the magic in the line 49 must be producing a +X form from any prefix of the context card.  In effect we are testing independence from the context card.

I'm wondering whether we need to do the absolute name, won't the relative_name always have a leading + when then don't match, and always not when they do?

card/mod/01_core/set/all/collection.rb
```diff
@@ -177,15 +177,57 @@ def set_search_params_variables! hash
...
-    Card::Content.new(card.content, card).find_chunks( Card::Chunk::Reference ).each do |chunk|
+  def each_reference_with_args args={}
+    Card::Content.new(_render_raw(args), card).find_chunks( Card::Chunk::Reference ).each do |chunk|
       yield(chunk.referee_name.to_s, nest_args(args,chunk))
     end
   end
 
-  def map_nests args={}, &block
+
+  def each_nested_chunk args={}
```
GerryG: Oh, I didn't realize we had code treating the referencees as a collection.  That's cool, and this looks like a big upgrade to that initial step.

card/mod/01_core/set/all/collection.rb
```diff
@@ -177,15 +177,57 @@ def set_search_params_variables! hash
...
+
+  def nested_fields args={}
+    result = []
+    each_nested_field(args) do |chunk|
+      result << chunk
+    end
+    result
+  end
+
+  def unique_chunks chunk, processed_set, &block
```
GerryG: Ah, so this is where it handles making the fields unique.  Very nice!

card/mod/05_standard/set/all/links.rb
```diff
@@ -14,6 +14,25 @@ def link_to text, href, opts={}
     end
   end
 
+  # link to url, view, card or related card
+  #
+  def smart_link link_text, target, html_args={}
```
GerryG: Ok, looks cool, but what is a smart_link ?

card/lib/card/name.rb
xithan: Good point. Checking whether the relative name starts with a '+'  seems like the right solution for all cases.

( I had to add the additional case in line 53 is for structured cardtypes. The page /new/mycardtype has an empty name so the absolute name of a field is the same as the relative name. Your suggestion would handle that)

card/mod/05_standard/set/all/links.rb
xithan: Probably not the best name choice. I needed a method that can create different link types (like card_link or view_link) depending on the argument so this method is 'smart' about creating the right link type. Maybe 'wagn_link' instead?

PR #444, opened 2015-07-02
rails 4 (wikirate)

xithan:


GerryG:
Ok, so if this is merged, then the version change for paperclip probably needs to be here too.  Unless you are merging the active job branch right away.

PR #450, opened 2015-07-06
fix rss format with a search type sturcture with found_by

chuenlok:
https://wagn.airbrake.io/projects/95878/groups/1421123009760488992


ethn:
woot!

PR #455, opened 2015-07-13
fix bug: cardio config overrides deck config

xithan:
This problem occurs in Rails 4 but only in production mode. 
In production mode card/lib/cardio.rb#set_config gets called twice, in development only one time.
The second call overrides the deck configuration. I don't know why this is happening. This PR cures only the symptoms.


GerryG:
Should this be targeted at a base rails4 branch?  Also, were you able to figure out why it got called twice, or did you have to do more of a workaround?

PR #457, opened 2015-07-15
customizable modal footer

xithan:
use modal_footer view for the footer in a modal window


ethn:
We decided to go a different direction, no?  Shall I close?

PR #461, opened 2015-07-24
fix broken renaming

xithan:


ethn:
any idea why this breaks?

PR #463, opened 2015-07-28
Wikirate

chuenlok:


card/mod/03_machines/lib/javascript/wagn.js.coffee
```diff
@@ -169,6 +169,7 @@ $(window).ready ->
   $('body').on 'loaded.bs.modal', null, (event) ->
     unless event.slotSuccessful
       wagn.initializeEditors $(event.target)
+      $(event.target).find(".card-slot").trigger("slotReady")
       event.slotSuccessful = true
```
ethn: these seem like a random assortment of todos.  Have we looked to see if there are more things that need to be called?  Should they be grouped and named accordingly?

PR #464, opened 2015-07-28
replace paperclip gem with carrierwave

xithan:


card/db/migrate_core_cards/20150724123438_update_file_and_image_cards.rb
```diff
@@ -0,0 +1,26 @@
...
+      if card.content.present?
+        attach_array = card.content.split "\n"
+        attach_array[0].match(/\.(.+)$/) do |match|
+          extension = $1
+          if attach_array.size > 3  # mod file
+            codecard = card.cardname.junction? ? card.left : card
+            card.update_attributes :content=>":#{codecard.codename}/#{attach_array[3]}.#{extension}"
+          else
+            card.update_attributes :content=>"~#{card.id}/#{card.last_action_id}.#{extension}"
+          end
```
GerryG: I gather this changes the content from paperclip virtual attributes to what carrierwave uses.  Nice that it is simply a path of some form.

card/mod/04_settings/set/type/setting.rb
```diff
@@ -1,5 +1,13 @@
 require_dependency 'json'
-  
+
+def self.member_names
+  @@member_names ||= begin
+    Card.search( :type_id=>SettingID, :return=>'key' ).inject({}) do |hash, card_key|
```
GerryG: I used to use inject like this until a colleague kept pointing out that each_with_object is better in this case (assuming we have it in ruby 1.9.3).  It only works when your block in changing the object via side-effects, which you are in this case.

ethn: nice.

PR #465, opened 2015-07-29
make regex ruby 1.9.3 compatible

xithan:


xithan:
I close this and open it again to run CI with ruby 1.9.3
I can't find a way to rebuild something on semaphore that has passed

PR #466, opened 2015-07-29
make regex ruby 1.9.3 compatible

xithan:
ruby 1.9.3 doesn't support the word boundary anchor \b


GerryG:
I thought rails 4.0 forces ruby 2+ ?

xithan:
Rails 4 recommends ruby 2+ but cloudstore uses ruby 1.9.3. I think that forces us to support 1.9.3

ethn:
right.  rails 4 requires 1.9.3 but recommends 2.2.  So now the same goes for Wagn :)

See http://rubyonrails.org/download/

PR #467, opened 2015-07-29
add single autocomplete input for pointer

chuenlok:
Any better name for the new option?


card/db/migrate_core_cards/20150729143608_import_new_pointer_option.rb
```diff
@@ -0,0 +1,9 @@
+# -*- encoding : utf-8 -*-
+
+class ImportNewPointerOption < Card::CoreMigration
+  def up
+    options_card = Card.fetch "*input+*right+*options"
+    options_card<<"single_autocomplete"
```
xithan: I guess what you use here as name will show up in the editor as option so if "single autocomplete" is the right name for this then better without underscore.

card/mod/02_basic_types/set/type/pointer.rb
```diff
@@ -57,6 +57,25 @@ def wrap_item item, args={}
...
+      <ul class="pointer-list-editor #{extra_css_class}" data-options-card="#{options_card_name}">
+        <li class="pointer-li">
+        <span class="input-group">
+          #{ text_field_tag 'pointer_item', items[0], :class=>'pointer-item-text form-control' }
+        </span>
+        </li>
+      </ul>
+    }
+  end
+
```
xithan: I see two possibilities to reduce code repetition:
1) move all the logic of line 61-66 to `default_list_args` and store it in args and then call `default_list_args` in `default_single_autocomplete_args`
or 2) make the drag&drop handle, the remove and the add button optional in `list_item` and `list` view and turn it off in in  `default_single_autocomplete_args`. Then you just call `list` view in `single_autocomplete` view

chuenlok: I will update the code based on your suggestion :)

PR #468, opened 2015-07-30
fix multiple editor on layout

chuenlok:
My theory is that, `layout_type.rb` is in 05_standard which gets the editor from `include Html::Format`. `html.rb` was in 05_standard as well but we moved it to 02_basic_types some days ago and therefore the editor is overridden in `rich_html/form.rb` in 05_standard. 

If this is the case, I just need to redefine the editor of ace editor for layout_type again and also remove the extra `data-card-type-code` property to trigger the ace editor as tinymce is defined in class already.

### DIFF
--- a/card/mod/05_standard/set/all/rich_html/form.rb
+++ b/card/mod/05_standard/set/all/rich_html/form.rb
@@ -182,7 +182,7 @@ format :html do
 # FIELD VIEWS
 
   view :editor do |args|
-    text_area :content, :rows=>3, :class=>'tinymce-textarea card-content', :id=>unique_id, "data-card-type-code"=>card.type_code
+    text_area :content, :rows=>3, :class=>'tinymce-textarea card-content'
   end
 
   view :edit_in_form, :perms=>:update, :tags=>:unknown_ok do |args|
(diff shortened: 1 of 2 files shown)


card/mod/05_standard/set/all/rich_html/form.rb
```diff
@@ -182,7 +182,7 @@ def content_field form, options={}
 # FIELD VIEWS
 
   view :editor do |args|
-    text_area :content, :rows=>3, :class=>'tinymce-textarea card-content', :id=>unique_id, "data-card-type-code"=>card.type_code
+    text_area :content, :rows=>3, :class=>'tinymce-textarea card-content', :id=>unique_id
   end
```
ethn: why would we remove this?  Isn't it needed somewhere?

ethn: nevermind; we never want tinymce and ace at the same time!

would be nice if the ace editor stuff only showed up if BOTH the ace-editor class AND the type-code were there.

chuenlok: will do it later.

PR #471, opened 2015-08-04
introducing abstract sets using permissions

ethn:
Basic idea is to make it so that you can create set API definitions (events, views, etc) abstractly and then use them in specific sets.

First use case is permissions (where it made little sense to have, eg, *create inheriting from *comment, but since comment was the first permission loaded, that's exactly what we were doing)


card/mod/04_settings/set/abstract/permission.rb
```diff
@@ -0,0 +1,99 @@
...
+end
+
+def testoo
+  'bestoo'
+end
+
+format :html do
+
+  view :pointer_core do |args| #, :view=>:core, :mod=>Type::Pointer::HtmlFormat
+    %{<div class="pointer-list">#{ render_pointer_items args }</div>}
```
ethn: the PR obscures this change, but this is the only real change since this code was in comment.rb.  It breaks if you use the view aliasing syntax, since abstract sets are loaded before anything else.  We may need to come up with a way to postpone aliasing until after everything is loaded so mods aren't so sensitive to load order.  ...or a way to express dependencies ?

card/mod/04_settings/set/right/create.rb
```diff
@@ -1,3 +1,3 @@
-include Comment
+include Abstract::Permission
```
ethn: Do we like "Abstract" more than "Lib"?

ethn: ...might actually be less confusing, since there's already a <modname>/lib dir....

xithan:
next step: a shortcut to include all formats of a given set

xithan:
events don't work in abstract sets. There is a `create_card_changes` event in an *all set.  When I  define an event with `:before=>:create_card_changes` in an abstract sets I get `undefined method '_create_card_changes_callbacks'`

xithan:
realized that abstract events can't work anyway as long as we define them globally. 
If I include the same event in `type/file` and `type/image` the latter will override the former.

xithan:
I found a solution for abstract events. I can do the following in an abstract set:

```ruby
def self.included host_class
  set_name = host_class.name.underscore.gsub('/','_')
  host_class.class_eval <<-RUBY, __FILE__, __LINE__+1 
    event :write_identifier_#{set_name}, :before=>:store do
      ...
    end
  RUBY
end
```

PR #472, opened 2015-08-05
ensure_set method

xithan:
creates set constant if it doesn't exist

### DIFF
--- a/card/lib/card/set.rb
+++ b/card/lib/card/set.rb
@@ -236,6 +236,22 @@ class Card
     end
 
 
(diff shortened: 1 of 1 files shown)


GerryG:
I sort of get what this might be for, but where's the code that uses it?  I suppose it could be in wiki-rate or other extension mods.

ethn:
yeah, it's in wikirate.  we haven't decided whether we love this solution yet.  it seems pretty safe, but it requires a lot of mod developers.

PR #475, opened 2015-08-07
search reference tracking

xithan:


card/mod/01_core/chunk/query_reference.rb
```diff
@@ -0,0 +1,65 @@
...
+  # 1) {"name":"Alfred"}
+  # 2a) {"name":["in","Alfred"]}
+  # 3a) {"plus_right":["Alfred"]}
+  # but not in
+  # 2b) "content":"foo", "Alfred":"bar"
+  # 3b) {"name":["Alfred", "Toni"]}
+  # It's not possible to distinguish between 2a) and 2b) or 3a) and 3b) with a simple regex
+  # hence we use a too general regex and check for query keywords after the match
+  # which of course means that we don't find references with query keywords as name
+  class QueryReference < Reference
```
GerryG: I am pretty impressed you were able to get the chunk code to process queries for name references.  Seems like it might be better to scan the json as parsed objects, but maybe it is better to just stay that the string level for all types.

ethn: I suspect we may eventually want to move the JSON parsing into chunk code, especially once we have caching for parsed content.

PR #482, opened 2015-08-12
reorder the priority of js format in pointer

chuenlok:
Example: http://dev.wikirate.org/sebastian_jekutsch+csr_report_available+Acer_Inc_+2011+source?view=edit&slot[items][structure]=source_item&item=content&debug=script

```
nest item, :view=>(params[:item] || args[:item] || :core)
```
from `pointer.rb`
If `params[:item]` is in front of `args[:item]`, 
```
@js_tag = if params[:debug] == 'script'
      script_card.format(:js).render_core :item => :include_tag
    elsif script_card
      javascript_include_tag script_card.machine_output_url
    end
```
from `head.rb`
will render the content view for the items instead of view `:include_tag`.

### DIFF
--- a/card/mod/02_basic_types/set/type/pointer.rb
+++ b/card/mod/02_basic_types/set/type/pointer.rb
@@ -197,7 +197,7 @@ end
 format :js do
   view :core do |args|
     card.item_cards.map do |item|
-      nest item, :view=>(params[:item] || args[:item] || :core)
+      nest item, :view=>( args[:item] || params[:item] || :core)
     end.join "\n\n"
   end
 end


ethn:
in that example, why is item=content in the url?

The current order seems right to me.  wagneering > coding > default  (although I like the idea that sometimes coders can choose to insist, as in the hide/show functionality)

chuenlok:
We need to show the content view of the item so it is in the url.
Otherwise, it will be like 
image
So is there a better way to handle this?

ethn:
hmm, I'm realizing that it's weird that params[:item] changes the item view of *every* collection on the page.  Seems like it should only affect the main card, no?   And, now that I look at it, it seems like that might actually work if we just remove this params[:item].  Does it work on searches and pointers if we do that?

ethn:
Let me know when you've tried removing params[:item] from this line altogether.  If item views of searches and pointers still work in the main card as expected, I think we should get rid of this.

GerryG:
The default view for included content and items of Search/Pointer is 'content', right?

ethn:
The default view for Search/Pointer items is :closed, at least in HtmlFormat.  (Different in different formats).

The default view for other nests has *traditionally* been "content", and this is supported through a semi-secret rule, eg http://wagn.org/*all+*default_html_view.  In new decks, this is set to titled.

There have been proposals to make *all* nests (including items) default to titled view for uniformity.

PR #483, opened 2015-08-12
handle urls as source for file and image cards

xithan:


ethn:
appears to need merge of latest from master.

card/mod/05_standard/set/type/file.rb
```diff
@@ -1,6 +1,6 @@
 include Abstract::Attachment
 
-set_specific_attributes :file
```
ethn: didn't take long to validate the effort we put into this :)

PR #484, opened 2015-08-12
fetch can now return trashed cards

ethn:
we need action.card to return a card (with set modules), even if it's in the trash.  In the long run, we probably want a lot of other trash handling, too.

What I've implemented so far is pretty unobtrusive; it really only changes things if you're fetching with :look_in_trash => true.  It basically bypasses the caching system.

Might want to look at using this mechanism when we actually retrieve from trash.

### DIFF
--- a/card/mod/01_core/set/all/fetch.rb
+++ b/card/mod/01_core/set/all/fetch.rb
@@ -11,17 +11,17 @@ module ClassMethods
   #   - database
   #   - virtual cards
   #
-  # "mark" here means a generic identifier -- can be a numeric id, a name, a string name, etc.
+  # "mark" here means one of three unique identifiers
+  #    1. a numeric id (Integer)
+  #    2. a name/key (String or Card::Name)
+  #    3. a codename (Symbol)
   #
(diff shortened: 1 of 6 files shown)


GerryG:
I think I'll be really happy to see this released and helping me see the trash at gerry.wagn.org

ethn:
it will be released soon, but there won't really be any interface to it; it was done to make actions work.  baby steps :)

GerryG:
I figured that, but I'm also guessing that right now there are methods that might include deleted cards that don't now.  If this just gives the capability to list deletes in recent, that would be a big help.  That might not just happen, but probably isn't that hard.
I know, the big UI work needed is to make /recent be more like card history in accounting for Acts and Actions to group changes.  Fixing the old interface isn't worth that much.

GerryG:
I note that it is the difference between github comment boxes and wagn discussion boxes that makes me a bit nuts.  github needs a second newline to get paragraph spacing.  Wagn doesn't show the right spacing when typing the comment, so it makes you think you need extra newlines.

PR #486, opened 2015-08-14
trying to fix initializers

ethn:
we're going to need a way to (a) get all mod dirs loaded up first (including those configured in a deck's application.rb file), and then (b) do stuff to all those mods later.

I haven't really gotten the timing right, but this introduces the separation.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -12,6 +12,17 @@ task :release do
   system %(git tag -a v#{version} -m "Wagn Version #{version}";  git push --tags wagn)
 end
 
+task :cp_tmpsets do
+  system %(
+    cd ..
+    cp -r sites/core-dev/tmp/set* decko-tmpsets
+    cd decko-tmpsets
+    git commit -a -m 'updated from core-dev'
+    git push; git push wagn
+    cd ../gem
+    git submodule update --remote
+  )
+end
 
 def version
   File.open(File.expand_path( '../card/VERSION', __FILE__ )).read.chomp
--- a/card/lib/cardio.rb
+++ b/card/lib/cardio.rb
@@ -69,6 +69,7 @@ module Cardio
       add_path 'tmp/set_pattern', :root => root
 
       add_path 'mod'
+
       add_path "db"
       add_path 'db/migrate'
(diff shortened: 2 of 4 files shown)


Rakefile
```diff
@@ -12,6 +12,17 @@ task :release do
...
+  system %(
+    cd ..
+    cp -r sites/core-dev/tmp/set* decko-tmpsets
+    cd decko-tmpsets
+    git commit -a -m 'updated from core-dev'
+    git push; git push wagn
+    cd ../gem
+    git submodule update --remote
+  )
+end
```
ethn: (unrelated.  task to update the tmpsets based on my own dir arrangement)

card/lib/cardio.rb
```diff
@@ -76,8 +77,22 @@ def set_paths paths
       add_path "db/seeds", :with => "db/seeds.rb"
 
       add_path 'config/initializers',  :glob => '**/*.rb'
-      paths['config/initializers'] << "#{gem_root}/mod/**{,/*/**}/initializers"
-      paths['config/initializers'] << "#{root}/mod/**{,/*/**}/initializers"
+
+    end
+
```
ethn: this wasn't working (so airbrake wasn't getting initialized)

GerryG: I'm still confused about all the config and paths objects, and when they are loaded.
How did you get this to work then?  Did you copy stuff or some other shim?

ethn: set_mod_paths

ethn: good because it uses the mod paths; bad because it happens before decks can override said paths.

GerryG: Ok, I think I get everything except the 'override said paths' part.

So, it works for mods in the deck under mod/.../initializers/, but sometimes you want to change that?  Is this something you need to do now for deck modules and have to work-around, or?

ethn: afaik, nobody needs custom paths for mod directories now; just seems likely at some point.  The current code won't yet support that (at least for initializers).

PR #491, opened 2015-08-28
Silent changes

xithan:
option to turn off notifications when you change attributes
usage:
````
 card.update_attributes! :name=>'new_name', :silent_change=>true
````

### CODE REVIEW

File: card/lib/card.rb [hunk 1]
```diff
@@ -47,6 +47,7 @@ def self.serializable_attr_accessor *args
     :comment, :comment_author,    # obviated soon
     :update_referencers,          # wrong mechanism for this
     :update_all_users,            # if the above is wrong then this one too
+    :silent_change,               # and this probably too
```
xithan: We could use blocks instead:
```
silent do 
   without_updating_referencers do 
     --
   end
end

File: card/db/schema.rb [hunk 2]
```diff
@@ -38,7 +38,7 @@
   create_table "card_changes", force: :cascade do |t|
     t.integer "card_action_id", limit: 4
     t.integer "field",          limit: 4
-    t.text    "value",          limit: 65535
+    t.text    "value",          limit: 16777215
```
ethn: is there a benefit to explicit limits?

card/db/schema.rb
```diff
@@ -66,20 +66,20 @@
...
-    t.string   "name",                limit: 255,   null: false
-    t.string   "key",                 limit: 255,   null: false
-    t.string   "codename",            limit: 255
     t.integer  "left_id",             limit: 4
-    t.integer  "right_id",            limit: 4
-    t.integer  "current_revision_id", limit: 4
     t.datetime "created_at",                        null: false
     t.datetime "updated_at",                        null: false
+    t.integer  "current_revision_id", limit: 4
+    t.string   "name",                limit: 255,   null: false
```
ethn: fwiw, I like the old order; doubt that was an intentional change?

card/lib/card.rb
ethn: I'll merge this.  Will be good to have a grand API discussion after the Post-it Summit :)

card/db/schema.rb
xithan: no, that's auto generated

ethn: I get that, but it's still curious to me, because it's based on the column order in the db, which is, in turn, based on a previous copy of this file.  So I wonder how your db got the weird order?

xithan: no idea

card/db/schema.rb
GerryG: I'd tend towards no limit unless there is a benefit. The benefit (or lack) may depend on which db you are using, you could get different underlying types with different behavior, but generally we want the largest it will give us.

We had an extended debate and comparison of the performance of different scenarios.  This was all for Oracle, but it could be like this or worse on others.  Bottom line is the *LOB types are often stored very differently and it will take several round trips to the DB if your query returns one of these.

To summarize, you could potentially get some performance gain by having a split field where it used a type that fits, plus a second column for large content. So, you could query just the short column in general, but make a different query when you need to return big values too.

All the complexity probably isn't worth it, so just go with the biggest possible.

ethn: is it safe to assume that we started doing this because rails defaults shifted, not because somebody decided it was a good idea?

GerryG: But do we know how/why or is this maybe accidental?

I'm guessing there are some details of all of it that are unclear at this point. I'm just curious what we do and don't know.

ethn: No, we don't really know how/why.  my guess is that defaults for autogeneration of schema.rb changed with rails 4. I haven't looked into that, though, so.

PR #492, opened 2015-08-29
read events

xithan:


card/mod/01_core/set/all/location_history.rb
```diff
@@ -0,0 +1,10 @@
+event :discard_deleted_locations, :after=>:store, :on=>:delete do
+  Env[:controller].discard_locations_for self
+  if success.target == self
+    success.target = :previous
+  end
+end
+
+event :save_current_location, :before=>:render_view, :on=>:read do
+  Env[:controller].save_location
+end
```
xithan: after the event post-it summit and the grand API discussion a relationship therapy for cards and controller?

GerryG:
I think I get it.  Initially we didn't have any events triggering on the read phase, or on read actions at all.

Good to have this getting even more regularized.

wagn/rails/controllers/card_controller.rb
```diff
@@ -248,7 +243,9 @@ def show view = nil, status = 200
     else
       args = { :text=>result, :status=>status }
       args[:content_type] = 'text/text' if format == :file
-      render args
+      card.run_callbacks :render_view do
```
ethn: why the name "render_view"?  seems a little misleading, since it doesn't get called whenever we render a view.  should it just be "read" or "show" maybe?

xithan: "read" doesn't work because then API would be `:before=>:read, :on=>:read`. I chose "render_view" because it's not before/after "show" it's before/after we render the view in show.

card/lib/card/env.rb
```diff
@@ -2,6 +2,57 @@ class Card
...
+        #warn "sess #{session.class}, #{session.object_id}"
+        session[:history] ||= [card_path('')]
+        if session[:history]
+          session[:history].shift if session[:history].size > 5
+          session[:history]
+        end
+      end
+
+      def save_location card
+        return if Env.ajax? || !Env.html? || !card.known? || (card.codename == 'signin')
```
ethn: not suggesting any major change right now, but the html? test makes me wonder if read events should be view specific.  (or if they should just be called from views?)

I do think the "signin" thing should be handled in the set somehow (at some point)

wagn/rails/controllers/card_controller.rb
ethn: re read: you're completely right.  read is an "action", so it can't be an event.

re view -- but there might be many views, even for the one card.   

Seems like we might want to give "page" a formal meaning at some point.

Oh well, guess I can't do better yet.

xithan:
To what does "how come?" refer?

wagn/rails/controllers/card_controller.rb
xithan: render_page? but for an ajax request "page" is not the best description. Maybe just "render" because that's the point where all the view rendering starts?

xithan: just noticed that this callback is in the wrong spot. I correct that in a new PR and call it 'show'

PR #497, opened 2015-09-08
performance log improvements

xithan:


card/mod/01_core/set/all/fetch.rb
```diff
@@ -23,48 +23,50 @@ module ClassMethods
...
+        else
+          card.include_set_modules unless opts[:skip_modules]  # need to load modules here to call the right virtual? method
+          return unless card.virtual?
+        end
+        card.name = mark.to_s if mark && mark.to_s != card.name
+      end
+
+      card.include_set_modules unless opts[:skip_modules]
+      card
+    end
```
ethn: I assume there's no change but the "with_logging" here, right?

ethn:
Wow, Philipp.  This is pretty impressive stuff.  I thought when I saw the demo today that it must not have been as hard as I had expected. Looking at the code, I see that it was hard, but that you tackled it in a really cool way.  Not sure I understand how you did this so fast!

card/mod/01_core/set/all/fetch.rb
xithan: right, only the indentation changed with the "with_logging" block

xithan:
Yeah, the result is not very impressive: only five numbers. 
> Not sure I understand how you did this so fast!

I split one big file into several small files so the diff looks like I wrote a lot of new code ;)

PR #503, opened 2015-09-17
file rendering speedup / cleanup

ethn:
I notices loads of file-related queries on even simple views (eg html core) and requests.  This is probably not the most urgent optimization issue, but if this gets cleaned up, it gets rid of a ton of noise in dev logs, and I think it may make it easier to find bigger problems.


card/mod/05_standard/lib/file_uploader.rb
```diff
@@ -57,8 +57,8 @@ def filename
   def extension
     if file && file.extension.present?
       ".#{file.extension}"
-    elsif original_filename
-      File.extname(original_filename)
+#    elsif original_filename
+#      File.extname(original_filename)
```
ethn: Is there a reason the above is necessary?  Those lookups are super expensive, and we do them many times (which is weird, too.  Why does this get called so many times??)

card/mod/03_machines/lib/card/machine.rb
```diff
@@ -178,7 +178,7 @@ def machine_output_path
     end
 
     def ensure_machine_output
-      if !output = fetch(:trait => :machine_output) or !output.selected_content_action_id
+      if !output = fetch(:trait => :machine_output) #or !output.selected_content_action_id
```
ethn: I suspect the above is actually needed.  But I think we can do this more efficiently (see comment re selected actions below)

card/mod/05_standard/set/type/file.rb
```diff
@@ -1,5 +1,15 @@
...
 
+module SelectedAction
+  def select_action_by_params params
+    # test whether request is for current action using card content
+    # if so, we can skip the action lookup
+    rev_id = params[:rev_id]
+    super unless rev_id && db_content =~ /\/#{rev_id}\./
+  end
+end
+include SelectedAction
```
ethn: right now, most file requests send a rev_id, and we end up looking in the database for that rev_id.  But there's no need to do that, since it's right there in the db content.

I'm not sure this is the best implementation of a fix (feels a bit hacky), but it appears to save a ton of lookups.

At present the machine output is doing a lot of these same kinds of extra lookups.  I think if we fixed that we would be able to load, for example, the home page of a default wagn -- and all associated images -- with zero queries once the cards are cached.  Would be very good to get that kind of page down under 100ms on most servers.

card/mod/01_history/set/all/content_history.rb
```diff
@@ -70,9 +70,13 @@ def with_selected_action_id action_id
...
-  @selected_action_id ||
- (@current_action && (new_card? || @current_action.new_content? || db_content_changed?) && @current_action.id) ||
-  last_content_action_id
+  @selected_action_id || new_content_action_id || last_content_action_id
+end
+
+def new_content_action_id
+  if @current_action && (new_card? || @current_action.new_content? || db_content_changed?)
+    @current_action.id
+  end
```
ethn: no functional change here.

card/mod/05_standard/set/type/file.rb
```diff
@@ -1,5 +1,20 @@
...
+  end
+
+  def last_content_action_id
+    # find action id from content (saves lookups)
+    db_content.to_s.split(/[\/\.]/)[1]
+  end
+
+end
+include SelectedAction
+
```
ethn: here's the meet of this PR.

card/mod/02_basic_types/set/type/pointer.rb
```diff
@@ -179,13 +179,19 @@ def wrap_item item, args
 
 
 format :css do
+
+  #generalize to all collections?
+  def default_item_view
+    params[:item] || :content
+  end
```
ethn: fixes debug=style

card/mod/05_standard/lib/file_uploader.rb
xithan: there used to be cases where this was the only way to get the extension but could be that this is no longer necessary

PR #508, opened 2015-09-23
Fastquery

ethn:
general premise of this PR: subselects are too slow on large datasets, so we need to make WQL produce more conventional joins.


card/lib/card/query/card_clause.rb
```diff
@@ -29,8 +29,9 @@ def build query
 
       def initialize query
         @mods = MODIFIERS.clone
-        @clause, @joins = {}, {}
+        @conditions, @joins = {}, {}
```
ethn: @clause is a poor/confusing name

xithan:
Sorry for the spam. I invited this nice pet to watch us: http://houndci.com  Looks like we have to train the hound to be less yappy. It shouldn't bark about single quotes but maybe the hound gives us grounds to switch to Ruby 1.9 hash syntax.

card/lib/card/query.rb
```diff
@@ -13,61 +29,69 @@ class Query
...
+          Card.fetch name, new: {}
         end
       else
-        simple_run retrn
+        get_results retrn
       end
     end
 
-    def simple_run retrn
+    def get_results retrn
```
xithan: a comment what the return value is would be helpful here

card/lib/card/query.rb
```diff
@@ -163,14 +196,22 @@ def interpret_by_key clause
     end
```
xithan: picky: `val` instead of `clause[key]` seems to be an option in line 155

card/lib/card/query.rb
```diff
@@ -1,11 +1,27 @@
...
+  # Query "statements" (objects, really) are made in WQL (Wagn Query
+  # Language). Because WQL is used by Wagneers, the primary language
+  # documentation is on wagn.org. (http://wagn.org/WQL_Syntax). Note that the
+  # examples there are in JSON, like Search card content, but statements in
+  # Card::Query are in ruby form.
+  #
+  # In Wagn's current form, Card::Query generates and executes SQL statements.
+  # However, the SQL generation is largely (not yet fully) separated from the
+  # WQL statement interpretation.
+
```
xithan: a problem I have with following the code: I don't know what exactly we want to build. An example how the query object representation of an WQL statement looks like could help

card/lib/card/query.rb
ethn: wouldn't that just alter the local variable definition of val without changing the hash?

card/lib/card/query.rb
ethn: right.

card/lib/card/query.rb
ethn: Yeah, I'll write that up.  Just thinking about what I'd write makes me realize there are significant inconsistencies.  The main thing is to populate @joins and @conditions, but while @joins is a list of Join objects, @conditions is more complex.

Almost everything else from the raw statement goes in @mods and is pretty much ready for SqlStatement.

Sigh...  It's reasonably clean code now, but it's not super simple :)

PR #509, opened 2015-09-25
Fix file upload as a field

chuenlok:


wagn/features/attach.feature
```diff
@@ -41,4 +41,19 @@ Feature: Conflict
...
     Then I should see an image of size "large" and type "jpg"
 
+  Scenario: uploading a file as a field in a form
+    When I go to  "/new Cardtype"
+    And I fill in "card_name" with "complicated_card"
+    And I press "Submit"
+    And I create HTML card "complicated card+*type+*structure" with content "{{+image}}{{+description}}"
+    And I go to "/new complicated_card"
+    And I fill in "card_name" with "Vignesh has a complicated relationship"
+    And I wait until ajax response done
```
xithan: That's better than just waiting a fixed time. I suggest you replace the "wait a sec" statements in my scenarios above with that.

card/mod/05_standard/set/abstract/attachment.rb
```diff
@@ -110,6 +110,13 @@ def upload_cache_card
   @upload_cache_card ||= Card["new_#{attachment_name}".to_sym ]
 end
 
+def cached_upload= value
+  @cached_upload = value
+end
+
+def cached_upload
+  @cached_upload
```
xithan: I'm wondering whether the name should be more explicit about what it contains. `action_id_of_cached_upload`? If you think that's too long then at least a comment here that it is an action id would be useful

ethn: agree.

chuenlok:
I don't understand why that only test fails. Any idea?

wagn/features/attach.feature
```diff
@@ -36,9 +36,43 @@ Feature: Conflict
...
+    And I upload the image "image.png"
+    And I wait until ajax response done
+    And I press "Submit" 
+    And I edit "Vignesh has a complicated relationship"
+    And I upload the image "image2.jpg"
+    And I wait until ajax response done
+    Then I should see "image2.jpg 69.8 KB"
+    And I press "Submit"
+    Then I should see an image of size "medium" and type "jpg"
+    Then I wait until ajax response done
```
chuenlok: this line solves the problem. I think ending a test too quickly on some cases may affect db cleaner working.

chuenlok:
This test will fail until having Philipp's new subcards API.

chuenlok:
I don't what to do for `Colon assignment without proper spacing` as the default for coffee should be 1 space front wrapping the colon. 
https://github.com/scalableminds/coffeelint-min-colon-spacing/blob/master/src/min_colon_spacing.coffee

card/mod/05_standard/set/abstract/attachment.rb
```diff
@@ -24,15 +24,16 @@ def self.included host_class
...
     upload_cache_card.selected_action_id = action.id
     upload_cache_card.select_file_revision
     assign_attachment upload_cache_card.attachment.file, action.comment
   end
 end
 
-event :assign_attachment_on_update, after: :prepare, on: :update, when: proc { |c| c.save_preliminary_upload? } do
-  if (action = Card::Action.fetch(Card::Env.params[:cached_upload]))
+event :assign_attachment_on_update, after: :prepare, on: :update,
+      when:  proc { |c| c.save_preliminary_upload? } do
```
chuenlok: and better way to do that?

wagn/features/attach.feature
ethn: like.

card/mod/03_machines/lib/javascript/wagn_mod.js.coffee
```diff
@@ -4,27 +4,65 @@ window.wagn ||= {} #needed to run w/o *head.  eg. jasmine
...
+    '.pointer-checkbox-list' : -> 
+      pointerContent @find('input:checked').map( -> $(this).val() )
+    '.pointer-select-list' : -> 
+      pointerContent @find('.pointer-select select').map( -> $(this).val() )
+    '.pointer-mixed' : -> 
+      element = '.pointer-checkbox-sublist input:checked,\
+                .pointer-sublist-ul input'
+      pointerContent @find(element).map( -> $(this).val() )
+    # must happen after pointer-list-ul, I think
+    '.perm-editor' : -> permissionsContent this
```
ethn: I think the hound may have led us to less readable code here.

card/mod/05_standard/set/abstract/attachment.rb
ethn: it's saying `when:` needs to align with `after:`

card/mod/05_standard/set/abstract/attachment.rb
ethn: this is the only thing I'm really seeing that needs to happen to get this PR merged.  seems like the val of @cached_upload would either be a File or true/false.  Hard to guess that it's an action_id.

PR #511, opened 2015-09-28
RSS errors on search type and pointer

chuenlok:


card/mod/02_basic_types/set/type/pointer.rb
```diff
@@ -218,6 +218,14 @@ def default_item_view
   end
 end
 
+format :rss do
```
chuenlok: I found that rss rendering in pointer cards does not contain cards details as search type does. I am not sure whether we want to do it like search type.

card/mod/05_standard/set/type/search_type.rb
```diff
@@ -188,8 +192,8 @@ def default_search_params
 format :rss do
   view :feed_body do |args|
     case raw_feed_items args
-    when Exception ; @xml.item { render :search_error }
-    when Integer   ; @xml.item { render :search_count }
+    when Exception ; @xml.item( render :search_error )
```
chuenlok: `{ } `does not work but `( )`. Not sure whether it is related to library upgrade or historical problems

PR #512, opened 2015-09-29
Queryclean

ethn:
most of these changes are cleanup, but also includes fix for context problem and adds API to do this:

    Query.run(statement)

instead of having to do this:

    Query.new(statement).run

(which still works, too)

### DIFF
--- a/card/lib/cardio.rb
+++ b/card/lib/cardio.rb
@@ -8,7 +8,7 @@ module Cardio
   CARD_GEM_ROOT = File.expand_path('../..', __FILE__)
 
   ActiveSupport.on_load :card do
-    if Card.count > 0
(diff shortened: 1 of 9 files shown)


card/lib/cardio.rb
```diff
@@ -8,7 +8,7 @@ module Cardio
   CARD_GEM_ROOT = File.expand_path('../..', __FILE__)
 
   ActiveSupport.on_load :card do
-    if Card.count > 0
+    if Card.take
```
ethn: ~150x faster.

wagn: :rocket:

xithan: :rocket:

PR #513, opened 2015-09-29
new subcards api and list/listed_by relations

xithan:
The "list" and "listed by" cardtypes provide a way to establish many-to-many relationships. The pattern is as follows:
Let A,B be cards of type letter and 1,2 cards of type number.
If you create list cards
 - A+numbers -> [[1]]\n[[2]]
 - B+numbers -> [[1]]
then the following listed by cards will contain the related letters:
 - 1+letters -> [[A]]\n[[B]]
 - 2+letters -> [[A]]]

The content of listed by cards is cached and gets updated after every change on a related list card.


card/lib/card/subcards.rb
```diff
@@ -0,0 +1,236 @@
...
+    subcards.add_field name, args
+  end
+
+  def remove_subcard name_or_card
+    subcards.remove name_or_card
+  end
+
+  def remove_subfield name_or_card
+    subcards.remove_field name_or_card
+  end
```
xithan: above the new subcards API

card/mod/01_core/set/all/fetch.rb
```diff
@@ -58,20 +50,47 @@ def fetch mark, opts={}
         return
       else
         card.include_set_modules unless opts[:skip_modules]  # need to load modules here to call the right virtual? method
-        return unless card.virtual?
+        return unless card.virtual? || opts[:subcard]
       end
-      card.name = mark.to_s if mark && mark.to_s != card.name
+      card.name = mark.to_s if mark && mark.to_s != card.name && !opts[:subcard]
```
xithan: This feels a bit hacky. Subcards are often new cards so I couldn't fetch them from the cache and I don't want to change the name if I fetch them with a name variant.
Maybe we want a more general solution?
Something in the direction of `Card.fetch 'name', include_new: true` instead of `Card.fetch 'name', subcard: true`?

card/mod/01_core/set/all/fetch.rb
```diff
@@ -86,10 +105,15 @@ def known? mark
...
-  def expire name
+  def expire name, subcards=false
     #note: calling instance method breaks on dirty names
     key = name.to_name.key
     if card = Card.cache.read( key )
+      if subcards
+        card.expire_subcards
+      else
+        card.preserve_subcards
+      end
```
xithan: the workaround for the "expire" problem. There are many cases where the subcards have to survive a "expire" call

card/spec/lib/card/subcards_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      @card.save!
+      expect(Card['A+sub'].content).to eq 'sub content'
+    end
+  end
+  #
+  # describe '#remove_subfield' do
+  #
+  # end
+
+end
```
xithan: I don't know why the wagn logo changed. I'll restore the old version.

card/mod/01_core/set/all/tracked_attributes.rb
```diff
@@ -113,18 +115,18 @@ def extract_subcard_args! args={}
...
     end
   end
   # FIXME really shouldn't be instantiating all the following bastards.  Just need the key.
   # fix in id_cache branch
-  self.dependents.each       { |c| c.expire }
-  # self.referencers.each      { |c| c.expire }
-  self.name_referencers.each { |c| c.expire }
+  self.dependents.each       { |c| c.expire(true) }
+  # self.referencers.each      { |c| c.expire(true) }
+  self.name_referencers.each { |c| c.expire(true) }
```
xithan: the only case where we want to get rid of the subcards

.ruby-style.yml
```diff
@@ -102,14 +102,14 @@ Style/CollectionMethods:
   PreferredMethods:
     collect: map
     collect!: map!
-    find: detect
```
ethn: nice thing about detect is that you wouldn't confuse it with ActiveRecord#find...

card/mod/01_core/set/all/fetch.rb
ethn: nice.

card/mod/01_core/set/all/fetch.rb
```diff
@@ -20,59 +20,70 @@ module ClassMethods
...
+
+  def fetch_local mark, opts = {}
+    fetch mark, opts.merge(:local_only=>true)
+  end
+
+  def fetch_id mark
+    if mark.is_a?(Integer)
+      mark
+    elsif mark.is_a?(Symbol) && Card::Codename[mark]
+      Card::Codename[mark]
```
ethn: shouldn't a symbol always be a codename?  If Card::Codename[mark] is nil, we should return nil, no?

card/mod/01_core/set/all/phases.rb
```diff
@@ -1,31 +1,32 @@
...
 def abortable
   yield
 rescue Card::Abort => e
   if e.status == :triumph
-    @supercard ? raise( e ) : true
+    @supercard ? raise(e) : true
   elsif e.status == :success
     if @supercard
-      @supercard.subcards.delete_if { |k,v| v==self }
+      @supercard.subcards.delete(key)
```
ethn: ahhhh.

card/mod/01_core/spec/set/all/initialize_spec.rb
```diff
@@ -19,38 +19,68 @@
...
+
+    it 'handles type_code' do
+      expect(Card.new(type_code: 'phrase').type_id).to eq(Card::PhraseID)
+      expect(Card.new(type_code: :phrase).type_id).to eq(Card::PhraseID)
+    end
+
+    it 'handles type_id' do
+      expect(Card.new(type_id: Card::PhraseID).type_code).to eq(:phrase)
+    end
+  end
```
ethn: nice to have these.

ethn: I suppose we should probably add a :skip_type_lookup test, too.

PR #516, opened 2015-10-01
Joinnot

ethn:
"not" is now working as a subjoin.


card/lib/card/query/attributes.rb
```diff
@@ -292,15 +286,18 @@ def conjoin val, conj
...
+        #subselect = Query.new clause_to_hash(val).merge( return: :id )
+        #join_alias = "not#{table_id force=true}"
+        #joins << Join.new(
+        #  from: self,
+        #  to_table: subselect,
+        #  to_alias: join_alias,
+        #  side: 'LEFT'
+        #)
+        #add_condition "#{join_alias}.id is null"
       end
```
ethn: the new "not"

PR #519, opened 2015-10-07
Arglessness

ethn:
making it so that `|args|` is optional in view definition.

### DIFF
--- a/card/lib/card/format.rb
+++ b/card/lib/card/format.rb
@@ -204,7 +204,8 @@ class Card
         args = default_render_args view, args
         with_inclusion_mode view do
           Card::ViewCache.fetch(self, view, args) do
-            send "_view_#{ view }", args
+            method = method "_view_#{ view }"
+            method.arity == 0 ? method.call : method.call(args)
           end
         end
       end
--- a/card/lib/card/name.rb
+++ b/card/lib/card/name.rb
@@ -5,61 +5,81 @@ require 'smart_name'
 
 class Card
   class Name < SmartName
+    RELATIVE_REGEXP = /\b_(left|right|whole|self|user|main|\d+|L*R?)\b/
 
     self.params  = Card::Env # yuck!
     self.session = proc { Card::Auth.current.name }
     self.banned_array = ['/']
 
     def star?
-      simple? and '*' == s[0,1]
+      simple? && '*' == s[0, 1]
     end
 
     def rstar?
-      right and '*' == right[0,1]
+      right && '*' == right[0, 1]
     end
 
     def trait_name? *traitlist
       junction? && begin
         right_key = right_name.key
-        !!traitlist.find do |codename|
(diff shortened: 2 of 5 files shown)


card/lib/card/format.rb
```diff
@@ -204,7 +204,8 @@ def render view, args={}
         args = default_render_args view, args
         with_inclusion_mode view do
           Card::ViewCache.fetch(self, view, args) do
-            send "_view_#{ view }", args
+            method = method "_view_#{ view }"
+            method.arity == 0 ? method.call : method.call(args)
```
ethn: possible that it's wiser to handle this at view definition time than here.  but it works :)

card/lib/card/name.rb
```diff
@@ -5,61 +5,81 @@
...
+      !relative?
+    end
+
+    def stripped
+      s.gsub RELATIVE_REGEXP, ''
+    end
+
+    def starts_with_joint?
+      s =~ /^\+/
     end
```
ethn: all of this stuff was mostly to make template_rule simpler, and I was only working on that because it was in base.rb (which it shouldn't be)

card/mod/02_basic_types/set/all/base.rb
```diff
@@ -89,34 +86,38 @@ def show view, args
     end.inspect
   end
 
-
-
-  #none of the below belongs here!!
-
+  # none of the below belongs here!!
 
   view :template_rule, tags: :unknown_ok do |args|
```
ethn: I don't think this method is well covered in our tests.  I probably have to remedy that before I can merge this.

card/mod/02_basic_types/set/all/base.rb
```diff
@@ -7,38 +7,37 @@ def show view, args
...
 
-  view :name,     closed: true, perms: :none do |args| card.name                           end
-  view :key,      closed: true, perms: :none do |args| card.key                            end
-  view :title,    closed: true, perms: :none do |args| args[:title] || card.name           end
-
-  view :linkname, closed: true, perms: :none do |args| card.cardname.url_key               end
-  view :url,      closed: true, perms: :none do |args| card_url _render_linkname           end
-  view :url_link, closed: true, perms: :none do |args| web_link card_url(_render_linkname) end
+  simple_args = { closed: true, perms: :none }
+  view(:name,     simple_args) { card.name                           }
```
ethn: I haven't gotten the local configs set up yet, but rubocop seems more permissive on this stuff locally.  (And it seems to matter whether it's done for alignment purposes)

card/lib/card/name.rb
xithan: the tests fail because you call `starts_with_joint?` on a string in line 70

ethn: yeah, saw that this morning.  Happy that it's tested :)

PR #522, opened 2015-10-19
bugfix: mod files not editable

xithan:
the test that should have detected this was wrong too.


card/mod/05_standard/lib/file_uploader.rb
```diff
@@ -146,7 +146,7 @@ def retrieve_dir
   end
 
   def mod_file?
-    @mod ||= model.mod_file?
+    model.mod_file?
```
ethn: this is kind of an expensive lookup, no?  Is it safe to assume this is getting incorrectly cached at some point?  Is it possible we just need to clear it?

The builds appear to be getting slower, but it may not be because of this PR.

wagn: This was the reason that if you update a mod file the file was stored in the wrong place (in the mod instead of the files directory). It doesn't seem very expensive to me. It checks regular expressions on the card's content

ethn: OK, I was thinking there were rule lookups there, but I think I was mistaken.

PR #527, opened 2015-10-28
fix paging problem

chuenlok:
item structure will lose after click the paging more than once

### DIFF
--- a/card/mod/05_standard/set/all/rich_html/wrapper.rb
+++ b/card/mod/05_standard/set/all/rich_html/wrapper.rb
@@ -11,9 +11,9 @@ format :html do
 
     @@slot_option_keys.inject(options_hash) do |hash, opt|
       if args[opt].present?
-        hash[opt] = args[opt] 
+        hash[opt] = args[opt]
       elsif Env.params['slot'].present? && Env.params['slot'][opt.to_s].present?
-        hash[opt] = Env.params['slot'][opt.to_s] 
+        hash[opt] = Env.params['slot'][opt.to_s]
       end
       hash
     end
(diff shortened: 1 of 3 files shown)


wagn/features/step_definitions/wagn_steps.rb
```diff
@@ -342,6 +348,12 @@ def signed_in_as username
   page.has_css?(".diff-#{css_class}", text: text)
 end
 
+Then /^I should see css class "([^\"]*)" within "(.*)"$/ do |css_class, selector|
```
chuenlok: Can we exempt the line restriction in cucumber? It seems regular expression does not support multi-lines. I am not sure.

card/mod/05_standard/set/all/rich_html/wrapper.rb
```diff
@@ -10,7 +10,11 @@ def slot_options args
     options_hash[:subslot] = 'true' if args[:subslot]
 
     @@slot_option_keys.inject(options_hash) do |hash, opt|
-      hash[opt] = args[opt] if args[opt].present?
+      if args[opt].present?
+        hash[opt] = args[opt]
+      elsif Env.params['slot'].present? && Env.params['slot'][opt.to_s].present?
+        hash[opt] = Env.params['slot'][opt.to_s]
+      end
```
ethn: I think this will make the params override every slot on the page.

chuenlok: How about adding a checking like
```ruby
elsif card.id == Card[Env.params['id']].id && Env.params['slot'].present? && Env.params['slot'][opt.to_s].present?
        hash[opt] = Env.params['slot'][opt.to_s]
      end
```

ethn: that's making a mess.   The question is why the args aren't already there as they should be.

wagn/features/step_definitions/wagn_steps.rb
ethn: could do.  could also just put regexp in a variable or variables in regexp...

card/mod/05_standard/set/all/rich_html/wrapper.rb
ethn: I've got a pr coming.  Your tests are a big help!!

PR #528, opened 2015-11-02
Fix paging

ethn:
Uses henry's test but attempts a simpler fix.


card/mod/05_standard/set/all/rich_html/content.rb
```diff
@@ -11,7 +11,7 @@ def show view, args
       self.render :layout
     else
       view ||= args[:home_view] || :open
-      @inclusion_opts = args.delete(:items)
+      @inclusion_opts = args[:items].clone
```
ethn: Looks like I may be the guilty party; I added this line in late June.  

Setting the @inclusion_opts means we get the correct opts for the first xhr request (page 2).  But deleting items means we don't correctly pass the arguments onto the second xhr request (page 3).

PR #529, opened 2015-11-05
reduce unnecessary descendant processing

ethn:
Trying to speed up a specific wikirate migration.  Issue is that there are ~145K cards ending in +cached_count, and an update to that card goes through and instantiates all of them even though they are not impacted.

It's just clearing their caches, but that should be unnecessary unless the name is changed, which it isn't.

### DIFF
--- a/card/mod/01_core/set/all/name.rb
+++ b/card/mod/01_core/set/all/name.rb
@@ -2,9 +2,9 @@ require 'uuid'
 
 module ClassMethods
   def uniquify_name name, rename=false
-    return name unless Card[name]
+    return name unless Card.exists?(name)
     uniq_name = "#{name} 1"
-    while Card[uniq_name]
+    while Card.exists?(uniq_name)
       uniq_name.next!
     end
     return uniq_name unless rename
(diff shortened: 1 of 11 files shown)


card/mod/01_core/set/all/name.rb
```diff
@@ -120,23 +120,37 @@ def left_or_new args={}
...
   end
-  @dependents
+  @descendant_names
+end
+
+def descendants
+  # children and children's children
+  # NOTE - set modules are not loaded
+  # -- should only be used for name manipulations
+  descendant_names.map { |name| Card.quick_fetch name }
```
ethn: renamed to descendants, since this means children of children.

card/mod/01_core/set/all/tracked_attributes.rb
```diff
@@ -143,12 +143,10 @@ def extract_type_id! args={}
...
+end
 
-  # FIXME: really shouldn't be instantiating all the following bastards.
-  # Just need the key.
-  # fix in id_cache branch
-  dependents.each       { |c| c.expire(true) }
-  # self.referencers.each      { |c| c.expire(true) }
+event :expire_related_names, before: :expire_related, changed: :name do
+  # FIXME: look for opportunities to avoid instantiating the following
+  descendants.each { |c| c.expire(true) }
```
ethn: main change is separating out this event, which only happens on name changes.

PR #531, opened 2015-11-11
Tokens

ethn:
Tokenized authentication support for all requests

Not ready to merge


wagn/rails/controllers/card_controller.rb
```diff
@@ -38,93 +32,126 @@ def delete
...
+    end
+  end
+
+  def card_attr_from_params
+    # clone so that original params remain unaltered.  need deeper clone?
+    opts = params[:card] ? params[:card].clone : {}
+    # for /new/:type shortcut.  we should fix and deprecate this.
+    opts[:type] ||= params[:type] if params[:type]
+    # move handling to Card::Name?
+    opts[:name] ||= params[:id].to_s.gsub('_', ' ')
```
xithan: looks like houndci has a valid point here. 
I run this script https://github.com/JuanitoFatas/fast-ruby/blob/master/code/string/gsub-vs-tr.rb with the result that gsub is 3.85x slower than tr

card/lib/card/auth.rb
```diff
@@ -33,13 +46,12 @@ def encrypt password, salt
 
       # find accounted by email
```
xithan: you forgot to adjust the comment :)

card/lib/card/auth.rb
```diff
@@ -13,19 +13,32 @@ module Auth
...
-        if Card.config.no_authentication ||
-           password_authenticated?(account, password.strip)
-          accounted.id
+        account = Auth[email]
+        case
+        when !account                                 then nil
+        when !account.active?                         then nil
+        when Card.config.no_authentication            then account
+        when password_valid?(account, password.strip) then account
         end
```
xithan: much more readable :+1:

ethn: agree.  I should note that both Auth[] and Auth.authenticate now return +*account cards (rather than ids of cards with accounts), which is part of why it's simpler. So in authenticating PK's email, it would formerly return PK's id but now returns the card PK+*account

If we use either in mods, we will need to change things accordingly.

card/lib/card/auth.rb
ethn: ...which implies you already figured out my last comment :)  thanks for spotting that.

card/mod/05_standard/set/right/account.rb
```diff
@@ -97,35 +77,34 @@ def token_error tcard, val
...
-      message: "Sorry, this token has expired. Please check your email for a new password reset link."
-    }
-    abort :success
   else
-    abort :failure, "error resetting password: #{result}" # bad token or account
+    error_msg = errors.first.last
+    send_reset_password_token
+    msg = "Sorry, #{error_msg}. " \
+          'Please check your email for a new password reset link.'
+    success << { id: '_self', view: 'message', message: msg }
```
ethn: this is a functional change.  Any time there is a token problem, we just send out a new email with a new token.  Simplifies the code, and is mostly what people want.

card/mod/05_standard/set/right/token.rb
```diff
@@ -3,3 +3,29 @@
...
+  errors.add *error if error
+end
+
+def expired?
+  !permanent? && updated_at <= term.ago
+end
+
+def permanent?
+  false
+end
```
ethn: Haven't decided how to make a token permanent yet, but I was thinking to make it so you could add an +*expiration card to customize whether and when a token expires.

card/mod/05_standard/set/right/token.rb
```diff
@@ -3,3 +3,29 @@
 view :raw do |args|
   "Private data"
 end
```
ethn: btw, tokens show up in history.

wagn/rails/controllers/card_controller.rb
ethn: good boy, :dog2:

PR #533, opened 2015-11-12
test case for delete children card, fix empty file card content

chuenlok:
I found that the deletion of children is working normally. The problems seems only on wikirate's metric values.

Also included the fix of the file subcard. This will cause the empty file subcards are not removed while processing subcards.


chuenlok:
I just found a problem of deleting subcards but I don't think I can make a test case in simple wagn.

Let me give an example here.
Assume we have these cards in database
```
Richard Mills+Employees+CA+2014+value
Richard Mills+Employees+CA+2014+source
Richard Mills+Employees+CA+2014
Richard Mills+Market Value+CA+2014+value
Richard Mills+Market Value+CA+2014+source
Richard Mills+Market Value+CA+2014
Richard Mills+Assets+CA+2014+value
Richard Mills+Assets+CA+2014+source
Richard Mills+Assets+CA+2014
Richard Mills+Annual Profits+CA+2014+value
Richard Mills+Annual Profits+CA+2014+source
Richard Mills+Annual Profits+CA+2014
Richard Mills+Annual Sales+CA+2014+value
Richard Mills+Annual Sales+CA+2014+source
Richard Mills+Annual Sales+CA+2014
CA
```
with set rules
```
Metric value+value+*type plus right
Metric value+source+*type plus right
```
(Metric value is like `Richard Mills+Annual Sales+CA+2014`)

If I try to delete card `CA`, some of the `+value` and `+source` cards are not deleted.

As we are deleting `CA`, cards like `Richard Mills+Annual Sales+CA`, `Richard Mills+Employees+CA` and their subcards will be deleted as well. 
```
event :validate_delete_children, after: :approve, on: :delete do
  children.each do |child|
    child.trash = true
    add_subcard child
    unless child.valid?
      child.errors.each do |field, message|
        errors.add field, "can't delete #{child.name}: #{message}"
      end
    end
  end
end
```
This part will mark all the subcards to trash. 

```
event :store_subcards, after: :store do
  subcards.each do |subcard|
    subcard.save! validate: false if subcard != self # unless @draft
  end

  # ensures that a supercard can access subcards of self
  # eg. <user> creates <user+*account> creates <user+*account+*status>
  # <user> changes <user+*account+*status> in event activate_account
  Card.write_to_local_cache self
end
```
this event will update those cards as trashed.

After the `store_subcards`, 
```
event :reset_type_specific_fields, after: :store do
  Auth.as_bot do
    binding.pry
    Card.search left: { left_id: type_id }, right: {codename: 'type_plus_right'} do |set_card|
      set_card.reset_set_patterns
    end
  end
end
```
`set_card.reset_set_patterns` will clear the cached of the cards in the set patterns. 
Therefore, after the `+value`, `+source` of `Richard Mills+Employees+CA+2014` are deleted, the `set_card.reset_set_patterns` will reset the trash flags of other `+value`, `+source` cards in the cache that erase the trash flags. 

The result of deletion will be 

~~Richard Mills+Employees+CA+2014+value~~

~~Richard Mills+Employees+CA+2014+source~~

~~Richard Mills+Employees+CA+2014~~

Richard Mills+Market Value+CA+2014+value

Richard Mills+Market Value+CA+2014+source

~~Richard Mills+Market Value+CA+2014~~

Richard Mills+Assets+CA+2014+value

Richard Mills+Assets+CA+2014+source

~~Richard Mills+Assets+CA+2014~~

Richard Mills+Annual Profits+CA+2014+value

Richard Mills+Annual Profits+CA+2014+source

~~Richard Mills+Annual Profits+CA+2014~~

Richard Mills+Annual Sales+CA+2014+value

Richard Mills+Annual Sales+CA+2014+source

~~Richard Mills+Annual Sales+CA+2014~~

~~CA~~

ethn:
Henry, this is a very clear and helpful explanation of the problem.  Thank you!

I think the solution probably involves cleaning up our phase handling.  In general, the concept behind phases was that ALL cards involved in a given act would complete a phase before going on to the next.

I think we currently have six phases:
1. prepare - perform any card manipulations to get them ready for approval
2. approve - approve for storage
3. store  - actually commit changes to database
4. stored - maybe we should rename this "clean"?  it's mostly about cleaning things up...
5. extend - extra stuff to happen before finishing act
6. subsequent - stuff that happens after act completion (using active jobs)

So, in theory, it should fix things to move the `reset_type_specific_fields` from `after: :store` to `after: :stored`, because all the subcards would be deleted before the reset_set_patterns was called on *any* card.

In practice, I don't think it would work, because I don't think the above are actually implemented as phases.  I may be wrong about this, but I think it currently actually works that these so-called phases are bundled together like this:

I. run prepare/approve for each card
2. run store/stored/extend for each card
3. run subsequent for each card

In other words, we pretend to have 6 phases, but we really have 3.

I think we should fix that, and I would also say that every phase should probably be a verb, so perhaps "stored" becomes "clean", "extend" becomes "finish", and subsequent becomes "followup"?

This may require some deep digging; right now we're really piggybacking on active record callbacks,

xithan:
How will we handle adding subcards if we make all six phases real phases? We allow it only in the prepare phase? That's a problem if you want to approve a card before you use it to add subcards. I think that's happening when we add sources on Wikirate.

ethn:
this works when I merge it in now.  Yay for phases!

PR #535, opened 2015-11-16
fix invalid tag operator

chuenlok:


card/lib/card/query/value.rb
```diff
@@ -15,7 +15,10 @@ def initialize rawvalue, query
       def parse_value rawvalue
         operator =
           case rawvalue
-          when Array;     rawvalue.shift
+          when Array
+            _operator = rawvalue[0]
+            rawvalue = rawvalue[1..-1]
+            _operator
```
ethn: ok, I get it now.

card/mod/05_standard/spec/set/type/search_type_spec.rb
```diff
@@ -35,6 +35,41 @@
...
+      )
+      html = search_card.format.render_open
+      expect(html).to have_tag('ul', with: { class: 'pagination paging' }) do
+        with_tag 'a', text: '1'
+        with_tag 'a', text: '2'
+        with_tag 'a', text: '3'
+      end
+      expect(html).to have_tag('div', with: { class: 'search-result-list' }) do
+        with_tag 'div', with: { id: 'Apple' }
+      end
```
ethn: This should probably be in Query spec, not in search type spec;  I don't think we need to render any html to test.  Something like:
````
query = { right_plus: ....
Card.search(query).items.should ==
````

Comment (chuenlok) [inline card/mod/05_standard/spec/set/type/search_type_spec.rb]: The problem won't happen in a normal query. 
The function call`Card.search(s)` in 
```
def item_cards params={}
  s = query(params)
  raise('OH NO.. no limit') unless s[:limit]
  # forces explicit limiting
  # can be 0 or less to force no limit
  Card.search(s)
end
```
will affect the `@query`'s params.

For example in wikirate's source filter,
part of the `@query` will be changed from 
`:right_plus=>[{"name"=>["in", "tag", "company", "topic"]` to 
`:name=>["tag", "company", "topic"]`.

This affect the further query of count for paging.

ethn: OK, thanks.  I'll have a look at this.

I might be able to write a simpler test that just makes sure we don't alter the query.

Question: does it also fix the bug if we change `@statement = statement.clone` to `@statement = statement.deep_clone` in query.rb?

chuenlok: it seems the `deep_clone` will cause stack level too deep as it will initialize the `query` again.

ethn: really?  wow, I thought it was just a simple hash.  I guess in the recursion process that becomes untrue?

PR #540, opened 2015-11-24
Operator

ethn:
building upon Henry's work on the WQL operator problem.


card/lib/card/query.rb
```diff
@@ -97,8 +97,7 @@ def initialize statement
     # Card::Fetch)
 
     def self.run statement
-      query = new statement
-      query.run
+      new(statement.deep_clone).run
```
ethn: I predict that if we deep clone before the initialization we won't have the same test failures.  We'll see ;)

card/lib/card/query/value.rb
```diff
@@ -13,14 +13,12 @@ def initialize rawvalue, query
...
-          when Integer;   '='
-          else raise("Invalid Condition Clause #{rawvalue}.inspect}")
-          end
-        [operator, rawvalue]
+        case rawvalue
+        when String  then ['=', rawvalue]
+        when Integer then ['=', rawvalue]
+        when Array   then [rawvalue[0], rawvalue[1..-1]]
+        else raise("Invalid Condition Clause #{rawvalue}.inspect}")
+        end
```
ethn: arguably more readable than the original.  much more concise than the patch.

card/lib/card/query.rb
ethn: Note that this isn't necessary to fix the specific operator bug, but it's probably wise to prevent other similar problems this way..

card/spec/lib/card/query_spec.rb
```diff
@@ -8,6 +8,13 @@
     Card::Query.run @query.reverse_merge return: :name, sort: :name
   end
 
+  it 'should not alter original statement' do
+    @query = { right_plus: { name: %w( in tag source ) } }
+    query_clone = @query.deep_clone
+    subject # runs query
+    expect(query_clone).to eq(@query)
+  end
```
ethn: I was able to reproduce the bug with this testing. I removed the other new tests, because they do a lot more processing for the same effect.  (I'd really like to get our test suite time down)

PR #541, opened 2015-11-25
climate change

xithan:
refactoring the worst spots for better "codeclimate"


card/lib/card/auth.rb
```diff
@@ -135,12 +139,12 @@ def as given_user
...
-        @@as_id   = tmp_id
-        @@as_card = tmp_card
-
-        value
+        yield
+      ensure
+        if block_given?
+          @@as_id   = tmp_id
+          @@as_card = tmp_card
+        end
```
xithan: seems safer to me to `ensure` that we return after the block to the previous settings

PR #542, opened 2015-11-27
make non-createable types configurable

xithan:


card/lib/card/query/sql_statement.rb
```diff
@@ -150,7 +150,7 @@ def trash_condition query
       def permission_conditions query
         return if Auth.always_ok?
         read_rules = Auth.as_card.read_rules
-        read_rule_list = read_rules.nil? ? 1 : read_rules.join(',')
+        read_rule_list = read_rules.present? ? read_rules.join(',') : 1
         "#{query.table_alias}.read_rule_id IN (#{read_rule_list})"
```
xithan: had a case where `read_rules` was an empty array and this produced invalid sql

card/lib/card/auth.rb
```diff
@@ -214,7 +213,9 @@ def always_ok_usr_id? usr_id
       def createable_types
         type_names = Auth.as_bot do
           Card.search type: Card::CardtypeID, return: :name,
-                      not: { codename: ['in'] + NON_CREATEABLE_TYPES }
+                      not: {
+                        codename: ['in'] + Card.config.non_createable_types
+                      }
```
xithan: made NON_CREATEABLE_TYPES a config option so that I can add something in a wikirate mod. The default is set in cardio.rb

ethn: I'll merge this in for now, but I would suspect this fits more naturally as a class variable than a Card.config setting.  I would think that the rule of thumb should be that we only use Card.config if we really want people manually changing the setting.

xithan: How can I change a class variable in a mod/deck? I tried to change NON_CREATEABLE_TYPES in wikirate but it didn't work

ethn: something like what we do in set.rb with Card::Format.extract_class_vars ?

ethn: (not a perfect analogue, I know...)

PR #543, opened 2015-12-03
fix disastrous "return"

xithan:


card/mod/01_core/chunk/link.rb
```diff
@@ -74,7 +74,7 @@ def inspect
     def replace_reference old_name, new_name
       replace_name_reference old_name, new_name
 
-      if Card::Content == @link_text
+      if Card::Content === @link_text
```
ethn: surprised hound is ok with this; I thought it preferred `is_a?`

PR #544, opened 2015-12-09
fix bad cache writing

ethn:
the raw mark (often relative) was being used for cache writing.  things like _L+note.

this was never being looked up.

### DIFF
--- a/card/lib/card/query.rb
+++ b/card/lib/card/query.rb
@@ -70,11 +70,12 @@ class Card
 
     DEFAULT_ORDER_DIRS =  { :update => "desc", :relevance => "desc" }
 
-    attr_reader :statement, :mods, :conditions,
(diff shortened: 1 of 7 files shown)


card/lib/card/query.rb
```diff
@@ -70,11 +70,12 @@ class Query
...
     DEFAULT_ORDER_DIRS =  { :update => "desc", :relevance => "desc" }
 
-    attr_reader :statement, :mods, :conditions,
+    attr_reader :statement, :mods, :conditions, :comment,
       :subqueries, :superquery
     attr_accessor :joins, :table_seq, :unjoined, :conditions_on_join
 
-    def initialize statement
+    def initialize statement, comment=nil
+      @comment = comment
```
ethn: add optional comment to queries.  Search cards send their name as a comment.

Comment shows up in rails logging.

PR #545, opened 2015-12-10
nest options for name variants

xithan:


card/mod/02_basic_types/set/all/base.rb
```diff
@@ -6,9 +6,20 @@ def show view, args
...
   # NAME VIEWS
-
   simple_args = { closed: true, perms: :none }
-  view(:name,     simple_args) { card.name                           }
+  view :name, simple_args do |args|
+    return card.name unless args[:variant]
+    args[:variant].split(/[\s,]+/).inject(card.name) do |name, variant|
+      if ::Set.new([
+        :downcase, :upcase, :singularize, :pluralize, :capitalize,
+        :swapcase, :reverse, :succ
```
ethn: these are all verbs. `variant: singularize` seems a little odd to me.

I would suggest: lowercase, uppercase, singular, plural, and capitalized.  (the others are pretty funny.  maybe they should be easter eggy?)

ethn:
Good stuff!

card/mod/02_basic_types/set/all/base.rb
xithan: ok, maybe I was to lazy with just using the method names. I'm going to deverbify them. I'm fine with the easter egg solution. There is also titleize in ActiveSupport. That's probably another option worth to mention in the documentation.

PR #546, opened 2015-12-10
Cleanfetch

ethn:
a further attempt to clean up fetch handling (with an eye to some optimization ideas)

### DIFF
--- a/card/mod/01_core/set/all/fetch.rb
+++ b/card/mod/01_core/set/all/fetch.rb
@@ -61,7 +61,7 @@ module ClassMethods
     if mark.is_a?(Integer)
       mark
     else
-      card = quick_fetch mark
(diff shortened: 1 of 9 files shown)


card/mod/01_core/set/all/fetch.rb
```diff
@@ -24,23 +24,16 @@ module ClassMethods
...
-    elsif !card || (card.type_unknown? && !skip_type_lookup?(opts))
+    if (new_card = new_for_cache card, mark, opts)
+      card = new_card
       needs_caching = true
-      card = new_for_cache mark, opts # new (or improved) card for cache
     end
 
+    return if card.nil?
     write_to_cache card, opts if needs_caching
     standard_fetch_results card, mark, opts
```
ethn: fetch has never looked so fetching.

PR #549, opened 2015-12-15
Transpeed

ethn:
trying to speed up transactions

### DIFF
--- a/card/mod/01_core/set/all/content.rb
+++ b/card/mod/01_core/set/all/content.rb
@@ -34,10 +34,6 @@ def clean_html?
   true
 end
 
-def history?
-  false
-end
-
 def save_content_draft _content
   clear_drafts
 end
--- a/card/mod/01_core/set/all/permissions.rb
+++ b/card/mod/01_core/set/all/permissions.rb
@@ -32,32 +32,39 @@ end
 def who_can action
   # warn "who_can[#{name}] #{(prc=permission_rule_card(action)).inspect},
   # #{prc.first.item_cards.map(&:id)}" if action == :update
-  permission_rule_card(action).first.item_cards.map &:id
+  permission_rule_card(action).item_cards.map &:id
+end
+
+def permission_rule_id_and_class action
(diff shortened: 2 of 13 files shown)


card/mod/01_core/set/all/permissions.rb
```diff
@@ -138,24 +145,26 @@ def ok_to_comment
...
-      end
+event :clear_read_rule, before: :store, on: :delete do
+  self.read_rule_id = self.read_rule_class = nil
+end
+
+event :set_read_rule, before: :store, on: :save do
+  # avoid doing this on simple content saves?
+  read_rule_id, read_rule_class = permission_rule_id_and_class(:read)
+  self.read_rule_id = read_rule_id
+  self.read_rule_class = read_rule_class
```
ethn: I did some cleanup here because this was slow, but the real reason was the unnecessary cache clearing

card/mod/05_email/set/all/follow.rb
```diff
@@ -2,30 +2,33 @@
...
+event :cache_expired_for_new_set,
+      before: :store, on: :create,
+      when: proc { |c| c.type_id == Card::SetID } do
   Card.follow_caches_expired
 end
 
-event :cache_expired_for_type_change, before: :store, changed: :type_id do
-  #FIXME expire (also?) after save
+event :cache_expired_for_type_change, before: :store, on: :update,
+                                      changed: :type_id do
```
ethn: the `on: :update` here and in the next event make a big difference.  `changed: :type_id` doesn't exclude all create events (where the type_id is changed from nothing to something, I suppose), so the rule cache was getting cleared on ALL create events.

card/mod/01_core/set/all/content.rb
```diff
@@ -34,10 +34,6 @@ def clean_html?
   true
 end
 
-def history?
-  false
-end
```
xithan: Why did you remove it from the core mod? We call that method in the standard and machine mod. If someone doesn't want to use the history mod he wiil get errors.

PR #550, opened 2015-12-15
Fix file upload as a field

chuenlok:


ethn:
we need a solution in the model, not the format.  the model needs to be smart enough not to update files/images with empty content when no new file/image is chosen.

card/mod/05_standard/set/abstract/attachment.rb
```diff
@@ -202,7 +207,11 @@ def mod_file?
 
 
 def assign_set_specific_attributes
-  if @set_specific && @set_specific.present?
+  # if the attachment is a empty string, just keep the content
+  attachment_name_str = attachment_name.to_s
+  if @set_specific && @set_specific.present? &&
```
chuenlok: here is the evil place that makes the db_content to be `nil`. I am not sure this is a real "fix". Please comment.

wagn/features/support/env.rb
```diff
@@ -20,7 +20,7 @@
 # prefer to use XPath just remove this line and adjust any selectors in your
 # steps to use the XPath syntax.
 Capybara.default_selector = :css
-
+Capybara.default_wait_time = 60
```
chuenlok: the default value is 2s which may not be enough for some machines(like mine).

card/mod/05_standard/set/abstract/attachment.rb
```diff
@@ -56,7 +56,7 @@ def assign_attachment file, original_filename
 end
 
 def file_ready_to_save?
-  attachment.file &&
+  attachment.file.present? &&
```
chuenlok: `attachment.file` always exists. The `present?` is overridden to check if any file inside the `file`.

card/mod/05_standard/spec/set/type/email_template_spec.rb
```diff
@@ -87,120 +86,127 @@ def create_field name, args={}
...
+    context 'with context card' do
+      let(:context_card) do
+        file = File.new(File.join FIXTURES_PATH, 'mao2.jpg')
+        Card.create(
+          name:    'Banana',
+          content: 'data content [[A]]',
+          subcards: {
+            '+email' => { content: 'gary@gary.com' },
+            '+subject' => { type: 'Pointer', content: '[[default subject]]' },
+            '+attachment' => { type: 'File', file: file }
```
chuenlok: should create a attachment with a file instead of plain content

card/mod/05_standard/set/abstract/attachment.rb
ethn: first off, just `@set_specific.present?` should suffice on this line.

PR #552, opened 2015-12-15
improve name variants

xithan:


card/mod/02_basic_types/set/all/base.rb
```diff
@@ -10,13 +10,23 @@ def show view, args
...
+        name.titleize
       else
-        name
+        if ::Set.new([
+          :downcase, :upcase, :swapcase, :reverse, :succ
+        ]).include?(variant.to_sym)
+          name.send variant
+        else
+          name
+        end
```
ethn: the logic is good.  the code complexity...definitely good enough for a ski vacation :)

wondering if rubocop/hound/codeclimate can be configured to treat view and event block complexity the same way it treats method complexity
?

PR #554, opened 2015-12-16
config fixups

ethn:
right now, performance_logger configs aren't working for me in production. I think this will fix that.

### DIFF
--- a/wagn/lib/wagn/application.rb
+++ b/wagn/lib/wagn/application.rb
@@ -9,19 +9,18 @@ if defined?(Bundler)
   # Bundler.require(:default, :assets, Rails.env)
 end
 
-
 module Wagn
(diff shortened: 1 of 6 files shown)


wagn/lib/wagn/application.rb
```diff
@@ -31,37 +30,37 @@ def inherited(base)
...
-        config.no_authentication     = false
-        config.files_web_path        = 'files'
-
-        config.email_defaults        = nil
-
-        config.token_expiry          = 2.days
-        config.revisions_per_page    = 10
-        config.request_logger        = false
-        config.performance_logger    = false
-        config.sql_comments          = true
```
ethn: main changes is moving these things to cardio.rb

PR #555, opened 2015-12-21
tweak pointer API

xithan:
added
- phase methods "added_items", "dropped_items", and "changed_items"
- spec helper method to test in event phase


card/lib/card/spec_helper.rb
```diff
@@ -76,4 +73,45 @@ def render_card_with_args view, card_args={}, format_args={}, view_args={}
...
+      define_method :in_phase_test, event_block
+    end
+    Card.define_callbacks :in_phase_test
+    kind =  ([:before, :after, :around] & opts.keys).first
+    name = opts.delete(kind)
+    Card.set_callback name, kind, :in_phase_test, prepend: true
+    opts[:trigger].call
+  ensure
+    Card.skip_callback name, kind, :in_phase_test
+  end
```
xithan: above is the meat: the `in_phase` method that you can use in rspec tests to check behaviour in events.

card/lib/card/spec_helper.rb
```diff
@@ -76,4 +73,45 @@ def render_card_with_args view, card_args={}, format_args={}, view_args={}
...
+        rescue NameError
+        else
+          return method.call(*args)
+        end
+        begin
+          value = eval(m.to_s, $rspec_binding)
+          return value
+        rescue NameError
+        end
+        super
```
xithan: The tricky part to get the right context - either Card or Rspec so that you can use card methods as in usual events but also rspec methods like `expect`.
I couldn't get around the global variable `$rspec_binding` to pass the test's binding into the block. Suggestions for a better solution are welcome.

PR #557, opened 2016-01-04
API to allow empty file and image cards

xithan:


ethn:
looks good.  I'm liking "set_specific_attributes"!

PR #558, opened 2016-01-05
fix submit buttons

xithan:
make disable_with a data attribute

### DIFF
--- a/card/mod/01_core/set/all/type.rb
+++ b/card/mod/01_core/set/all/type.rb
@@ -18,6 +18,10 @@ def type_name
   type_card.try :name
 end
 
+def type_name_or_default
+  type_card.try(:name) || Card.quick_fetch(Card.default_type_id).name
+end
+
 def type_cardname
   type_card.try :cardname
 end
--- a/card/mod/05_standard/set/abstract/attachment.rb
+++ b/card/mod/05_standard/set/abstract/attachment.rb
@@ -117,7 +117,7 @@ def original_filename
 end
 
 def unfilled?
-  !attachment.present? && !save_preliminary_upload? && super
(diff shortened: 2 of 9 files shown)


card/mod/05_standard/set/abstract/attachment.rb
```diff
@@ -117,7 +117,7 @@ def original_filename
 end
 
 def unfilled?
-  !attachment.present? && !save_preliminary_upload? && super
+  !attachment.present? && !save_preliminary_upload? && !subcards.present?
```
xithan: change 1 of 2: consider attachment cards with content but without attachment as empty

card/mod/05_standard/set/all/rich_html/form.rb
```diff
@@ -212,31 +233,52 @@ def process_relative_tags args
...
     form.radio_button method, tag_value, options
   end
 
+  def submit_button args={}
+    args.reverse_merge!(
+      situation: 'primary',
+      data: {}
+    )
+    text = args.delete(:text) || 'Submit'
+    args[:data][:disable_with] ||= args.delete(:disable_with) || 'Submitting'
```
xithan: change 2 of 2: make :disable_with a data argument

card/mod/01_core/set/all/type.rb
```diff
@@ -18,6 +18,10 @@ def type_name
   type_card.try :name
 end
 
+def type_name_or_default
+  type_card.try(:name) || Card.quick_fetch(Card.default_type_id).name
```
ethn: fwiw, could do `type_name ||`

card/mod/05_standard/set/abstract/attachment.rb
ethn: makes sense.

PR #559, opened 2016-01-05
first go at reference improvements

ethn:
mostly just code cleanup, but I think it may improve performance, too.

### DIFF
--- a/card/lib/card/content.rb
+++ b/card/lib/card/content.rb
@@ -59,7 +59,7 @@ class Card
       each_chunk.select { |chunk| chunk.is_a?(chunk_type) }
     end
 
-    def process_content_object &block
+    def process_each_chunk &block
       each_chunk { |chunk| chunk.process_chunk &block }
       self
     end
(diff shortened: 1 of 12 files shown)


card/mod/01_core/set/all/references.rb
```diff
@@ -1,115 +1,103 @@
...
+  Card.connection.execute(
+    "update cards set references_expired=NULL where id=#{id}"
+  )
   expire if refresh
 
-  rendered_content ||= Card::Content.new(raw_content, card=self)
+  rendered_content ||= Card::Content.new raw_content, self
   rendered_content.find_chunks(Card::Chunk::Reference).each do |chunk|
-    if referee_name = chunk.referee_name # name is referenced (not true of commented inclusions)
-      referee_id = chunk.referee_id
```
ethn: Chunk#referee_id was doing a full card fetch; all we need is the id.  I think this may have been slowing things down a bunch?

card/lib/card/content.rb
```diff
@@ -59,7 +59,7 @@ def find_chunks chunk_type
       each_chunk.select { |chunk| chunk.is_a?(chunk_type) }
```
xithan: while you are working on this can you remove the non-existent `msg` variable in line 53?

ethn: yes!

PR #563, opened 2016-01-08
reset @current_view after render

xithan:


card/lib/card/format.rb
```diff
@@ -292,6 +293,13 @@ def rescue_view e, view
       end
     end
 
+    def current_view view
```
ethn: ?  I see `view` below.  Was this fixed in a later commit?

card/lib/generators/card/migration/templates/card_migration.erb
```diff
@@ -3,8 +3,7 @@
 class <%= migration_class_name %> < <%= @migration_parent_class %>
   def up
 <% if @migration_action == 'import' -%>
-    import_json "<%= @json_filename %>"
+    import_json '<%= @json_filename %>'
```
ethn: :)

ethn:
looks good!

card/lib/card/format.rb
ethn: (nevermind, I saw the commit history)

PR #565, opened 2016-01-15
refactor cache

xithan:
* fix subcard/card_accessor issue
* don’t preselect current_set in rules editor

### DIFF
--- a/card/lib/card/cache.rb
+++ b/card/lib/card/cache.rb
@@ -120,7 +120,7 @@ class Card
 
     def read key
       @soft.read(key) ||
-        (@hard && (ret = @hard.read(key))  && @soft.write(key, ret))
+        (@hard && (ret = @hard.read(key)) && @soft.write(key, ret))
     end
 
     def write key, value
@@ -148,7 +148,7 @@ class Card
       @soft.dump
     end
 
-    def reset hard=false
+    def reset
       @hard.reset if @hard
       @soft.reset
     end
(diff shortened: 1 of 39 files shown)


card/lib/card/cache.rb
```diff
@@ -36,41 +36,44 @@ def renew
...
-      def reset_local
-        cache_by_class.each do |cc, cache|
-          if Card::Cache === cache
-            cache.reset_local
+      def reset_hard
+        cache_by_class.each do |_klass, cache|
+          if cache.is_a? Card::Cache
+            cache.hard.reset
+          else
+            warn "reset class #{cc}, #{cache.class} #{caller[0..8] * "\n"} ???"
```
ethn: do you ever see these warnings?  we might be able to get rid of this test.  (ancient debugging?)

xithan: never seen it. I remove it

card/lib/card/cache.rb
```diff
@@ -96,127 +99,70 @@ def obj_to_key obj
...
-      @local[key]
-    end
-
-    def write_variable key, variable, value
-      key = @prefix + key
-      if @store && (object = @store.read key)
-        object.instance_variable_set "@#{ variable }", value
-        @store.write key, object
-      end
-      value
```
ethn: I keep thinking we need something like `write_variable`, but I'm guessing we're not using it yet?

ethn: oh! nevermind; it's in the other class.

PR #566, opened 2016-01-16
factoring out reference expirations

ethn:


ethn:
this is not really ready -- need faster mechanism for updating references when structure changes.

PR #570, opened 2016-01-20
fix fetch virtual card with new args problem

chuenlok:


card/mod/01_core/spec/set/all/fetch_spec.rb
```diff
@@ -218,16 +218,31 @@
...
+                                         supercard: Card['home'] }
+
+        c = Card.fetch('Home+testsearch'.to_name)
+        assert c.virtual?
+        expect(c.type_code).to eq(:search_type)
+        expect(c.raw_content).to eq('{"plus":"_self"}')
+        patterns = c.instance_variable_get('@patterns').map(&:to_s)
+        expect(patterns).to include('Search+*type')
+      end
+    end
```
ethn: this fails without the changes to fetch.rb, right?

PR #580, opened 2016-02-17
get rid of F's in codeclimate

xithan:


xithan:
looks like 268 lines are too many for a class to get a D

PR #583, opened 2016-02-18
fix subcard issues

chuenlok:
renaming problems and keep name as blank if adding subcard with blank name

### DIFF
--- a/card/lib/card/subcards.rb
+++ b/card/lib/card/subcards.rb
@@ -99,7 +99,7 @@ class Card
       end
     end
 
-    def rename old_name, _new_name
+    def rename old_name, new_name
       return unless @keys.include? old_name.to_name.key
       @keys.delete old_name.to_name.key
       @keys << new_name.to_name.key
@@ -228,7 +228,7 @@ class Card
     end
 
     def absolutize_subcard_name name
-      if @context_card.name =~ /^\+/
+      if @context_card.name =~ /^\+/ || name.blank?
         name.to_name
       else
         name.to_name.to_absolute_name(@context_card.name)
--- a/card/mod/01_core/set/all/name.rb
+++ b/card/mod/01_core/set/all/name.rb
@@ -31,10 +31,14 @@ def name= newname
 
   newkey = cardname.key
   if key != newkey
+    was_in_cache = Card.cache.soft.delete key
     self.key = newkey
+    # keep the soft cache up-to-date
+    Card.write_to_soft_cache self if was_in_cache
     # reset the old name - should be handled in tracked_attributes!!
     reset_patterns_if_rule
     reset_patterns
+
   end
   if @subcards
     subcards.each do |subcard|
@@ -54,9 +58,7 @@ def name= newname
       subcard.name = subcard.cardname.replace_part name_to_replace, newname
     end
   end
-
   write_attribute :name, cardname.s
-  Card.write_to_soft_cache self
 end
 
 def cardname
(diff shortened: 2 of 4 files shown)


card/lib/card/subcards.rb
```diff
@@ -228,7 +228,7 @@ def new_by_attributes name, attributes={}
     end
 
     def absolutize_subcard_name name
-      if @context_card.name =~ /^\+/
+      if @context_card.name =~ /^\+/ || name.blank?
```
ethn: when does this happen?  seems like we need some inline comments here.

card/lib/card/subcards.rb
```diff
@@ -99,7 +99,7 @@ def catch_up_to_stage stage_index
       end
     end
 
-    def rename old_name, _new_name
+    def rename old_name, new_name
```
ethn: doh!

card/mod/01_core/set/all/name.rb
```diff
@@ -31,10 +31,14 @@ def name= newname
 
   newkey = cardname.key
   if key != newkey
+    was_in_cache = Card.cache.soft.delete key
     self.key = newkey
+    # keep the soft cache up-to-date
+    Card.write_to_soft_cache self if was_in_cache
```
ethn: I'd like to see this handled in `#key=`.  Something like:

````
def key= newkey
  was_in_cache = Card.cache.soft.delete key
  write_attribute :key, newkey
  # keep the soft cache up-to-date
  Card.write_to_soft_cache self if was_in_cache
  newkey
end
````

This way the cache is handled properly if the key is set directly (though that's generally a bad idea) and `name=` is less crazy.

File: card/spec/lib/card/stage_director_spec.rb [hunk 4]
```diff
@@ -219,7 +219,7 @@
   describe 'subcards' do
     it "has correct name if supercard's name get changed" do
       Card::Auth.as_bot do
-        $changed = false
+        changed = false
```
ethn: where did the perl come from? :)

ethn:
a few superficial comments, but this looks good!

card/lib/card/subcards.rb
xithan: If you want to use a autoname rule for a subcard. For example in wikirate we add a source card as follows:
`add_subcard '', type_id: SourceID`.  Without the fix above the source card gets the same name as its supercard and the autoname rule is not used because the name is not empty. I think we should fix that in smartname. Or do you have reasion why an empty name should absolutize to the context name?

card/mod/01_core/set/all/name.rb
xithan: good point (developed right in the battle against hordes of failing wikirate tests - no time for pretty solutions ;)

card/spec/lib/card/stage_director_spec.rb
xithan: Is used in a completely different context (in a event on a card object but with the rspec binding as fallback for missing methods). I wrote the test to fix a bug and I wanted to make sure that it doesn't fail because of this variable.

card/lib/card/subcards.rb
ethn: Ah, ok.  I can't think of a case other than autoname where the name should be empty by here.  (I've been a little confused all along about how exactly to handle autoname, but this seems like it fits our current approach)

ethn:
merged this into master...

PR #584, opened 2016-02-19
trying life without set member caching

ethn:
New Relic tells us memcached set is eating up TONS of time.  A quick look at the logs made it clear that a TON of cache writes have to do with set_members.

This is a first step in an attempt to obviate many of those calls.


ethn:
life is good.

(this is in the wikirate branch now.  closing.)

PR #585, opened 2016-02-19
fix card renaming issues

chuenlok:
renaming card won't update the embed content in related cards


card/mod/03_machines/lib/javascript/wagn.js.coffee
```diff
@@ -232,7 +232,7 @@ $(window).ready ->
     $(this).closest('form').submit()
 
   $('body').on 'click', '.renamer-updater', ->
-    $(this).closest('form').find('#card_update_referencers').val 'true'
+    $(this).closest('form').find('#card_update_referers').val 'true'
```
ethn: good catch!  I must have done all my substitutions on ruby files only :(

PR #596, opened 2016-02-24
Cachestamp

ethn:


card/lib/card/cache.rb
```diff
@@ -29,21 +18,10 @@ def [] klass
       end
 
       def renew
-        cache_by_class.keys do |klass|
-          raise "renewing nil cache: #{klass}" unless klass.cache
-          cache_by_class[klass].system_prefix = system_prefix(klass)
+        cache_by_class.each do |_klass, cache|
+          cache.soft.reset
+          cache.hard.renew if cache.hard
```
ethn: make sure we've got the latest stamp at the beginning of every request

card/lib/card/cache.rb
```diff
@@ -51,16 +29,18 @@ def restore
         prepopulate
       end
 
-      def reset_global # deprecated
-        reset_all
+      def reset_global
+        cache_by_class.each do |_klass, cache|
+          cache.soft.reset
+          cache.hard.annihilate if cache.hard
```
ethn: "annihilate" is the new word for clearing the entire cache (and that of every site using the service).

This means reset_global and reset_all are now different.

card/lib/card/cache/persistent.rb
```diff
@@ -1,33 +1,67 @@
...
+
+  def reset
+    @stamp = new_stamp
+    @prefix = nil
+    Cardio.cache.write stamp_key, @stamp
+  end
+
+  def stamp
+    @stamp ||= Cardio.cache.fetch stamp_key { new_stamp }
+  end
```
ethn: note that each class cached has its own stamp and can be independently cleared.

PR #597, opened 2016-02-26
Recentacts

ethn:
This is a start to a new act-based handling of recent changes.  

Other things we *ought to have*:
1. better filtering of cards we don't care about
2. grouping multiple acts on the same card(?)
3. interface touchups

### DIFF
--- a/card/mod/01_core/spec/set/all/content_spec.rb
+++ b/card/mod/01_core/spec/set/all/content_spec.rb
@@ -9,7 +9,7 @@ describe Card::Set::All::Content do
       @card.save_content_draft('booboo')
       @card.reload
       expect(@card.drafts.length).to eq 1
-      expect(@card.drafts[0].new_values[:content]).to eq 'booboo'
+      expect(@card.drafts[0].value(:content)).to eq 'booboo'
     end
   end
 end
--- a/card/mod/01_core/spec/set/all/content_spec.rb
+++ b/card/mod/01_core/spec/set/all/content_spec.rb
@@ -9,7 +9,7 @@ describe Card::Set::All::Content do
       @card.save_content_draft('booboo')
       @card.reload
       expect(@card.drafts.length).to eq 1
-      expect(@card.drafts[0].value(:content)).to eq 'booboo'
+      expect(@card.drafts[0].value(:db_content)).to eq 'booboo'
     end
   end
 end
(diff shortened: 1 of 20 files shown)


card/mod/01_core/spec/set/all/content_spec.rb
```diff
@@ -9,7 +9,7 @@
       @card.save_content_draft('booboo')
       @card.reload
       expect(@card.drafts.length).to eq 1
-      expect(@card.drafts[0].new_values[:content]).to eq 'booboo'
+      expect(@card.drafts[0].value(:db_content)).to eq 'booboo'
```
ethn: also supports `value(:content)`

card/mod/01_history/lib/card/action.rb
```diff
@@ -61,47 +69,46 @@ def delete_old
...
+    def changes
+      @changes ||=
+        begin
+          hash = {}
+          card_changes.each do |change|
+            hash[change.field.to_sym] = change
+          end
+          hash
+        end
     end
```
ethn: the core of the new approach is that the caching is more on-demand, and further behind the API

card/mod/01_history/lib/card/action.rb
```diff
@@ -112,25 +119,35 @@ def field_index field
...
 
     def new_content?
-      new_value_for :db_content
+      value :db_content
     end
 
     def new_name?
-      new_value_for :name
+      value :name
     end
```
ethn: I've been working out this "new" vocab.  I understand where it came from, but it's an easier translation to the database if we refer to the "new" value as just the "value" of the change.  (it's the value field of the change row, after all")

card/mod/01_history/lib/card/change.rb
```diff
@@ -31,7 +31,7 @@ def field= value
     end
 
     def field
-      TRACKED_FIELDS[read_attribute(:field)]
+      TRACKED_FIELDS[ attribute(:field)]
```
ethn: this was not intentional, but I guess it still works??

ethn:
I added code to address the needs for handling permissions and bad data that I mentioned before.  I'd still rather ensure we're not getting bad data, but for now those checks are in place.

card/mod/01_history/lib/card/action.rb
xithan: just a style remark: you can save five lines if you use `each_with_object`

PR #598, opened 2016-02-27
spec to check order of stages for a junction card

xithan:


xithan:
This PR with only test code is dedicated to the developers among us who tend to skip reading test code to give them a chance to appreciate my beautiful event test API ;)

PR #599, opened 2016-03-01
action.rb: tidy up

xithan:


card/mod/01_history/lib/card/action.rb
```diff
@@ -158,10 +117,6 @@ def action_type
       TYPE[read_attribute(:action_type)]
     end
 
-    def set_act
-      self.set_act ||= acts.last
-    end
```
xithan: don't know what this was for but I don't see it anywhere used.

card/mod/01_history/lib/card/action.rb
```diff
@@ -101,43 +97,6 @@ def previous_value field
...
-      field = interpret_field field
-      if @previous_changes && @previous_changes.key?(field)
-        @previous_changes[field]
-      else
-        @previous_changes ||= {}
-        @previous_changes[field] = card.last_change_on field, before: self
-      end
-    end
-
-    def field_index field
```
xithan: I removed this one and made the other methods in this section private

PR #600, opened 2016-03-02
mix o fixes

ethn:


card/lib/card/cache/persistent.rb
```diff
@@ -13,7 +13,8 @@ def database_name
 
   def initialize opts
     @store = opts[:store]
-    @klass = opts[:klass]
+    @klass = opts[:class]
```
ethn: oops.

card/lib/card/query/sql_statement.rb
```diff
@@ -206,7 +206,7 @@ def sort_field key, as, dir
           when 'id'             then "#{table}.id"
           when 'update'         then "#{table}.updated_at"
           when 'create'         then "#{table}.created_at"
-          when /^(name|alpha)$/ then "LOWER( #{table}.key )"
+          when /^(name|alpha)$/ then "#{table}.key"
```
ethn: key is already lower, and I think "LOWER" may prevent SQL from using the index.

card/lib/cardio.rb
```diff
@@ -29,7 +29,6 @@ def default_configs
         recaptcha_private_key:  nil,
         recaptcha_proxy:        nil,
 
-        cache_store:            [:file_store, 'tmp/cache'],
```
ethn: go back to memory default.

card/mod/05_standard/set/self/recent.rb
```diff
@@ -15,8 +15,13 @@
...
     acts = Act.all_viewable.order(id: :desc).page(page).per(ACTS_PER_PAGE)
     acts.map do |act|
-      format = act.card.format :html
-      format.render_act args.merge(act: act, act_context: :absolute)
+      if (act_card = act.card)
+        format = act_card.format :html
+        format.render_act args.merge(act: act, act_context: :absolute)
+      else
+        Rails.logger.info "bad data, act: #{act}"
+        ''
```
ethn: hack for handling bad data.

card/mod/05_standard/set/self/recent.rb
```diff
@@ -1,4 +1,4 @@
-ACTS_PER_PAGE = 10
+ACTS_PER_PAGE = 50
```
ethn: we actually don't have paging at present, but most of the time in the page is taken up with the initial "all_viewable" query.  (really it's because of the order by)

PR #601, opened 2016-03-02
tweak API for tabs 

xithan:
also 
- phase API fix 
- added `wagn:migration:up VERSION=` to run a card migration again
- let `Card#field` fetch virtual cards 
- edit structure: a way to define a list of fields/cards that will be rendered in edit mode


ethn:
shall I wait for this to be mergeable before I review?

card/lib/card/format/nest.rb
```diff
@@ -12,6 +12,11 @@ def subformat subcard
                        mainline: @mainline, form: @form
       end
 
+      def field_subformat field
+        field = card.cardname.field(field) unless field.is_a?(Card)
+        subformat field
+      end
```
xithan: Are there better existing options to render nests for fields in a view?
With this I can do something like `field_subformat('about')._render_content` to get the same result that `{{+about}}` would produce

card/mod/01_core/set/all/references.rb
```diff
@@ -65,9 +65,8 @@ def update_references_out
 # insert entries in reference table
 def create_references_out
   ref_hash = {}
-  content_obj = Card::Content.new raw_content, self
-  content_obj.find_chunks(Card::Content::Chunk::Reference).each do |chunk|
-    interpret_reference ref_hash, chunk.referee_name, chunk.reference_code
+  each_reference_out do |referee_name, ref_type|
+    interpret_reference ref_hash, referee_name, ref_type
   end
```
xithan: extracted this so that it's easier to override what a reference is in a certain set

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -163,34 +175,23 @@ namespace :wagn do
...
-      end
+    desc 'Runs the "up" for a given deck cards migration VERSION.'
+    task :up => :environment do
+      version = ENV['VERSION'] ? ENV['VERSION'].to_i : nil
+      verbose = ENV['VERBOSE'] ? ENV['VERBOSE'] == 'true' : true
+      raise 'VERSION is required' unless version
+      ActiveRecord::Migration.verbose = verbose
+      ActiveRecord::Migrator.run :up, Cardio.migration_paths(:deck_cards),
+                                 version
     end
```
xithan: that task is new, everything else in that file is refactoring

card/lib/card/set.rb
```diff
@@ -169,22 +172,42 @@ def stage_method method, opts={}, &block
...
+    # @option opts [Symbol, Array<Symbol>] :except don't include these formats
+    # @example
+    # include_set Type::Basic, except: :css
+    def include_set set, opts={}
+      set_type = set.abstract_set? ? :abstract : :nonbase
+      @@modules[set_type][set.shortname].each do |set_mod|
+        include set_mod
+      end
+      include_set_formats set, opts
     end
```
ethn: very cool!!

PR #604, opened 2016-03-09
trait bug (to be solved)

xithan:
no time to look deeper into this but I added a test to identify the following bug:
if a trait's type is defined by a string (or symbol) instead of an id then `left` is nil

### DIFF
--- a/card/lib/card/set.rb
+++ b/card/lib/card/set.rb
@@ -3,6 +3,7 @@
 class Card
   module Set
     include Event
+    include Trait
     mattr_accessor :modules, :traits
     @@modules = { base: [], base_format: {}, nonbase: {}, nonbase_format: {} }
 
@@ -169,24 +170,6 @@ class Card
       end
     end
 
-    #
-    # ActiveCard support: accessing plus cards as attributes
-    #
-    def card_accessor *args
-      options = args.extract_options!
-      add_traits args, options.merge(reader: true, writer: true)
-    end
-
-    def card_reader *args
-      options = args.extract_options!
-      add_traits args, options.merge(reader: true)
-    end
-
-    def card_writer *args
-      options = args.extract_options!
-      add_traits args, options.merge(writer: true)
-    end
-
     def ensure_set &block
       set_module = yield
     rescue NameError => e
@@ -220,6 +203,7 @@ class Card
 
       # make the set available for use
       def register_set set_module
+        binding.pry if set_module.to_s == "Card::Set::Type::Optic"
         return if set_module.abstract_set?  # noop; only used by explicit
         # inclusion in other set modules
         if set_module.all_set?
@@ -283,6 +267,7 @@ EOF
 
       def clean_empty_module_from_hash hash
         hash.each do |mod_name, modlist|
+          binding.pry if mod_name == "Type::Optic"
           modlist.delete_if { |x| x.instance_methods.empty? }
           hash.delete mod_name if modlist.empty?
         end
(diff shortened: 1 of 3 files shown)


card/lib/card/set.rb
```diff
@@ -3,6 +3,7 @@
 class Card
   module Set
     include Event
+    include Trait
```
xithan: no changes here. I only moved trait stuff to a separate module

card/spec/lib/card/set/trait_spec.rb
```diff
@@ -0,0 +1,52 @@
+
+describe Card::Set::Trait do
+  class Card; module Set; module Type; module Phrase
+    extend Card::Set
+    card_accessor :write, type: 'pointer'
+    card_accessor :read, type_id: Card::PointerID
```
xithan: the `read` accessor works fine, the `write` accessor doesn't

card/spec/lib/card/set/trait_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  context 'if accessor type is defined by a string' do
+    it 'has left' do
+      in_stage :prepare_to_validate, on: :create,
+               trigger: -> { subject } do
+        # test API doesn't support sets for event
+        # so we check the name
+        return unless name == 'joke'
+        expect(write_card.left.class).to eq Card
```
xithan: fails because `write_card.left` returns nil

card/spec/lib/card/set/trait_spec.rb
```diff
@@ -0,0 +1,62 @@
+
+describe Card::Set::Trait do
+  class Card
+    module Set
+      module Type
+        module Phrase
+          extend Card::Set
+          card_accessor :write, type: :pointer
+          card_accessor :read, type_id: Card::PointerID
```
xithan: the read accessor works fine, the write accessor doesn't

card/spec/lib/card/set/trait_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  context 'if accessor type is defined by a string' do
+    it 'has left' do
+      in_stage :prepare_to_validate, on: :create,
+                                     trigger: -> { subject } do
+        # test API doesn't support sets for event
+        # so we check the name
+        return unless name == 'joke'
+        expect(write_card.left.class).to eq Card
```
xithan: fails because `write_card.left` returns nil

card/lib/card/set.rb
ethn: +1

card/spec/lib/card/set/trait_spec.rb
ethn: dumb question: did you try type_id?

ethn: dumb question #2.  This is not supposed to be card_reader and card_writer, right?  You actually want a card name named +*read and/or +*write?

dumb question #3.  Does the `*write` card exist?  Or another card with the codename "write"?  `*read` exists for permission rules, of course.  If it doesn't exist, Wagn should probably fail loudly...

ethn: dumb question #4: we're not building a new layer on the permissions system are we?

xithan: re Q2-4: It's only about testing `card_accessor`. I needed cards with codenames in the test db and we create`*write` with codename when we seed the test db. `read` was then the natural next candidate
re Q1: I can pass a symbol to type_id?

xithan: `card_accessor :write, type_id: :pointer` creates a basic card but at least `left` is working. To avoid a misunderstanding: this is not about getting that card_accessor working it's about using the `type` argument. We should either support that or at least give a sensible error message if somebody uses `type` instead of `type_id` (because we support it almost everywhere else). It took me a whole day to figure out that that was the reason while scored formulas were not working on wikirate.

ethn: ah, ok.  If you don't understand why :type doesn't work (I don't), let's at least make it fail loudly.

PR #605, opened 2016-03-26
Wikirate metrics

ethn:
just making this PR so it's easy to see and comment on changes.


card/lib/card/active_record_helper.rb
```diff
@@ -5,8 +5,9 @@ def create_card args
       Card.create! args
     end
 
+    # if card with same name exists move it out of the way
     def create_card! args
-      create_card args.reverse_merge(rename_if_conflict: :new)
+      create_card args.reverse_merge(rename_if_conflict: :old)
```
ethn: I think :new is probably a better default, because it puts the burden on the new migration to use codenames.

card/lib/card/core_ext.rb
```diff
@@ -67,6 +67,14 @@ def new_from_semicolon_attr_list attr_string
   end
 end
 
+class Array
+  def to_pointer_content
+    map do |item|
+      "[[#{item}]]"
+    end.join "\n"
+  end
+end
```
ethn: (not in love with this approach)

ethn: (but ok for the moment)

card/lib/card/format/render.rb
```diff
@@ -132,8 +133,14 @@ def add_class options, klass
         options[:class] = [options[:class], klass].flatten.compact * ' '
       end
 
+      def id_counter
+        return @parent.id_counter if @parent
+        @id_counter ||= 0
+        @id_counter += 1
+      end
```
ethn: these are instance methods, so I don't think they'll insure a unique id across the page.

we could make another module and extend format with that to make it a class method...

card/lib/card/migration.rb
```diff
@@ -44,7 +44,8 @@ def schema_mode mig_type=type
 
       ActiveRecord::Base.table_name_suffix = new_suffix
       ActiveRecord::SchemaMigration.reset_table_name
-      yield
+      paths = Cardio.migration_paths(type)
+      yield(paths)
```
ethn: were things totally broken without this?

card/lib/card/set.rb
```diff
@@ -3,6 +3,7 @@
 class Card
   module Set
     include Event
+    include Trait
```
ethn: yay!

card/lib/card/stage.rb
```diff
@@ -3,18 +3,43 @@ class Card
...
+  # add subcard                  yes! yes! yes  | yes  yes  yes |    yes
+  # remove subcard               yes! yes! yes  | yes  no   no! |    no!
+  # validate                     yes  yes  yes! |      no       |    no
+  # unsecure change              yes  yes! no   |      no!      |    no!
+  # secure change                     yes       | yes! no!  no! |    no!
+  # abort                             yes!      |      yes      |    yes?
+  # fail
+  # create other cards
+  # has id                            no        | no   no?  yes |    yes
+  # dirty attributes                  yes       |      yes      |    no
```
ethn: this is interesting stuff.  thanks for adding. Will want to ponder more deeply at some point...

card/mod/01_core/set/all/collection.rb
```diff
@@ -253,23 +253,19 @@ def nest_args args, chunk=nil
...
+      tabs[tab_name] = nest_path(name, nest_args).html_safe
+
+      active_name ||= tab_name
+      # warning: nest changes nest_args
+      active_content ||= nest(Card.fetch(name, new: {}), nest_args)
     end
-    tab_panel tab_buttons, tab_panes, args[:tab_type]
+    lazy_loading_tabs tabs, active_name, active_content,
+                      type: args[:tab_type]
   end
```
ethn: looks much improved.

card/mod/01_core/set/all/references.rb
```diff
@@ -121,6 +120,15 @@ def reference_values_array ref_hash
...
 end
 
+# invokes the given block for each reference in raw_content with
+# the reference name and reference type
+def each_reference_out
+  content_obj = Card::Content.new raw_content, self
+  content_obj.find_chunks(Card::Content::Chunk::Reference).each do |chunk|
+    yield(chunk.referee_name, chunk.reference_code)
+  end
+end
```
ethn: nice

card/mod/05_standard/set/all/rich_html/form.rb
```diff
@@ -28,6 +31,28 @@ def single_card_edit_slot args
...
+    fields.map do |field, opts|
+      nested_card =
+        if field.is_a?(Card)
+          field
+        else
+          fetch_nested_card inc_name: field
+        end
+      nest nested_card, opts
+    end.join "\n"
+  end
```
ethn: neat.

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -6,8 +6,20 @@ WAGN_SEED_PATH = File.join(
   ENV['DECKO_SEED_REPO_PATH'] || [Cardio.gem_root, 'db', 'seed'], 'new'
 )
 
+def run_card_migration core_or_deck
```
ethn: MUCH nicer!  we might even consider putting this inside Cardio or something similar...

card/lib/card/migration.rb
xithan: no, this is to avoid repetition. In all blocks we passed to that method we called `Cardio.migration_paths`

card/lib/card/format/render.rb
```diff
@@ -132,8 +133,14 @@ def add_class options, klass
...
+
       def unique_id
-        "#{card.key}-#{Time.now.to_i}-#{rand(3)}"
+        "#{card.key}-#{id_counter}"
```
xithan: what about  `"#{card.key}-#{object_id}-{id_counter}"`?

card/lib/card/active_record_helper.rb
xithan: hm, but it creates bad names. I changed it because I got something like "Formula 1" on Wikirate because there was a useless "formula" basic card in the db.
I think in a deck card migration it's likely that you want the card exactly as in your migration (there is already a ! in the method name as a warning) but in a core card migration you are maybe right that it's better to mess less with the deck cards of other people.

card/lib/card/format/render.rb
ethn: I like it.  (could skip second dash.  doesn't matter)

PR #615, opened 2016-04-22
simpler junction fetch

xithan:
changed fetch API such that it is easier to get a junction card if you have only the parts of the name.
For example
`Card.fetch BasicID, 'A', :type_plus_right` fetches 'Basic+A+*type plus right'

### DIFF
--- a/card/mod/01_core/set/all/fetch.rb
+++ b/card/mod/01_core/set/all/fetch.rb
@@ -15,6 +15,8 @@ module ClassMethods
   #    1. a numeric id (Integer)
   #    2. a name/key (String or Card::Name)
   #    3. a codename (Symbol)
+  #   or any combination of those. If you pass more then one mark they get
+  #   joined with a '+'
   # @param [Hash] opts ({})
   #   Options:
   #     :skip_virtual               Real cards only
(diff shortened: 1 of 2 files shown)


card/mod/01_core/set/all/fetch.rb
```diff
@@ -224,6 +225,19 @@ def write_to_soft_cache card
     Card.cache.soft.write "~#{card.id}", card.key if card.id && card.id != 0
   end
 
+  def compose_mark parts, opts
+    return normalize_mark(parts.first, opts) if parts.size == 1
+    parts.map do |p|
+      normalized = normalize_mark p, {}
+      normalized.is_a?(Integer) ? fetch(normalized).name : normalized.to_s
```
ethn: should use quick_fetch here.

card/mod/01_core/set/all/fetch.rb
```diff
@@ -224,6 +225,19 @@ def write_to_soft_cache card
     Card.cache.soft.write "~#{card.id}", card.key if card.id && card.id != 0
   end
 
+  def compose_mark parts, opts
```
ethn: this is cool.  I've wanted something like this.

PR #618, opened 2016-05-04
make default nest option configurable

xithan:
so that I can introduce formula nests in wikirate where
`{{metric | 2011}}` is interpreted as `{{metric | year: 2011}}`

### DIFF
--- a/card/mod/01_core/chunk/include.rb
+++ b/card/mod/01_core/chunk/include.rb
@@ -123,7 +123,7 @@ module Card::Content::Chunk
       return if attr_string.blank?
       attr_string.strip.split(';').each do |pair|
         value, key = pair.split(':').reverse
-        key ||= DEFAULT_OPTION.to_s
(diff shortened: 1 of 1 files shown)


card/mod/01_core/chunk/include.rb
```diff
@@ -122,7 +123,7 @@ def each_option attr_string
       return if attr_string.blank?
       attr_string.strip.split(';').each do |pair|
         value, key = pair.split(':').reverse
-        key ||= 'view'
+        key ||= self.class::DEFAULT_OPTION.to_s
```
ethn: Do we prefer this to, say, a class instance variable?

ethn: (regardless, feel free to merge)

xithan: It's not supposed to be changed (except by subclasses) so a constant is probably the right choice.

PR #619, opened 2016-05-05
better imports

xithan:
get card content for imports from separate files so that changes are reviewable in PRs.


card/lib/generators/card/migration/templates/card_migration.erb
```diff
@@ -3,7 +3,7 @@
 class <%= migration_class_name %> < <%= @migration_parent_class %>
   def up
 <% if @migration_action == 'import' -%>
-    import_json '<%= @json_filename %>'
+    import_cards '<%= @json_filename %>'
```
ethn: cool!

PR #626, opened 2016-05-20
replace tinyMCE with ProseMirror

xithan:
I backed ProseMirror a while ago on indiegogo (http://prosemirror.net/#project) and it seems to be in good shape now (it's still beta but at least I have a "prioritize your bug" perk in case we run into an issue :)

There are some style decisions to make if we switch to ProseMirror, for example whether we want to keep a toolbar or use ProseMirror's tooltip option.

I reorganized wagn_mod.js.coffee but I didn't remove the tinyMCE code in this PR


card/mod/03_machines/set/self/script_slot.rb
```diff
@@ -1,5 +1,7 @@
 include_set Abstract::CodeFile
 
 def source_files
-  %w( wagn_mod.js.coffee wagn.js.coffee )
+  %w( wagn_mod wagn_editor wagn_layout wagn_navbox wagn_upload wagn ).map do |n|
+    "#{n}.js.coffee"
+  end << 'script_prosemirror.js'
```
xithan: I'm not sure if we want prosemirror in here or add it with a migration as separate card to *all+*script. We have a separate card for ace editor code but wagn_editor.js.coffee doesn't work without the ace and prosemirror library.

chuenlok:
This supports markdown!!!!

wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -313,7 +313,7 @@ namespace :wagn do
                      end
           card.update_column :db_content,
                              card.attachment.db_content(mod: mod_name)
-          card.last_action.change_for(2).first
```
xithan: @ethn don't forget to check rake files when you make api changes

xithan:
I resubmit this so that all the merged transaction code disappears in the diff.

PR #627, opened 2016-05-21
Basic I18n setup

rhwentworth:
This pull request includes basic setup infrastructure for the internationalization of English text embedded in Wagn Ruby files.

The change...

1. Enables the use of Internationalization by adding card/config/initializers/internationalization.rb, which (1) tells the I18n gem where to find locale files for the card gem, and (2) sets up a development debug mechanism whereby if the environment variable WAGN_I18N_DEMARK is set, then text that was fetched via I18n.translate or I18n.localize (or aliases of these) will be visually demarcated to differentiate it from any text that is still embedded in a source file somewhere. This can be used by a developer to manually inspect a development website instance and notice what text is emerging from localization/translation and what text isn't.

2. Using the i18n gem, localizes date and time formatting, and extracts English text from a few Ruby files, to demonstrate how text internationalization works in association with Ruby files. (A subsequent pull request will offer much more of the same, as English text is extracted from a much larger number of source files.) A new locale file card/config/locales/en.yml holds the localized date/time formats and extracted English text.

3. Adds an RSpec spec, card/spec/lib/card/i18n_spec.rb, which uses the gem i18n-tasks to check for missing or unused translation keys, by parsing the Ruby source for I18n.translate calls and comparing this to the contents of en.yml (and any other locale files that might be added). Since i18-tasks can only be installed for Ruby 2.1 or greater, for lower Ruby versions the specs that require i18n-tasks will be skipped and reported as "Pending."

4. Adjusts the templates for generating new decks so that if testing is desired, then the i18n-tasks will be installed if Ruby version is >= 2.1; otherwise warnings will be offered but the user should be able to proceed to create and use the deck and run tests without errors.

Note that at this stage internationalization is essentially a refactoring rather than an addition of new functionality. Unless the debugging variable WAGN_I18N_DEMARK is set, there should be no changes to the functionality that web site users experience. This means that the functionality of these changes can largely be verified through the non-breakage of existing specs/tests rather than through the creation of new specs/tests. For example, there are existing specs for the classes that render dates/times, and these tests would fail if the new localized date/time formatting weren't functioning to reproduce the prior functionality. Proper behavior of RSpec specs has been verified, but I have not been able to verify the behavior of Cucumber tests due to the lack of a functional environment for running these.


rhwentworth:
Hmm... I'm puzzled by the stats apparently claiming some enormous numbers of commits and files are involved in this pull request. I expected only one commit affecting in the neighborhood of 10 files to be involved. Alas, no time to investigate this now.

xithan:
wrong branch. We usually use only the master branch. Looks like the develop branch is much behind master. If your make your pull request to master the commits flood will disappear.

PR #631, opened 2016-05-29
import rake tasks

xithan:
I moved WikiRate's import tasks to Wagn and improved them.
Check card.rake and the comments in import.rb for more details

The equivalent to WikiRate's 
`rake wikirate:import_from_dev name=import_migration_name`
`rake wagn:migrate`
with this new approach is:
`rake card:add_remote dev http://dev.wikirate.org` (only for the first time)
`rake card:pull_export from=dev` (pulls all items of the export card but not the export card itself; shortcut for `rake card:deep_pull_items export from=dev `)
`rake card:merge`


card/mod/03_machines/set/type/skin.rb
```diff
@@ -3,7 +3,4 @@
 
 store_machine_output filetype: 'css'
 
-include Pointer
-format      { include Pointer::Format     }
-format(:html) { include Pointer::HtmlFormat }
-format(:css) { include Pointer::CssFormat  }
+include_set Pointer
```
ethn: I know this isn't new, but I still love this.

card/lib/card/migration/import.rb
```diff
@@ -0,0 +1,211 @@
...
+        end
+
+        def content_path card_key
+          File.join CARD_CONTENT_DIR, card_key
+        end
+      end
+
+      # Handles the card attributes and remotes for the import
+      class MetaData
+        DEFAULT_PATH = Card::Migration.data_path('cards.yml').freeze
```
ethn: I _think_ this is right.  I'm getting myself in a little tangle about whether this should be stored in the database vs in a file.  Because, does it really describe the state of the database or the filesystem?  It already stores some migration data in the db, right?

xithan: It's like a migration file but it doesn't have anything to do with migrations. It controls the db changes of the `card:merge` rake task (similar as a migration file controls the db changes that `wagn:migrate` does)

xithan: Arguably the merge dates in `cards.yml` describe the state of the database.  But I doubt that moving that to the db is a good solution.

PR #632, opened 2016-06-03
fix sending extra emails to user because of renaming

chuenlok:
Same renaming notification emails are sent multiple times while some card contents' contain renamed card's reference.

I am not sure whether it should send one email saying name and contents of related card changed or send one email saying card name changed.

### DIFF
--- a/card/mod/05_email/set/all/notify.rb
+++ b/card/mod/05_email/set/all/notify.rb
@@ -68,7 +68,7 @@ def silent_change?
 end
 
 def notable_change?
-  !silent_change? && !supercard && current_act &&
+  !silent_change? && !supercard && current_act && current_act.card_id == id &&
     Card::Auth.current_id != WagnBotID && followable?
 end
 
--- a/card/mod/05_email/set/all/notify.rb
+++ b/card/mod/05_email/set/all/notify.rb
@@ -68,8 +68,12 @@ def silent_change?
 end
 
 def notable_change?
-  !silent_change? && !supercard && current_act && current_act.card_id == id &&
+  !silent_change? && current_act_card? &&
     Card::Auth.current_id != WagnBotID && followable?
+end
+
+def current_act_card?
+  current_act && current_act.card_id == id
 end
 
 event :notify_followers_after_save, :integrate,
--- a/card/mod/05_email/spec/set/all/notify_spec.rb
+++ b/card/mod/05_email/spec/set/all/notify_spec.rb
@@ -251,6 +251,10 @@ Use this link to unfollow /update/Joe_User+*follow?card%5Bsubcards%5D%5Banother+
       Card[card_name].update_attributes! content: new_content
     end
 
+    def update_name card_name, new_name='updated content'
+      Card[card_name].update_attributes! name: new_name, update_referers: true
+    end
+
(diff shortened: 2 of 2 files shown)


card/mod/05_email/set/all/notify.rb
```diff
@@ -68,7 +68,7 @@ def silent_change?
 end
 
 def notable_change?
-  !silent_change? && !supercard && current_act &&
+  !silent_change? && !supercard && current_act && current_act.card_id == id &&
```
ethn: If I understand this, you're saying that only changes to the current act card are notable.  I don't think that's what is intended at all.

card/mod/05_email/spec/set/all/notify_spec.rb
```diff
@@ -267,6 +271,16 @@ def update card_name, new_content='updated content'
...
     end
 
+    it 'sends only one notification per user'  do
+      Card.create! name: 'WOW', content: '{{Big Brother|link}}'
+      Card::Auth.as_bot do
+        Card.create! name: 'Users+*type+John+*follow', type_id: Card::PointerID,
+                     content: "[[*always]]\n"
+      end
+      expect_user('John').to receive(:send_change_notice).exactly(1)
+      update_name 'Big Brother'
```
ethn: does this fail without the change to notify.rb?

ethn:
What is triggering this problem?

@xithan can correct me if I'm wrong, but I thought that by design, we were only supposed to send one email per person per act.

So either (a) I'm wrong, (b) the above is broken, or (c) the renaming is creating more than one act. 

Can we diagnose that?

ethn:
To be clear, that one email might include lots of different changes.  That's why the proposed fix looks wrong to me.

card/mod/05_email/set/all/notify.rb
xithan: It's not really wrong. `!supercard` in the same line is supposed to do something similar (or the same) namely ensure that only one email per act is send. But it might be not the right way to do it.
Looks like the current behaviour is: We send an email to a user if the act card is followable and there was change to a card that is part of the act and followed by the user.

xithan:
Definitely not (a), probably (b)

card/mod/05_email/spec/set/all/notify_spec.rb
chuenlok: it fails without the change in `notify.rb`.

card/mod/05_email/set/all/notify.rb
chuenlok: I think Philipp is right. I thought the `!supercard` will block the sending but supercard was `nil` for the changed card that is part of the act( I am not sure because it was almost a month ago). I tried to looking to params I can use and then I though comparing the current act id and current card might be ok to stop that as there is a way to include all the changes in the email.

ethn: the suggestion is that the test be something like:

```
act_card? && !silent_change? && ...
```

...and that act_card look like your test:

```
def act_card?
  current_act && current_act.card_id == id
end
```
I'm a little concerned with names here and with how `followable?` works (what if main card isn't followable but subcards are?)

...but I can work on that later.

PR #634, opened 2016-07-25
transactions and subcards

xithan:
1. moves the intergration phase outside of the AR transaction
2. adds a possibility to process all stages of a subcard in one stage of its supercard
Expample `add_subcard 'name', content: 'hi', transact_in_stage: 'integrate'`
3. 1) and 2) together allow to process subcards in separate transactions
4. makes some attributes of Card::Env available in the integrate_with_delay stage
5. preserves dirty marks on card attributes in the integration phase


.ruby-style.yml
```diff
@@ -51,4 +51,6 @@ Style/ExtraSpacing:
   Enabled: false
 Lint/AssignmentInCondition:
   Enabled: true
-  AllowSafeAssignment: true
\ No newline at end of file
+  AllowSafeAssignment: true
+Style/Alias:
+  EnforcedStyle: prefer_alias_method
```
ethn: had to study to understand.  now I agree :)

card/config/locales/de.yml
```diff
@@ -23,7 +23,7 @@
 #    Format names for time and date are begun with a 'card_' prefix to
 #    minimize the chances of interference with other gems or applications.
 ---
-en:
```
ethn: aha!

card/lib/card.rb
```diff
@@ -55,7 +55,7 @@ def self.serializable_attr_accessor *args
     :silent_change,               # and this probably too
     :remove_rule_stash,
     :last_action_id_before_edit,
-    :skip_phases
+    :only_storage_phase
```
ethn: you didn't start the fire, but....

we should probably add comments for each of these.  

I'm also wondering whether this should really happen in card.rb.  I can't remember if I tried and failed to move it at some point...

card/lib/card.rb
```diff
@@ -73,11 +73,12 @@ def self.serializable_attr_accessor *args
...
-  before_validation :validation_phase, if: -> { run_phases? }
+  # Validation and integration phase are only called for the act card
+  # The act card starts those phases for all its subcards
+  before_validation :validation_phase, unless: -> { only_storage_phase? }
   around_save :storage_phase
-  after_save :integration_phase, if: -> { run_phases? }
-  after_commit :clean_up, if: -> { run_phases? }
-  after_rollback :clean_up, if: -> { run_phases? }
+  after_commit :integration_phase, unless: -> { only_storage_phase? }
+  after_rollback :clean_up, unless: -> { only_storage_phase? }
```
ethn: no cleanup after commit now?

card/lib/card/director_register.rb
```diff
@@ -50,7 +50,11 @@ def clear
       end
 
       def fetch card, opts={}
-        directors[card] ||= Card.new_director card, opts
+        return directors[card] if directors[card]
+        directors.each_key do |dir_card|
+          return dir_card.director if dir_card.name == card.name
+        end
```
ethn: Can this not be simplified if we use the card key as the hash key?

card/lib/card/env.rb
```diff
@@ -6,19 +6,17 @@ module Env
...
+        self[:session]    = c.request.session
+        self[:params]     = c.params
+        self[:ip]         = c.request.remote_ip
+        self[:ajax]       = c.request.xhr? || c.request.params[:simulate_xhr]
+        self[:html]       = [nil, 'html'].member?(c.params[:format])
+        self[:host]       = Card.config.override_host ||
+                            c.request.env['HTTP_HOST']
+        self[:protocol]   = Card.config.override_protocol ||
+                            c.request.protocol
       end
```
ethn: not important, but seems like we should return `self`

card/lib/card/env.rb
```diff
@@ -53,6 +51,17 @@ def html?
         !self[:controller] || self[:html]
       end
 
+      def serialize
+        keep = ::Set.new [:main_name, :params, :ip, :ajax, :html, :host,
```
ethn: should we make this a class instance variable or something?

card/lib/card/loader.rb
```diff
@@ -31,7 +31,7 @@ def load_chunks
       end
 
       def load_layouts
-        mod_dirs.inject({}) do |hash, mod|
+        mod_dirs.each_with_object({}) do |mod, hash|
```
ethn: does each_with_object reverse the argument order, or did we have it wrong?

card/lib/card/set/event.rb
```diff
@@ -23,53 +64,42 @@ def define_event event, opts, &final
...
-                  end
-              end
-            Object.const_get(event.to_s.camelize).perform_later(self, s_attr)
-          }
+          define_method(method_name, proc do
+            Object.const_get(event.to_s.camelize).perform_later(
+              self, serialize_for_active_job, Card::Env.serialize,
+              Card::Auth.current_id
+            )
+          end)
```
ethn: nice

card/mod/01_core/set/all/trash.rb
```diff
@@ -44,13 +44,13 @@ def delete!
...
     child.trash = true
     add_subcard child
-    next if child.valid?
-    child.errors.each do |field, message|
-      errors.add field, "can't delete #{child.name}: #{message}"
-    end
+    # next if child.valid?
+    # child.errors.each do |field, message|
+    #   errors.add field, "can't delete #{child.name}: #{message}"
+    # end
```
ethn: is this functionality replaced somehow?

card/mod/03_machines/lib/card/machine.rb
```diff
@@ -74,7 +74,7 @@ def self.define_machine_events host_class
       event_suffix = host_class.name.tr ':', '_'
       event_name = "reset_machine_output_#{event_suffix}".to_sym
       host_class.event event_name, after: :expire_related, on: :save do
-        reset_machine_output!
+        reset_machine_output
```
ethn: did delayed jobs break machines?

card/config/locales/de.yml
xithan: that's why master broke  after you merge both i18n PRs (#635)

card/lib/card.rb
xithan: it's part of the integration_phase now. Two after_commit callbacks don't work

wagn/features/update_includers.feature
```diff
@@ -8,6 +9,7 @@ Feature: Updates for Children of watched cards
   Background:
     Given I am signed in as Joe User
     And I create Book card "Ulysses"
+    And Jobs are dispatched
```
ethn: I'm super happy this is being tested.

I will make the pedantic comment that this would fail the "Vishal should understand it" test for whether the wording makes sense to show on a high level.  But I'm not sure how best to do this, unless we are actually going to translate waiting for jobs to finish into a specific user experience.

card/lib/card/director_register.rb
xithan: Maybe but it's very complicated to get it right. The key can change during the phases. Sometimes we start with an empty name, sometimes '+B' becomes 'A+B' or  'A' changes and so "A+B" has to change, too.  So you have to keep track of all these changes.

The reason why now the card object doesn't always work is that the integration phase is outside of the transaction. If we save a card in the integration phase that was part of the act then its card object has changed.

card/lib/card/loader.rb
xithan: as the name suggests: first the 'each' then the 'object' :)  inject arguments are reversed

card/mod/01_core/set/all/trash.rb
xithan: `child.valid?` means we mess with the order of the stages. child is added as subcard so it will be validated anyway and 'handle_subcard_errors' will take care of possible errors. We loose the "can't delete" in the error message at the moment. Not sure if the message will become less clear.

card/mod/03_machines/lib/card/machine.rb
xithan: yes, and thousand other things 

I removed the ! in the method name. I think the name makes clear what it does and there is nothing dangerous or unexpected.

wagn/features/update_includers.feature
xithan: I had a lot more of those but I included it in the "should be notified" step. 
The "Jobs are dispatche"  is needed here because  "And I create Book card "Ulysses"" adds notifications to the job queue and they stay there until we check later for notifications but then we don't expect those notifications.  
We could get rid of this by dispatching jobs if we make somebody watch something  (what we do in the next step)

PR #636, opened 2016-08-01
cache machine output parts

xithan:
cache all parts of a machine output file. If a input card was updated then update only the changed part of the machine output.


card/lib/card/format.rb
```diff
@@ -203,7 +201,9 @@ def process_content_object override_content=nil, opts={}
       content = override_content || render_raw || ''
       content_object = get_content_object content, opts
       content_object.process_each_chunk do |chunk_opts|
-        prepare_nest chunk_opts.merge(opts) { yield }
+        # Feels scary to just remove it but I can't make any sense of the
+        # "yield" and all tests pass without it
+        prepare_nest chunk_opts.merge(opts) #{ yield }
```
ethn: I think perhaps "process_content_object" used to take a block arg and I got rid of that?

ethn: (in other words, I think removing yield should be fine)

card/lib/card/format/nest/view.rb
```diff
@@ -0,0 +1,75 @@
...
+        # if nested in closed mode
+        def view_in_closed_mode homeview
+          approved_view = Card::Format.closed[homeview]
+          if approved_view == true then
+            homeview
+          elsif Card::Format.error_code[homeview] then
+            homeview
+          elsif approved_view then
+            approved_view
+          elsif !card.known? then
```
ethn: are you favoring #elsif over #when now?

ethn:
it looks like something in a script (perhaps the addition of the word "two"?) broke some query specs.

if/when you fix that, can you merge in the latest from master (if it's not already there) to see if we can get the diffs to update?

card/lib/card/format/nest/view.rb
xithan: The hound complained about the empty case statement and it made sense to me that you don't put conditions in a case block if they have nothing in common.

card/mod/03_machines/lib/card/machine.rb
```diff
@@ -151,12 +151,36 @@ def run_machine joint="\n"
...
               end
-      engine(input)
+      output = engine(input)
+      cache_output_part input_card, output
+      output
+    end
+
+    def fetch_cache_card input_card, new=nil
+      new &&= { type_id: PlainTextID }
+      Card.fetch input_card.name, :machine_cache, name, new: new
```
ethn: eg of new API idea,

this would probably be:
```
Card.fetch input_card.name, name solid_cache: { type_id: PlainTextID }
```
...and then you wouldn't need any of the save! handling.

card/mod/03_machines/lib/card/machine.rb
```diff
@@ -168,11 +192,23 @@ def reset_machine_output
...
+      end
+    end
+
+    def regenerate_machine_output
+      #lock do
+        run_machine
+      #end
+    end
+
+    def lock
```
ethn: like.

PR #644, opened 2016-08-10
admin tasks for solid and machine cache clearing

xithan:
* add link to `*stats` that deletes all `+*solid cache` cards
* add link to `*stats` that deletes all `+*machine cache` and `+*machine output` cards
* refactor stats.rb 
* rename `*stats` to `*admin`
* rename `+*stats` to `+*debug`
* show events tree and set modules in +*debug
* fix redirection to `*stats` for admin task
* javascript machine output: ~~add the card names of the input cards as javascript comments to the output~~ (correction: That was part of #636. This PR refactors only tests for that)
* split standard mod


card/mod/05_standard/set/self/all/tasks.rb
```diff
@@ -0,0 +1,17 @@
...
+  when :repair_references    then Card::Reference.repair_all
+  when :empty_trash          then Card.empty_trash
+  when :clear_view_cache     then Card::ViewCache.reset
+  when :delete_old_revisions then Card::Action.delete_old
+  when :repair_permissions   then Card.repair_all_permissions
+  when :clear_solid_cache    then Card.clear_solid_cache
+  when :clear_machine_cache  then Card.reset_all_machines
+  end
+  abort success: { card: Card[:stats] }
+end
```
xithan: Moved this from self/all.rb to self/all/tasks.rb so that the name of the file gives you at least a hint what the content is about.
I'm wondering why we update *all for these tasks. Can't we update *stats instead? 
That way the event would be in the same set as the interface for using it.

card/mod/05_standard/set/self/stats.rb
```diff
@@ -1,87 +1,85 @@
...
-def get_current_memory_usage
+def current_memory_usage
   `ps -o rss= -p #{Process.pid}`.to_i
 end
 
 def profile_memory &block
-  before = get_current_memory_usage
-  file, line, = caller[0].split(":")
+  before = current_memory_usage
+  _file, _line, = caller[0].split(":")
```
xithan: what mysterious side effects justify line 78?

card/mod/03_machines/set/all/reset_machines.rb
```diff
@@ -0,0 +1,14 @@
+module ClassMethods
+  def reset_all_machines
+    Auth.as_bot do
+      Card.search(right: { codename: "machine_cache" }).each do |card|
+        card.update_columns trash: true
+        card.expire
+      end
+      Card.search(right: { codename: "machine_output" }).each do |card|
```
ethn: not worth changing, but `{ codename: ["in", "machine_cache", "machine_output"] }` should also work.

card/mod/05_standard/set/self/all/tasks.rb
ethn: I think the rationale was that, eg wth repairing permissions or references, we were conceptually updating all cards (and were not updating the stats card).  Given how our events and subcards have evolved, it seems more normal now to do this as an update on the stats card (an admin card might make even more sense, but we'd have to migrate for that, and I'm not sure it's that important)

PR #647, opened 2016-08-12
introduce Modfiles

xithan:
a first small step towards gem based mods:
A "Modfile" defines which mods and in what order they are loaded.
If there is no Modfile the old method is used: ordered by directory names

The important changes are in card/loader.rb and card/loader/mod_dirs.rb. Everything else is renaming.


xithan:
Moving code around is very bad for the climate. Too many skeletons in the closet.

ethn:
:)

PR #650, opened 2016-08-16
new yard-friendly comments

ethn:
so far docs for act.rb. more coming.

using `yard server --reload` might help doc review.


xithan:
You should add all the changes in card/doc/** in a separate commit. Otherwise it's impossible to recognize and review what you changed.

ethn:
yeah. even that would mean you can't review the PR as a whole.

Some options:
1. separate repo/submodule. 
2. leave the docs out of git (forces others to generate them)
3. get rid of timestamps so that the only doc changes are _actual_ changes.

PR #651, opened 2016-08-16
buckets

xithan:
file storage options [as suggested on wagn.org](http://wagn.org/file_storage_options)


card/card.gemspec
```diff
@@ -40,7 +40,7 @@ Gem::Specification.new do |s|
     ["uuid",                       "~> 2.3"],
     # with carrierwave 0.11.1 and 0.11.2 wagn hangs in a loop on creating
     # style and script machine output
-    ["carrierwave",                "<= 0.11.0"],
+    #["carrierwave",                "<= 0.11.0"],
```
ethn: how come?

xithan: We can't use the gem. We have to use the master branch. The gem uses an older version of fog which isn't compatible with Rails 4.2

ethn: so, presumably development is still active and we'll be able to go back to the gem when it comes out?

card/mod/carrierwave/lib/carrier_wave/cardmount.rb
```diff
@@ -30,15 +32,15 @@ def mount_uploader column, uploader=nil, options={}, &block
...
         end
-        event :remove_previously_stored_#{column}_event, :store,
+        event :remove_previously_stored_#{column}_event, :finalize,
               on: :update, when: proc { |c| !c.history?} do
-          if @previous_model_for_#{column}
-            @previous_model_for_#{column}.include_set_modules
-          end
+          # if @previous_model_for_#{column}
+          #   @previous_model_for_#{column}.include_set_modules
+          # end
```
ethn: was this to gain access to delete events?

card/mod/carrierwave/set/abstract/attachment/storage_type.rb
```diff
@@ -0,0 +1,195 @@
...
+  return if db_content_changed? || !content.present?
+  case content
+  when %r{^:[^/]+/([^.]+)} then Regexp.last_match(1) # current mod_file format
+  when /^\~/               then false  # current id file format
+  else
+    if (lines = content.split("\n")) && (lines.size == 4)
+      # old format, still used in card_changes.
+      lines.last
+    end
+  end
```
ethn: might be nice to put everything past the first `when` in a `#deprecated_mod_file` method.

card/mod/carrierwave/set/all/file_utils.rb
```diff
@@ -0,0 +1,30 @@
...
+      # we don't want to delete uploads in progress
+      if older_than_five_days?(action.created_at) && (card = action.card)
+        # we don't want to delete uploads in progress
+        card.delete_files_for_action action
+        action.delete
+      end
+    end
+  end
+
+  def older_than_five_days? time
```
ethn: Maybe #too_old? and an overrideable default time?

card/card.gemspec
xithan: they promise a forthcoming version 1.0

card/mod/carrierwave/lib/carrier_wave/cardmount.rb
xithan: don't know why this was there but @previous_model doesn't exist anymore. Instead of the model they only store changed attributes.

card/mod/carrierwave/set/abstract/attachment/storage_type.rb
xithan: will do that when I replace the old mod file stuff with "storage_type = :coded" api

card/mod/carrierwave/set/all/file_utils.rb
xithan: I wrote that method instead of a comment about the 432000. Now I need the comment for the 432000 as default time :)

card/mod/carrierwave/Gemfile
```diff
@@ -0,0 +1,4 @@
+# the current release 0.11.2 uses an old version of the fog gem which isn't
+# compatible with rails 4.2
+gem "carrierwave", git: "https://github.com/carrierwaveuploader/carrierwave",
+                   branch: "master"
```
xithan: Noticed that this doesn't work. The Gemfile in a deck doesn't find Gemfiles in mods in the wagn gem. It only loads other Gemfiles in deck mods.

xithan: I forgot that I implemented an alternative load procedure for the gem. I added carrierwave as submodule and added the submodule directory to the load path. Then I can just require it:
```ruby
cw_path = File.expand_path "../../vendor/carrierwave/lib", __FILE__
$LOAD_PATH.unshift(cw_path) unless $LOAD_PATH.include?(cw_path)
require "carrierwave"
```

.gitignore
```diff
@@ -15,3 +16,5 @@ decko*gem
 *.DS_Store
 *.swp
 test_work/*
+*.orig
+card/spec/config/bucket_credentials.yml
```
ethn: safer to say something like **/bucket_credentials.yml?  or is there a template that this would cause problems for?

card/mod/admin/set/self/admin.rb
```diff
@@ -1,6 +1,10 @@
+# collect arrays of the form
+# [task symbol, { execute_policy: block, stats_policy: block }]
+basket :tasks
+
 event :admin_tasks, :initialize, on: :update do
```
ethn: it sure doesn't seem like rubocop is measuring view/event complexity.  that would probably knock down wikirate's scores quite a bit (and wagn some, too)

card/mod/admin/set/self/admin.rb
```diff
@@ -11,6 +15,10 @@
   when :repair_permissions   then Card.repair_all_permissions
   when :clear_solid_cache    then Card.clear_solid_cache
   when :clear_machine_cache  then Card.reset_all_machines
+  else
+    if (task_data = Hash[tasks][task])
+      task_data[:execute_policy].call
+    end
```
ethn: I like this.  wondering if we should register all the above tasks this way, too (at some point)

card/mod/carrierwave/Gemfile
ethn: so is this file not doing anything?

card/mod/carrierwave/lib/carrier_wave/cardmount.rb
```diff
@@ -30,15 +32,12 @@ def mount_uploader column, uploader=nil, options={}, &block
...
               on: :update, when: proc { |c| !c.history? } do
-          store_previous_model_for_#{column}
+          store_previous_changes_for_#{column}
         end
-        event :remove_previously_stored_#{column}_event, :store,
+        event :remove_previously_stored_#{column}_event, :finalize,
               on: :update, when: proc { |c| !c.history?} do
-          if @previous_model_for_#{column}
-            @previous_model_for_#{column}.include_set_modules
-          end
```
ethn: this wasn't needed?

card/mod/carrierwave/lib/carrier_wave/cardmount.rb
```diff
@@ -64,23 +63,38 @@ def write_uploader *args
...
         def remote_#{column}_url=(url)
-          send(:"#{column}_will_change!")
           db_column = _mounter(:#{column}).serialization_column
           send(:"\#{db_column}_will_change!")
+          if web?
+            self.content = url
+          else
+            send(:"#{column}_will_change!")
+            super
+          end
```
ethn: seems like this and above could potentially share a method, no?

card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
```diff
@@ -38,14 +16,139 @@ def full_filename for_file
...
+  # - :<code name>
+  #
+  # <revision> is the mod name if the file is coded or and action_id in any
+  # case
+  #
+  # Examples:
+  # *logo.png
+  # ~22/33-medium.png               # local
+  # :yeti_skin/standard-large.png   # coded
+  #
```
ethn: great stuff.

card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
```diff
@@ -38,14 +16,139 @@ def full_filename for_file
...
+  # Cards that support attachments (by default those are cards of type "file"
+  # and "image") accept a file handle as a card attribute.
+  #
+  # @example Attaching a file to a file card
+  #   Card.create name: "file card", type: :file,
+  #               file: File.new(path_to_file)
+  #
+  # @example Attaching a image to a image card
+  #   Card.create name: "file card", type: :image,
+  #               image: File.new(path_to_image)
```
ethn: with this clear articulation, I'm wondering for a moment whether this is the ideal API, or whether developers would find it easier to remember if it were always (`file:`)

(no need to change anything now, just a thought)

card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
```diff
@@ -62,43 +165,55 @@ def extension
...
+        else
+          "#{action_id}#{extension}"
+        end
+      end
     end
 
     def url opts={}
+      if model.cloud?
+        file.url
+      elsif model.web?
```
ethn: can't remember how case works.

could I do:

```
case model
  when cloud...
```
?

card/mod/carrierwave/Gemfile
xithan: yes, unless you add the carrierwave mod to your deck.

.gitignore
xithan: no, that should work

card/mod/carrierwave/spec/lib/carrier_wave/file_card_uploader_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  let(:local_file) { create_file_card :local }
+  let(:coded_file) { Card[:logo] }
+  let(:web_file) { create_file_card :web, "http://web.de/test.txt" }
+
+  describe "#db_content" do
+    context "coded file" do
+      subject { coded_file }
+      it "returns correct identifier" do
+        expect(subject.attachment.db_content)
```
ethn: something about this api seems funny to me:

`subject.attachment.db_content`

my instinct is that the card, not the uploader, should have db_content.  but I get what you're going for.  No change to recommend yet.

card/mod/carrierwave/spec/set/type/file_spec.rb
```diff
@@ -0,0 +1,463 @@
...
+        credentials: bucket_credentials(:aws),
+        subdirectory: "files",
+        directory: directory,
+        public: true,
+        attributes: { "Cache-Control" => "max-age=#{365.day.to_i}" },
+        authenticated_url_expiration: 180
+      }
+    }
+  end
+end
```
ethn: very handsome tests!

card/mod/admin/set/self/admin.rb
xithan: I wrote a decko-cop for events and views but at some point it stopped working.

card/mod/carrierwave/lib/carrier_wave/cardmount.rb
xithan: true

card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
xithan: it's only one line to change that (maybe plus some old migrations that use `image:`)

card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
xithan: as far as I know ruby doesn't support something like that.
You can do 
```
case model
when ->(m) { m.cloud? }
```
or override the === operator for symbols.

PR #652, opened 2016-08-17
more documentation

ethn:
Card::Cache,
Card::Change...

### DIFF
--- a/card/lib/card/auth/setup.rb
+++ b/card/lib/card/auth/setup.rb
@@ -8,10 +8,11 @@ class Card
       # app is not totally set up yet
       # @return [true/false]
       def needs_setup?
-        @simulating_setup_need || (
-          !Card.cache.read(SETUP_COMPLETED_KEY) &&
-          !Card.cache.write(SETUP_COMPLETED_KEY, account_count > 2)
-        )
+        @simulating_setup_need || begin
+          !Card.cache.fetch(SETUP_COMPLETED_KEY) do
+            account_count > 2
+          end
+        end
         # every deck starts with two accounts: WagnBot and Anonymous
       end
 
--- a/card/lib/card/auth/setup.rb
+++ b/card/lib/card/auth/setup.rb
@@ -8,11 +8,8 @@ class Card
       # app is not totally set up yet
       # @return [true/false]
       def needs_setup?
-        @simulating_setup_need || begin
-          !Card.cache.fetch(SETUP_COMPLETED_KEY) do
-            account_count > 2
-          end
-        end
+        @simulating_setup_need ||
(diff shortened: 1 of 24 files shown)


card/mod/history/lib/card/change.rb
```diff
@@ -1,9 +1,29 @@
...
+  # * the new _value_ of that field
+  # * the {Action action} of which the change is part
+  #
   class Change < ActiveRecord::Base
-    belongs_to :action, foreign_key: :card_action_id, inverse_of: :card_changes
+    belongs_to :action, foreign_key: :card_action_id,
+                        inverse_of: :card_changes
+
+    # lists the database fields for which changes are recorded
+    TRACKED_FIELDS = %w(name type_id db_content trash).freeze
```
xithan: enum?

card/lib/card/auth/setup.rb
```diff
@@ -8,10 +8,11 @@ module Setup
...
       def needs_setup?
-        @simulating_setup_need || (
-          !Card.cache.read(SETUP_COMPLETED_KEY) &&
-          !Card.cache.write(SETUP_COMPLETED_KEY, account_count > 2)
-        )
+        @simulating_setup_need || begin
+          !Card.cache.fetch(SETUP_COMPLETED_KEY) do
+            account_count > 2
+          end
+        end
```
xithan: looks much better with `fetch` but I'm not a fan of the additional block. What about
```ruby
@simulating_setup_need ||
  !Card.cache.fetch(SETUP_COMPLETED_KEY) { account > 2 }
```

card/lib/card/cache.rb
```diff
@@ -1,6 +1,22 @@
 # -*- encoding : utf-8 -*-
 
 class Card
+  def self.cache
+    Card::Cache[Card]
+  end
```
xithan: makes more sense here than in "fetch"

xithan:
nice how the documentation is growing 👍

card/mod/history/lib/card/change.rb
ethn: maybe.  there are things like this that I don't know how to replace:
```
 Card::Change::TRACKED_FIELDS.each_with_object
```
...because I don't know how to get access to the original array.  but maybe I could do `enum :field, TRACKED_FIELDS`  or something?

ethn: `You tried to define an enum named "field" on the model "Card::Change", but this will generate a class method "name", which is already defined by Active Record. (ArgumentError)`

card/lib/card/auth/setup.rb
ethn: much better

PR #654, opened 2016-08-22
change format of format objects from XyzFormat to Format::XyzFormat

ethn:
this one's for Matt.

### DIFF
--- a/card/.yardopts
+++ b/card/.yardopts
@@ -1,4 +1,6 @@
 **/lib/**/*.rb                    \
+**/mod/*/format/*.rb              \
+**/mod/*/chunk/*.rb               \
 tmpsets/set*/**/*.rb -            \
 **/*.rdoc                         \
 GPL                               \
--- a/card/lib/card/format.rb
+++ b/card/lib/card/format.rb
@@ -1,6 +1,23 @@
 # -*- encoding : utf-8 -*-
 
 class Card
+  # The {Format} class is a key strut in the MoFoS (Model-Format-Set)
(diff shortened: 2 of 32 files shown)


xithan:
What is the argument for using Format::XyzFormat and not Format::Xyz?

ethn:
I think I could go for `Format::Xyz`, but let me talk it out.

There could be cases where this causes namespace confusion, but not one that we couldn't address, at least in our own code.  I have an "almost" example: `Format::CSV`

If you have this in a module:
````
format :csv do
  CSV.generate_line ...
````
...then the CSV would be interpreted as referring to the set format class instead of the root CSV.  The reason I say "almost" is because with our current pattern it would probably be `Format::Csv` not `Format::CSV`, so the issue wouldn't actually occur.

Obviously you could fix that with `::CSV` if it actually were an issue.

There might be other issues that arise if there is a name that's shared by a type and format (eg HTML).  

I don't know.  Should I try it and see how it goes?

xithan:
What is the naming pattern for set specific formats? Used to be `Type::Pointer::HtmlFormat`. Do we want to keep that or change it to `Type::Pointer::Format::Html` or  `Type::Pointer::Format::HtmlFormat`?

ethn:
In this pull request, it's still `Type::Pointer::HtmlFormat`.  That actually looks cleaner to me than the other two options and may be a reasonable argument for keeping the Card::Format::XyzFormat naming convention.

xithan:
Then XyzFormat it is

PR #658, opened 2016-08-30
Heroku

xithan:
- make wagn work with heroku: needed a way to set bucket credentials with env variable
- files storage fixes:
  - public links only for local files
  - cache dir was hard coded to local
  - use absolute paths for assets in css files
- refactor admin tasks


card/mod/admin/set/self/admin.rb
```diff
@@ -9,16 +9,14 @@
...
   when :clear_view_cache     then Card::Cache::ViewCache.reset
   when :delete_old_revisions then Card::Action.delete_old
   when :repair_permissions   then Card.repair_all_permissions
   when :clear_solid_cache    then Card.clear_solid_cache
   when :clear_machine_cache  then Card.reset_all_machines
   else
-    if (task_data = Hash[tasks][task])
-      task_data[:execute_policy].call
-    end
+    task_data = tasks.find { |h| h[:name].to_sym == task }
```
xithan: our ruby-style.yml says that we prefer `find` over `detect`

card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
```diff
@@ -253,16 +260,18 @@ def action_id
...
 
     private
 
     def storage
       case @model.storage_type
-      when :cloud then ::CarrierWave::Storage::Fog.new(self)
+      when :cloud
+        require "carrierwave/storage/fog"
+        require "fog"
+        ::CarrierWave::Storage::Fog.new(self)
```
ethn: cool.  this means I can make it so wikirate no longer requires fog, right?

xithan: that should be true

PR #660, opened 2016-09-01
make Machine and InputMachine abstract sets

xithan:


ethn:
cool. somehow I thought this change wouldn't be quite this smooth.

xithan:
All the machine functionality was added to a set by including the Machine module. Apparently, it doesn't matter whether it's a module or a (abstract) set module.

PR #662, opened 2016-09-05
initial script and style machine output as coded file card

xithan:
also:
- corrected admin tasks route
- fix seed data generation


wagn/lib/wagn/tasks/wagn.rake
```diff
@@ -295,31 +298,16 @@ namespace :wagn do
...
-              .update_column :value, card.attachment.db_content(mod: mod_name)
-          card.expire
-          card = Card.fetch card.name
-
-          target_dir = card.store_dir
-
-          files.each do |version, path|
-            FileUtils.cp path, card.attachment.path(version)
-          end
+          binding.pry
```
ethn: remove?

card/db/migrate_core_cards/20150724123438_update_file_and_image_cards.rb
```diff
@@ -25,8 +25,9 @@ def update_db_content card
     attach_array[0].match(/\.(.+)$/) do |_match|
       extension = Regexp.last_match(1)
       if attach_array.size > 3 # mod file
+        mod_name = attach_array[3].sub(/^0\d_/, "")
         card.update_column :db_content,
-                           ":#{card.codename}/#{attach_array[3]}.#{extension}"
+                           ":#{card.codename}/#{mod_name}.#{extension}"
       else
         card.update_column :db_content,
                            "~#{card.id}/#{card.last_action_id}.#{extension}"
```
ethn: not vital, but wouldn't this be cleaner as:
```
  db_content =
      if attach_array.size > 3 # mod file
        mod_name = attach_array[3].sub /^0\d_/, ""
        ":#{card.codename}/#{mod_name}.#{extension}"
      else
        "~#{card.id}/#{card.last_action_id}.#{extension}"
      end
    card.update_column :db_content, db_content
  end
```
?

card/lib/card/mod/dirs.rb
```diff
@@ -17,12 +17,20 @@ def initialize mod_paths
...
+                "name conflict: mod with name \"#{mod_name}\" already loaded"
+        end
         @mods << mod_name
-        # TODO: do something if two mods have the same name?
         @paths[mod_name] = File.join @current_path, mod_name
       end
 
+      def path mod_name
+        @paths[mod_name]
+      end
```
ethn: not sure about this api with #mod and #path.   maybe it should be add_path and path?

ethn:
Are we using a new image resizing pattern?  I know there are plusses and minuses to both ways of doing it (images too small one way, images stretched the other...), but I'm worried that if we switch now we'll a lot of sites with a weird jumble of too-small and stretched images, and it will make it hard to design for those sites.

maybe it's a config thing?

xithan:
I didn't change the resize pattern. I think the sizes were just wrong for some mod images. Maybe because the resizing didn't work when the seed data was generated or it was on purpose for example for the favicon.

card/db/migrate_core_cards/20150724123438_update_file_and_image_cards.rb
xithan: why stop half way? :wink:
```ruby
basename =
      if attach_array.size > 3 # mod file
        mod_name = attach_array[3].sub /^0\d_/, ""
        ":#{card.codename}/#{mod_name}"
      else
        "~#{card.id}/#{card.last_action_id}"
      end
    card.update_column :db_content, "#{basename}.#{extension}"
```

card/lib/card/mod/dirs.rb
xithan: it's #mod so that the Modfile works similar to the Gemfile: gem 'wagn', mod 'carrierwave'

ethn:
It used to be that if you uploaded a 72x72px image, and "large" meant 500x500px, the "large" version would still only be 72x72; it wouldn't stretch larger.

I get (now) that this change didn't affect the resizing algorithm, but it does look like it got changed at some point, so it may be that some sites are already complicated by images reflecting a mix of the two systems.

ethn:
for example:

**OLD RESIZING**
http://wikirate.org/*favicon.png?size=original
http://wikirate.org/*favicon.png?size=large
(large = original)

**NEW RESIZING:**
original: http://wikirate.org/Oil_Natural_Gas_Corporation_Limited+image.png?size=original
large: http://wikirate.org/Oil_Natural_Gas_Corporation_Limited+image.png?size=large
(large > original)

PR #664, opened 2016-09-06
use raw view in history

xithan:


card/mod/bootstrap/set/all/bootstrap/table.rb
```diff
@@ -40,7 +40,7 @@ def table_cell cell
     if cell.is_a? Hash
       content_tag :td, cell.delete(:content).to_s.html_safe, cell
     else
-      content_tag :td, cell.html_safe
+      content_tag :td, String(cell).html_safe
```
ethn: why does one use this instead of .to_s?

card/mod/history/lib/card/action/differ.rb
```diff
@@ -45,9 +45,20 @@ def diff_object field, opts
...
           diff_args = opts || card.include_set_modules.diff_args
-          diff_object :content, diff_args
+          previous = raw_view previous_value(:content), diff_args[:format]
+          current = raw_view value(:content), diff_args[:format]
+
+          Card::Content::Diff.new previous, current, diff_args
         end
       end
+
+      def raw_view content, format=:text
```
ethn: nice solution. I expected we'd be setting the action_id rather than the content directly, but perhaps this makes more sense here.

card/mod/bootstrap/set/all/bootstrap/table.rb
xithan: in general `Type()` is stronger than `to_type`. With `Type()` you are saying give me a `Type` if there is any sensible way of doing so (eg. `"a".to_a` fails but `Array("a")` returns `["a"]`) . 
But I double-checked for the String case and learned that `to_s` is so widely spread that there is no difference (in fact String() calls to_s on it's argument)

xithan: more precisely: `String()` calls first `to_str` and then `to_s`

xithan: and there is a special case where `Type()` is more picky then `to_type`:
```ruby
Integer("no_number")  # => fail
"no_number".to_i      # => 0
```

card/mod/history/lib/card/action/differ.rb
xithan: was my first thought, too.  Could be that this simple solution doesn't work with file and image cards.

card/mod/bootstrap/set/all/bootstrap/table.rb
ethn: cool. thanks for explaining.

PR #665, opened 2016-09-06
fix relative root

xithan:
wagn's routes used to override deck routes

### DIFF
--- a/wagn/lib/wagn/application.rb
+++ b/wagn/lib/wagn/application.rb
@@ -76,7 +76,10 @@ module Wagn
         paths["app/models"] = []
         paths["app/mailers"] = []
 
-        add_path paths, "config/routes.rb", with: "rails/application-routes.rb"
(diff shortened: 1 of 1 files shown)


wagn/lib/wagn/application.rb
```diff
@@ -76,7 +76,10 @@ def paths
         paths["app/models"] = []
         paths["app/mailers"] = []
 
-        add_path paths, "config/routes.rb", with: "rails/application-routes.rb"
+        unless paths["config/routes.rb"].existent.present?
+          add_path paths, "config/routes.rb",
+                   with: "rails/application-routes.rb"
+        end
```
ethn: I don't think this will solve the problem (although it's good to have).

The routes are supposed to be relative to the Rails root; if you change the rails root, nothing should have to change in the routes file.

I'll put more detail in the PT story.

PR #666, opened 2016-09-07
simplify i18n

xithan:
In sets and formats you can just call `t(:key)`. The right scope is added automatically.

includes #662 because it uses the mod path api from that PR


ethn:
no semaphore, huh?

Not sure I understand the general pattern, but this one looks like the issue might be merge conflicts.

xithan:
merged into #670

PR #667, opened 2016-09-07
fix wrong carrierwave tmp path

chuenlok:
To override the `tmp_path` in carrierwave
```
    def tmp_path
      @tmp_path ||= File.expand_path(File.join('..', 'tmp'), root)
    end
```
from `wagn/card/vendor/carrierwave/lib/carrierwave.rb`

### DIFF
--- a/card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
+++ b/card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
@@ -1,4 +1,9 @@
 module CarrierWave
+  class << self
+    def tmp_path
+      @tmp_path ||= File.expand_path("tmp", root)
+    end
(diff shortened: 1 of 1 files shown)


card/mod/carrierwave/lib/carrier_wave/file_card_uploader.rb
```diff
@@ -1,4 +1,9 @@
 module CarrierWave
+  class << self
+    def tmp_path
+      @tmp_path ||= File.expand_path("tmp", root)
```
ethn: Should probably be 
```
@tmp_path ||= Card.paths["tmp"].existent.first`
```
That way if someone changes the tmp directory configuration, it will take effect everywhere.

PR #671, opened 2016-09-08
trash card

xithan:
lists all deleted cards and offers the possibility to put them back
_trash_-_my_wagn

### DIFF
--- a/card/mod/admin/set/self/trash.rb
+++ b/card/mod/admin/set/self/trash.rb
@@ -1,0 +1,29 @@
+format :html do
+  def trashed_cards
+    Card.where(trash: true)
+  end
+
+  view :core do |args|
+    rows =
+      trashed_cards.map do |card|
+        [card.name, Card[card.updater_id].name, put_back_link(card)]
+      end
+    output [
+      (content_tag(:p, empty_trash_link) if rows.present?),
+      table(rows, header: ['card','deleted by',''])
+    ]
+  end
+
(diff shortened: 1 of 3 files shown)


card/mod/admin/set/self/trash.rb
```diff
@@ -0,0 +1,29 @@
...
+  def empty_trash_link
+    card_link :admin, path_opts: { action: :update, task: :empty_trash,
+                                   success: { id: "~#{card.id}" } },
+                      text: "empty trash"
+  end
+
+  def put_back_link card
+    before_delete = card.actions[-2]
+    link_path = path action: :update, view: :open, action_ids: [before_delete],
+                     success: { id: card.name }
```
xithan: is it more likely that you want to rescue another card (= stay here) or do something with the card you just put back ( = redirect to that card)?

ethn: Not sure, but if you were trying to restore a bunch of cards at once, it would be annoying if this redirected.

So let's stay here and provide a link to the new card

ethn:
I commend your process for choosing the term "put back", but after chewing on it a little, I think "restore" might be clearer.  That's because in filesystems, most users think of files as having a "place" (directories) to put the file back in.  But I don't think people think of cards that way.  It's more like changing the card itself rather than moving it around.  (but maybe that's my developer bias speaking?)

PR #683, opened 2016-09-13
solid cache tweaks

xithan:
also: 
- don't use simplecov if rspec runs for a single file to speed up testing
- include_set can pass options to the included set
- reduce duplication in spec helpers for core and deck testing


wagn/lib/wagn/simplecov_helper.rb
```diff
@@ -1,60 +0,0 @@
-def card_simplecov_filters
```
xithan: use card's simplecov_helper instead

card/spec/support/card_spec_loader.rb
```diff
@@ -0,0 +1,84 @@
+class CardSpecLoader
+  class << self
```
xithan: shared code of wagn/lib/wagn/mods_spec_helper.rb and card/spec/spec_helper.rb

card/mod/solid_cache/set/abstract/solid_cache.rb
```diff
@@ -10,31 +10,36 @@
 
 card_accessor :solid_cache, type: :html
 
-format :html do
-  def default_core_args args
-    args[:solid_cache] = true unless args.key?(:solid_cache)
+def self.included host_class
+  host_class.format(host_class.try(:cached_format) || :base) do
```
ethn: cool solution

card/mod/solid_cache/set/abstract/solid_cache.rb
```diff
@@ -72,21 +77,36 @@ def event_name set, args
...
 
+def generate_content_for_cache changed_card=nil
+  format_type = try(:cached_format) || :base
+  format(format_type)._render_core(solid_cache: false,
+                                   changed_card: changed_card)
+end
+
+def updated_content_for_cache _changed_card=nil
+  generate_content_for_cache
+end
```
ethn: what's the rationale for separating these?  overrides in WikiRate?

card/spec/lib/card/view_cache_spec.rb
```diff
@@ -1,3 +1,5 @@
+# -*- encoding : utf-8 -*-
```
ethn: no objection to these, but they only matter if there are non-ascii characters in the code, right?

card/spec/support/card_spec_loader.rb
ethn: I like it.

card/spec/lib/card/view_cache_spec.rb
xithan: I get a bunch of "non-ascii character- default to utf-8" warnings each time I run the tests. Hoped that helped but it didn't.

card/mod/solid_cache/set/abstract/solid_cache.rb
xithan: yes, surgical updates using `changed_card` (in WikiRate) instead of regenerating everything

PR #684, opened 2016-09-14
migration to fix mod files

xithan:


card/db/migrate_core_cards/20160914132636_fix_mod_files.rb
```diff
@@ -0,0 +1,14 @@
+# -*- encoding : utf-8 -*-
+
+class FixModFiles < Card::Migration::Core
+  def up
+    Card.search(type: "image").each do |card|
+      next unless card.coded?
+      next unless card.content.include?("05_standard") ||
```
ethn: should we migrate Card::Change content, too?  Otherwise diffs/reversions will all be broken, no?

xithan: right

PR #685, opened 2016-09-14
Act.all_viewable is much faster and avoids :ghost: acts

ethn:
Many :recent pages were coming up blank (or at least many cards short), because of ghost acts.  And on wikirate the act_viewable queries were sometimes taking many seconds.

This fixes that query

### DIFF
--- a/card/mod/history/lib/card/act.rb
+++ b/card/mod/history/lib/card/act.rb
@@ -53,16 +53,17 @@ class Card
       # all actions that current user has permission to view
       # @return [Array of Actions]
       def all_viewable
-        joins = "JOIN card_actions ON card_acts.id = card_act_id " \
-                "JOIN cards ON cards.id = card_actions.card_id"
-        where = [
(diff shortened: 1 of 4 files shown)


card/mod/history/lib/card/act.rb
```diff
@@ -53,16 +53,17 @@ def find_all_with_actions_on card_ids, args={}
...
+          "card_actions.card_id is not null",
           "draft is not true",
-          Card::Query::SqlStatement.new.permission_conditions("cards")
+          Query::SqlStatement.new.permission_conditions("cards")
         ].compact.join " AND "
 
-        joins(joins).where(where).uniq
+        viewable_actions =
+          Action.joins(card_join).where(action_conditions).to_sql
+        where("card_id is not null AND EXISTS (#{viewable_actions})")
```
ethn: this was the secret -- a `where exists` instead of a `select distinct`.  we're heavily relying on `select distinct` in card queries, so this could be an opportunity for further optimization there...

PR #686, opened 2016-09-16
loader flags "no set module" and "simple load"

xithan:
"no set module" skips the last level in the set module hierarchy
"simple load" skips it completely


card/lib/card/set/loader.rb
```diff
@@ -52,15 +52,36 @@ def pattern_and_modules_from_path path
       end
 
       def tmp_file_template pattern, modules, content_path
+        content = File.read(content_path)
+        wrapped_content = tmp_file_wrapped_content content_path, content
+
+        # TODO: load directly without tmp file
+        return wrapped_content if content =~ /\A#!\s?simple load/
```
ethn: I like the distinction between `simple load` and `not set module`.  But is simple load really loaded right away, or is a tmp file created?  Should we just load it?

xithan: No, it creates a tmp file. I didn't want to spend a day on this feature that maybe will be never used. The question is: What should "load right away" mean? If we just load it here that would mean ignore mod and file order and load "simple load" files across all mods before everything else. Load it in order with tmp files seems like an involved task to me. Maybe only respect mod order is an acceptable compromise.

PR #687, opened 2016-09-17
thepath

ethn:
refactoring #path


card/mod/standard/set/all/links.rb
```diff
@@ -80,79 +71,126 @@ def view_link text, view, opts={}
...
-      opts.delete(:action)
-      base = "new/#{opts.delete(:type)}"
-    else
-      name = opts.delete(:name) || card.name
-      base = opts[:action] ? "card/#{opts.delete :action}/" : ""
+  # @param opts [Hash]
+  # @option opts [Symbol] :action card action (:create, :update, :delete)
+  # @option opts [Integer, String] :id
+  # @option opts [String, Card::Name] :name
+  # @option opts [String] :type
```
xithan: no symbol or type_id?

card/mod/standard/set/all/links.rb
```diff
@@ -80,79 +71,126 @@ def view_link text, view, opts={}
...
   end
 
-  def process_path_card_opts opts, name, linkname
-    opts[:card] ||= {}
-    if opts.delete(:known) == false && name.present? && name.to_s != linkname
-      opts[:card][:name] = name
+  def path_query opts
+    finalize_card_opts opts.delete(:card), opts
+    opts.delete :action
+    opts.empty? ? "" : "?#{opts.to_param}"
```
xithan: so you didn't address the original problem that you can't have `name` as a param?

ethn: no, not yet, but I don't really like the idea of `opts: { params: {} }`

I think `opts` should basically be `params` and anything that doesn't fit that should be separate.  In fact, I might ultimately get it to `def path params, opts`.

path should map very cleanly to routes.  :action, :card, and :id are pretty normal params but with weird roles in routes.  I want to switch :id to :mark in both routes.rb and #path, and it may help solve the name problem if I can get :name merged into :mark. 

:type is the other weird one.  will discuss above in your other comment.

card/mod/standard/set/all/links.rb
ethn: I can see altering the behavior here, but it's not immediately clear what we want.  Here's what currently happens:
1. if action is :new (the only not-a-real-action action supported), we use it to produce /new/Cardtype
2. otherwise if opts[:type] is a known card, we move it to params[:card][:type]

If we wanted to support symbols or Integers, would we want to turn them into cardname based paths or would we want to use our stringy versions and support things like new/:type_search?

xithan: What about this: We get rid of type and use `action: :new, mark: `cardtype` and if cardtype is not a cardtype we produce `/new/Basic?card[name]=cardtype`

card/mod/standard/set/all/links.rb
xithan: `mark` sounds like a good idea

PR #688, opened 2016-09-19
improve save helper

xithan:


card/lib/card/model/save_helper.rb
```diff
@@ -1,48 +1,111 @@
 class Card
   module Model
+    # Helper methods to create and update cards.
+    # The general pattern is:
+    # All methods use the ActiveRecord !-methods that throw exceptions if
+    # somethings fails.
+    # All !-methods in this module rename existing cards
+    # to resolve name conflicts)
```
ethn: Looks good!

At some point we may mention here that this API is intended as a helper for "external" scripts (seeding, testing, migrating, etc) and not for internal application code.

xithan: added your comment

PR #689, opened 2016-09-22
new api for #link_to and related methods

ethn:
it was too hard to keep all the different patterns straight.  short version:
```
link_to text, opts
link_to_card cardish, text, opts
link_to_view view, text, opts
link_to_related related_cardish, text, opts
link_to_resource resource, text, opts
smart_link_to text, opts
# all support :path opt
```

submitting PR for api feedback at this point (not all tests are passing yet).


xithan:
Is it possible to make `link_to` as smart as `smart_link_to` or is that a bad idea?

card/mod/standard/set/all/links.rb
```diff
@@ -1,81 +1,93 @@
-format do
-  # link is called by web_link, card_link, and view_link
-  # (and is overridden in other formats)
-  def link_to text, pathish, _opts={}
-    href = interpret_pathish pathish
+RESOURCE_TYPE_REGEXP = /^([a-zA-Z][\-+\.a-zA-Z\d]*):/
 
+format :html do
+  def link_to text=nil, opts={}
```
xithan: documentation for these methods would be great

card/mod/standard/set/all/links.rb
```diff
@@ -1,81 +1,93 @@
...
+    when %r{^/}               then "internal-link"
+    when /^mailto\:/          then "email-link"
+    when RESOURCE_TYPE_REGEXP then Regexp.last_match(1) + "-link"
+    end
+  end
+
+  def link_to_card cardish, text=nil, opts={}
+    opts[:path] ||= {}
+    name = opts[:path][:name] = Card::Name.cardish cardish
+    # @fixme - need smarter mark handling
```
xithan: is @fixme a yard thing? Rubocop expects `# FIXME: `

card/mod/standard/set/all/links.rb
```diff
@@ -1,81 +1,93 @@
...
+  def resource_type resource
+    case resource
+    when /^https?\:/          then "external-link"
+    when %r{^/}               then "internal-link"
+    when /^mailto\:/          then "email-link"
+    when RESOURCE_TYPE_REGEXP then Regexp.last_match(1) + "-link"
+    end
+  end
+
+  def link_to_card cardish, text=nil, opts={}
```
xithan: `link_to_card cardish, text_or_opts={}, opts=nil` to avoid the nil for card links with the name as text?

xithan: setting text to nil isn't very intuitive to get the card name as text

card/mod/standard/set/all/links.rb
xithan: what is the reasoning that we use a `path` key instead of the rails pattern with an `options` and a `html_options` hash?

ethn: yes, will definitely add documentation.

I played with html_options (although I called it format_opts, because we're using the same API in all formats).  It was much harder to unify the API that way.

For example link_to_resource (formerly web_link) takes, in essence, a fully formed path as its first argument. 

The current, obvious pattern is `link_to[_target] [target,] text, opts`.  If we switched that to `link_to[_target], [target,] text, path_opts, format_opts, then we'd either have to have a mostly nonsensical 3rd argument (path_opts) for link_to_resource, or we'd have to break the pattern.

Also, we have a common pattern of passing these arguments around quite a lot, (see button_link, toolbar.rb, etc)  If it's one hash, it ends up being a lot easier and cleaner to pass it around.

card/mod/standard/set/all/links.rb
ethn: I thought so. We should pick a convention.  Shall we go with FIXME for now?

card/mod/standard/set/all/links.rb
ethn: I think that would work for me.

ethn: ...but if we do that, it seems like we should do that for every method where text is optional.

xithan: agree

card/mod/standard/set/all/links.rb
xithan: then FIXME it is

card/mod/standard/set/all/links.rb
xithan: I buy the first argument, not so sure about the second. Two hashes seems cleaner to me because the path key has a completely different meaning compared to all other keys in the hash.

PR #699, opened 2016-10-07
recent changes, bootstrap helper and view caching

xithan:
status:
**view caching**
stubs are disabled and view caching is turned of for views on search cards
no tests

**bootstrap helper**
blocks inside methods called by bootstrap helper methods don't work
(had to remove haml blocks in recent changes because of that)

**recent changes**
new layout for acts in recent changes and history is implemented
new layout for actions NOT implemented
"rollback" and "hide changes" links are not in the mockup. decision needed what to do with them
accordion helpers need overhaul

_recent_-_my_wagn


card/lib/card/format/render.rb
```diff
@@ -3,27 +3,38 @@ class Format
...
+        return unless (view = renderable_view(view, args))
+        args = default_render_args view, args
+        cache_render view, args do
+          render! view, args
+        end
+      rescue => e
+        rescue_view e, view
+      end
+
+      def render! view, args
```
xithan: if we find another name for this we could use `render!` instead of `_render` for skipping permissions

card/mod/history/set/all/act_view.rb
```diff
@@ -0,0 +1,49 @@
...
+    args[:hide_diff]   ||= hide_diff?
+    args[:slot_class]  ||= "revision-#{act.id} history-slot list-group-item"
+    args[:action_view] ||= action_view
+    args[:actions]     ||= action_list args
+    act_context args
+  end
+
+
+  view :act do |args|
+    ActRenderer.new(self, args[:act], args).render
```
xithan: this is used to render recent changes and history. `args[:act_context]` switches between those two
`absolute` = recent changes
`relative` = history

PR #700, opened 2016-10-11
fixed broken edit rules links and single-use tokens

ethn:
trying to fix some urgent wikirate issues


ethn:
have to merge this rather quickly. sorrz.

PR #702, opened 2016-11-02
polish rollback, fix broken cuke

ethn:
So, the solution is lame ("wait a sec"), but at least I kinda-sort understand the problem and why the solution works.

Short version: it's a timing issue that involves cache clearing.  The reason we were seeing "chick" instead of "chicken" was that the cache hadn't expired yet. 

Basically, there's a race between the the "save as current" action and the fresh loading of the "First" page triggered by the last statement. It's possible that the caching shifted the timing dynamics because it's loading the layout faster.  (I thought at first there might be an issue with the cache clearing, but I can't explain the randomness.)

### DIFF
--- a/card/mod/history/lib/card/act/act_renderer.rb
+++ b/card/mod/history/lib/card/act/act_renderer.rb
@@ -12,11 +12,7 @@ class Card
       end
 
       def method_missing method_name, *args, &block
-        if block_given?
-          @format.send(method_name, *args, &block)
-        else
-          @format.send(method_name, *args)
-        end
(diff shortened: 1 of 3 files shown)


xithan:
If that's the reason why does
```
    When In the main card body I click "Save as current"
    Then I should see "chicken"
```
fail or even
```
    When In the main card body I click "Save as current"
    And I wait until ajax response done
    Then I should see "chicken"
```
?
There is no race. It just waits for the respond of "Save as current" and get's "chick" instead of "chicken"

PR #703, opened 2016-11-02
fix recaptcha

xithan:


card/mod/standard/set/all/rich_html/editing.rb
```diff
@@ -187,7 +187,7 @@ def rename_referer_effect_and_option
     [effect, option]
   end
 
-  view :edit_type, perms: :update do
+  view :edit_type, cache: :never, perms: :update do
```
ethn: did you run into a situation where this was needed?  hard for me to picture...

ethn: (same for the others below)

xithan: as I mentioned in slack editing the structure doesn't work anymore because of caching issues.
The spirit of these changes was: we don't win much by caching these views but loose a lot by caching them wrong

card/mod/core/set/all/type.rb
```diff
@@ -18,6 +18,8 @@ def type_name
   type_card.try :name
 end
 
+alias_method :type, :type_name
```
xithan: translating json to card attributes failed because that method doesn't exist.
I feel if we support `card = Card.create type: "basic"` then there should be a `card.type`.
If not we have to fix `ensure_card` 

There is also the general pattern that `something_name` returns a name object and`something` a string that we break here

card/mod/standard/set/all/rich_html/editing.rb
xithan: anyway, what's the point of caching forms?

card/mod/core/set/all/type.rb
ethn: I would be in support of `type` returning a String and `type_name` returning a Card::Name.  That seems to clarify the issue for me.

ethn: (thanks!)

card/mod/standard/set/all/rich_html/editing.rb
ethn: ok, fair enough.  for forms we can say the burden of proof is on showing the need to cache.

PR #704, opened 2016-11-03
key update

xithan:
waiting for the smartname gem update ...

### DIFF
--- a/card/db/migrate_core_cards/20161103154836_update_keys.rb
+++ b/card/db/migrate_core_cards/20161103154836_update_keys.rb
@@ -23,7 +23,20 @@ class UpdateKeys < Card::Migration::Core
       # card wasn't reachable anyway
       # (due to rails inflection update or smartname update)
     else
-      puts "key conflict: can't change #{key} to #{new_key}"
+      # example:
+      # "Matthias Taxes" can be in the database with two keys:
+      # "matthia_taxis" and "matthia_tax".
+      # Their keys will both be updated to "matthias_tax".
+      # "matthia_taxis" is a walking dead since the last rails inflection
+      # update (it's in the db but you can only get to via its id; all requests via
+      # name find the other one). If we update the walking dead
+      # "matthia_taxis" first to "matthias_tax" we want to replace it with
+      # the living "matthia_tax".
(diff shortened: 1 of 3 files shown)


card/db/migrate_core_cards/20161103154836_update_keys.rb
```diff
@@ -0,0 +1,11 @@
+# -*- encoding : utf-8 -*-
+
+class UpdateKeys < Card::Migration::Core
+  def up
+    Card.pluck(:id, :name, :key).each do |id, name, key|
+      new_key = name.to_key
+      next if new_key == key
+      Card.where(id: id).update_all(key: new_key)
```
ethn: Technically, we should probably add something to update referee_keys in the card_references table.  The impact would be quite small, because the id is far more for existing cards, but it would be a good measure for data quality.

card/db/migrate_core_cards/20161103154836_update_keys.rb
```diff
@@ -0,0 +1,11 @@
+# -*- encoding : utf-8 -*-
+
+class UpdateKeys < Card::Migration::Core
+  def up
+    Card.pluck(:id, :name, :key).each do |id, name, key|
```
ethn: Might be worth testing this on wikirate.  If this is fast enough without batches on wikirate, it should be fast enough anywhere.

card/spec/lib/card/subcards_spec.rb
```diff
@@ -21,6 +21,13 @@
       expect(Card["card with subs+sub1"].content).to eq "this is sub1"
     end
 
+    it "check unstable key bug" do
+      Card::Auth.as_bot do
+        Card.create! name: "Matthias", subcards: {"+name" => "test"}
+        expect(Card.exists? "Matthias+name").to be_truthy
+      end
+    end
```
ethn: nice.

card/db/migrate_core_cards/20161103154836_update_keys.rb
xithan: no problem on wikirate

xithan: matter of seconds

PR #710, opened 2016-11-13
fix toolbar hide bug

xithan:


card/mod/standard/set/all/rich_html/editing.rb
```diff
@@ -1,7 +1,7 @@
 format :html do
   ###---( TOP_LEVEL (used by menu) NEW / EDIT VIEWS )
   view :edit, perms: :update, tags: :unknown_ok do
-    voo.show! :toolbar, :help
+    voo.show :toolbar, :help
```
ethn: this all looks good to me.

As a rule, I don't know if there are many cases where we want to be using `show!`, but we had some functionality before to allow developers to override wagneers settings.

Perhaps we should review all the cases of `show!` currently in the code and see if they're appropriate?

PR #712, opened 2016-11-16
use rubocop-decko gem

xithan:


.ruby-style.yml
```diff
@@ -70,6 +70,13 @@ Style/FormatString:
     - format
     - sprintf
     - percent
+Metrics/BlockLength:
+  Enabled: true
+  Exclude:
+    - 'Rakefile'
+    - '**/*.rake'
+    - 'spec/**/*.rb'
+    - '**/*_spec.rb'
```
xithan: I think that should improve our codeclimate score

.rubocop.yml
```diff
@@ -1,7 +1,18 @@
...
+Decko/ViewLength:
+  Description: Avoid views longer than 15 lines of code.
+  Enabled: true
+  CountComments: false
+  Max: 15
+Decko/EventLength:
+  Description: Avoid events longer than 10 lines of code.
+  Enabled: true
+  CountComments: false
+  Max: 10
```
ethn: So can we just add one for format here that sets the max at 100?

PR #732, opened 2016-12-14
support images in emails

xithan:
nests of image cards in `+*html message` fields of email templates are translated to inline attachments of the email
currently the nest has to use core view to make it work


card/mod/core/set/all/collection.rb
```diff
@@ -171,7 +171,7 @@ def implicit_item_view
   end
 
   def voo_items_view
-    return unless (items = voo.items)
+    return unless (voo && items = voo.items)
```
xithan: to fix #contextual_content. It processes content but doesn't have a voo

card/mod/email/set/type/email_template.rb
```diff
@@ -78,14 +74,36 @@ def email_config args={}
...
-    text_message = args.delete(:text_message)
-    html_message = args.delete(:html_message)
-    attachment_list = args.delete(:attach)
-    alternative = (text_message.present? && html_message.present?)
-    mail = Card::Mailer.new_mail(args) do
+    config = card.email_config(args)
+    text_message = config.delete(:text_message)
+    attachment_list = config.delete(:attach)
+    email_card = card # card is not accessible in the new_mail block
+    mail = Card::Mailer.new_mail(config) do
```
xithan: I don't know what kind of black magic I discovered here. I can't access card in this block (because it is called with instance_exec) but email_card is available.

PR #736, opened 2016-12-19
raise error when missing codename in fetch

ethn:
see https://www.pivotaltracker.com/story/show/136107181

### DIFF
--- a/card/mod/core/set/all/fetch.rb
+++ b/card/mod/core/set/all/fetch.rb
@@ -83,11 +83,11 @@ module ClassMethods
     end
   end
 
-  def fetch_name mark, &block
+  def fetch_name mark
     if (card = quick_fetch(mark))
       card.name
     elsif block_given?
-      block.call
+      yield
     end
   end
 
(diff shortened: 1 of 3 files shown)


card/mod/core/set/all/fetch.rb
```diff
@@ -246,21 +246,33 @@ def normalize_fetch_args args
...
+    when /^\:(\w+)$/  # codename, eg ":options"
+      require_id_for_codename(Regexp.last_match[1].to_sym)
+    else
+      fullname_from_mark mark, opts[:new]
+    end
+  end
+
+  def require_id_for_codename mark
+    id = Card::Codename[mark]
+    raise Card::NotFound, "missing card with codename: #{mark}" unless id
```
xithan: is `Card::NotFound` right? in CardController#load_card we raise `Card::Error::NotFound`

card/mod/standard/set/abstract/wql_search.rb
```diff
@@ -29,7 +29,7 @@ def fetch_query args={}
 end
 
 def standardized_query_args args
-  args.symbolize_keys!
+  args = args.symbolize_keys
```
ethn: yes.  in some cases, the args that makes it here is not a normal Hash but a Card::Controller::Parameters (or something) object that does not respond to symbolize_keys! (with the bang).

not sure why it only shows up occasionally, but I've seen it, which is why I originally made the change.

card/mod/core/set/all/fetch.rb
ethn: Card::Error::NotFound is right, will fix.

PR #737, opened 2016-12-21
maintain name context and view after edit

xithan:
nests always switched to open view after editing ignoring the view they had before

### DIFF
--- a/card/mod/standard/set/all/rich_html/menu.rb
+++ b/card/mod/standard/set/all/rich_html/menu.rb
@@ -22,7 +22,8 @@ format :html do
   end
 
   def menu_path_opts
-    opts = { slot: { home_view: voo.home_view } }
+    opts = { slot: { home_view: (voo.home_view || @slot_view),
+                     name_context: context_names_to_params } }
     opts[:is_main] = true if main?
     opts
   end
@@ -67,7 +68,7 @@ format :html do
   end
 
   def menu_edit_link opts
-    menu_item "edit", "edit", opts.merge(view: :edit)
+    menu_item "edit", "edit", opts.merge(view: :edit, path: menu_path_opts)
   end
 
   def menu_discuss_link opts
--- a/card/lib/card/format/names.rb
+++ b/card/lib/card/format/names.rb
@@ -22,6 +22,11 @@ class Card
         context_name_list.split(",").map(&:to_name)
       end
 
+      def context_names_to_params
+        return unless @context_names
+        @context_names.join(",")
+      end
+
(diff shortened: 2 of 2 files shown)


card/mod/standard/set/all/rich_html/menu.rb
```diff
@@ -22,7 +22,8 @@
   end
 
   def menu_path_opts
-    opts = { slot: { home_view: voo.home_view } }
+    opts = { slot: { home_view: (voo.home_view || @slot_view),
```
xithan: I always forget what home view means. Is it the view used for the main card? Maybe we want always @slot_view here.

ethn: home view is specific to a given slot (nothing to do with main).

if, for example, you click "cancel" on a given form on a card in a slot, then you go back to that slot's home view.  if it was originally `content`, then its home view is `content`, and you go back to `content` when you click cancel.

ethn: this solution is ok, I guess, but the question for me is how we're getting here without a home view.  My first guess is that is one of the things that gets taken care of when we do `nest cardish, view: viewname` but NOT when we do `subformat(cardish).render_viewname`

Also, I'm a little surprised this isn't taken care of by the slotter handling...

ethn: do the links in question have the "slotter" class?

ethn: it looks to me like they don't.  they should; that would take care of most of this.  perhaps #menu_item should add the slotter class by default.  I'm guessing the page link is the only place we don't want it.

Then you could probably get rid of #menu_path_opts altogether.

xithan: the menu links are a bit special. The gear has its own slot (because the menu is loaded via ajax) which means the menu links update a different slot. I'm not sure if the slotter class can handle that.

xithan: so the difference between `@slot_view` and `home_view` is that if you edit something `@slot_view` becomes `edit` but `home_view` stays unchanged?

ethn: OK, then I probably won't touch this for now.

I'm thinking in the future we should write some custom (non-slot) javascript for the menu so that links within the main menu can remain simple slotter links and retain all the power that entails.  

but I'll merge this now and move on.

ethn: yes!

PR #741, opened 2016-12-22
better errors

xithan:
debugger was disabled

### DIFF
--- a/card/lib/card/format/error.rb
+++ b/card/lib/card/format/error.rb
@@ -8,9 +8,7 @@ class Card
         rendering_error e, view
       end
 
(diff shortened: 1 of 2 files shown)


card/lib/card/format/error.rb
```diff
@@ -8,9 +8,7 @@ def rescue_view e, view
         rendering_error e, view
       end
 
-      def debug_error e, view
-        Rails.logger.info "#{rendering_error e, view}:\n" \
-                          "#{e.class} : #{e.message}"
```
xithan: logging here creates a loop

PR #744, opened 2016-12-22
separate method for paging view

xithan:


card/mod/standard/set/abstract/search/paging.rb
```diff
@@ -58,6 +58,10 @@ def paging_path_args local_args={}
     }.merge(extra_paging_path_args).merge local_args
   end
 
+  def paging_view
+    (voo && voo.home_view) || :content
+  end
```
ethn: not called in gem?

PR #750, opened 2017-01-04
"image in email" fix

xithan:


card/lib/card/view/options.rb
```diff
@@ -98,6 +97,14 @@ def slot_options
         normalized_options.select { |k, _v| Options.all_keys.include? k }
       end
 
+      def closest_live_option key
+        if live_options.key? key
+          live_options[key]
+        elsif format.parent && format.parent.voo
+          format.parent.voo.closest_live_option key
+        end
```
ethn: Not sure this is exactly what we want.

So, we have Format ancestry, and we also have View ancestry.  Views (and voos) can have children and grandchildren all within one Format.  (You typically get a new format when you want a view of a new card).

With that in mind:
A. This version of "closest" skips the view ancestry and goes straight to the format ancestry, so it would seem to potentially miss a lot of "closer" options along the way
B. Do we really want our mechanisms for traversing the format ancestry to be inside the View class?

card/mod/carrierwave/set/type/image.rb
```diff
@@ -88,8 +88,17 @@ def new_image action
...
-  view :core do |args|
-    image_tag args[:inline_attachment_url].call(card.attachment.path)
+  view :source do
+    handle_source do |source|
+      image_tag card_url(source)
+    end
+  end
+
+  view :core do
+    url_generator = voo.closest_live_option(:inline_attachment_url)
```
ethn: ...continuing on the preview comment in View::Options, are we really in the wrong Format or just the wrong voo?

xithan: Both can happen. It depends on the view used for the nested image. Because of `contextual_content` the format is usually wrong

PR #753, opened 2017-01-05
fix edit structure

xithan:
support card object as argument

### DIFF
--- a/card/mod/core/set/all/collection.rb
+++ b/card/mod/core/set/all/collection.rb
@@ -228,10 +228,11 @@ format do
   end
 
   def normalized_edit_fields
-    edit_fields.map do |name, options|
-      options ||= Card.quick_fetch(name).name
+    edit_fields.map do |name_or_card, options|
+      next [name_or_card, options || {}] if name_or_card.is_a?(Card)
+      options ||= Card.fetch_name name_or_card
       options = { title: options } if options.is_a?(String)
-      [card.cardname.field(name), options]
+      [card.cardname.field(name_or_card), options]
     end
   end
 
--- a/card/lib/card/subcards.rb
+++ b/card/lib/card/subcards.rb
@@ -251,7 +251,7 @@ class Card
          (absolute_name.parts.size - @context_card.cardname.parts.size) > 2
         left_card = new_by_attributes absolute_name.left
         new_by_card left_card, transact_in_stage: attributes[:transact_in_stage]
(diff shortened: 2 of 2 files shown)


card/mod/core/set/all/collection.rb
```diff
@@ -228,10 +228,11 @@ def edit_fields
   end
 
   def normalized_edit_fields
-    edit_fields.map do |name, options|
-      options ||= Card.quick_fetch(name).name
+    edit_fields.map do |name_or_card, options|
+      next [name_or_card, options || {}] if name_or_card.is_a?(Card)
```
ethn: it doesn't seem like this should ever be a card.  we can't have fancy objects (like cards) as values of our voo options, and edit_structure is a voo option (as currently implemented)

ethn: And, looking in the code, I don't see where edit_structure ever contains cards?  Does this really fix something?

xithan: yes, it fixes metric value editing on wikirate

xithan: ok, there is a way to avoid that. Instead of 
``` ruby
year = card.fetch trait: :year, new: { content: card.year }
voo.edit_structure = [[year, "Year"]]
```
I can use
```ruby
card.add_subfield :year, content: card.year
voo.edit_structure = [[:year,  "Year"]]
```
But conceptually it's the same. I have to pass an card object to edit_structure. I just don't do it directly. I use the cache instead to pass it.

ethn: So, here's what I'm thinking we can say (and implement and document):

There are 3 ways to create a web form on Wagn:

1. passively.  Allow Wagn to autogenerate a form from the read content/structure
2. simple configuration.  via either:
 - voo.edit_fields (what we're currently doing with edit_structure.  has to be simple objects. creates a simple linear form).  Nest syntax version may need to be a simple comma separated list?
 - voo.edit_structure (like voo.structure, gives a card name.  This way you can really customize the layout of your form)
3. advanced (coder) configuration
- override `def edit_fields` (and do something to signal to wagn that we should use it)
- override something else to structure the whole form

Basic ideas:
1. voo stuff needs to stay simple, because every voo value has to be short and easily cached
2. we've got to make the coder API coder friendly.

Thoughts?

xithan: one thought: edit_structure should be a rule

PR #756, opened 2017-01-09
fixes to persistent caching

ethn:


card/lib/card/cache/persistent.rb
```diff
@@ -63,7 +63,7 @@ def annihilate
       # the cache. Note that Cardio.cache is a simple Rails::Cache, not
       # a Card::Cache object.
       def stamp
-        @stamp ||= Cardio.cache.fetch stamp_key { new_stamp }
+        @stamp ||= Cardio.cache.fetch(stamp_key) { new_stamp }
```
ethn: believe it or not, this was not working.  Our entire stamp system was broken because of missing parentheses.  I'm amazed things weren't much worse.

card/lib/card/cache/persistent.rb
```diff
@@ -96,17 +96,13 @@ def read key
       # @param key [String]
       # @param attribute [String, Symbol]
       def write_attribute key, attribute, value
-        if @store && (object = read key)
+        return value unless @store
+        @store.send(:local_cache).clear # clear rails local cache
```
ethn: this should help dramatically with view clearing.  seems to be working much closer to expectations.

I didn't even realize the rails caching had a local (middleware) layer.

PR #757, opened 2017-01-10
nest option for view cache

xithan:


card/lib/card/format/nest/fetch.rb
```diff
@@ -7,6 +7,7 @@ def fetch_nested_card cardish, opts={}
           case cardish
           when Card            then cardish
           when Symbol, Integer then Card.fetch cardish
+          when "_", "_self"    then card.context_card
```
xithan: Is there a good reason that we used to refetch instead of using the same card? That caused problems for rendering trashed cards (which happens for notification emails of deletions) because "{{_self}}" refers then to a new card instead of the trashed card.

ethn: no, I don't think there was any good reason.  This seems like a good patch.

card/spec/lib/card/format/render_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+      "nest_name:A;nest_syntax:A|content;view:contentcontent:show"
+    end
+
+    subject { Card::Cache[Card::View] }
+
+    it "can be changed with nest option" do
+      is_expected.to receive(:fetch).with cache_key
+      render_content "{{A|content}}"
+      is_expected.not_to receive(:fetch)
+      render_content "{{A|cache:never}}"
```
ethn: So my concerns with this approach:
1. I could be wrong, but I don't think there is anything conceptually special about the +companies card; I think it's a general bug, and we'd just be picking which appearances of that bug annoy us most, no?
2. In principle, there are some kinds of caching that I think wagneers may want to think about.  Eg, they might want to be able to do `cache: daily` on a complicated homepage or report.  But right now the caching concept is pretty heavy.
3.  I think it would need to be clearer what this means.  When you look at it, your first guess is that it means "never cache the homeview of this slot", but as implemented, it means "don't cache any view ever rendered within this slot".  So if it starts as content, then goes to history or edit, etc, none of those get cached.

It's a clever fix to an ugly bug, but now that I know the site is performing ok without view caching, I think we should probably keep it turned off until we get view caching working more reliably, at which point this feature won't be needed.

xithan: 1. more precisely, it's a general tool to fix (individually) this kind of bug, but not the general solution.
2. When I started that I thought the +companies nest was in a structure card. I didn't notice that it was overridden with a code structure. I'm fine with making it a code and not a wagneer option
3. I'm not sure what makes more sense. The value of this feature is that you can turn off caching for views you don't have a code rule for. Since there is no way to address subviews with that the question is if it's enough to have only the possibility to disable caching for the home view

ethn: Actually, I think the current code affects all three of these
1. the home view
2. views nested by the home view (it's listed as an "heir" option in options.rb)
3. other views navigated to by the user within the same slot, eg by clicking on edit

As a side note, I'm wondering if we should do something about all the content overridden by structure and structure overridden by code.  it's quite confusing.

PR #766, opened 2017-01-19
"final stages"

xithan:
Add event option to run event after all subcard events of the same stage.
The solution introduces an additional "final_stage" callback for every stage.


ethn:
I'm going to merge this because I don't feel like I can be picky right now, but I'd like to dig into this wording.  "final stage" sounds like the last stage, not the last part of a stage.  stage_completion, maybe?

PR #769, opened 2017-01-25
don't validate children if validation fails for parent

xithan:


card/mod/core/set/all/trash.rb
```diff
@@ -95,7 +95,8 @@ def delete!
   end
 end
 
-event :validate_delete_children, :prepare_to_validate, on: :delete do
+event :validate_delete_children, after: :validate_delete, on: :delete do
+  return if errors.any?
```
ethn: Elegante Chirurgie, Arzt Kühl

PR #770, opened 2017-01-26
first step towards haml view templates

xithan:


wagn/features/step_definitions/web_steps.rb
```diff
@@ -55,28 +55,6 @@
...
-  #                           .click
-  # li = find("li", text: value, visible: false)
-  # li.click
-  # # If the list element is too far down the list then the first click
-  # # scrolls it up but doesn't select it. It needs another click.
-  # # A selected item is no longer visible (because the list disappears)
-  # if li.visible?
-  #   li.click
-  # end
-end
```
xithan: Moved to wikirate. Wagn doesn't use (jquery's?) chosen elements.

card/mod/core/set/all/haml.rb
```diff
@@ -1,7 +1,82 @@
...
+    first_try = ::File.expand_path("../#{view}.haml", file)
+      .gsub(%r{/tmp/set/mod\d+-([^/]+)/}, '/mod/\1/view/')
+    return first_try if ::File.exist?(first_try)
+    basename = ::File.basename(file, ".rb")
+    ::File.expand_path("../#{basename}/#{view}.haml", file)
+      .gsub(%r{/tmp/set/mod\d+-([^/]+)/}, '/mod/\1/view/')
+  end
+
+  def haml_to_html haml, locals, a_binding=nil
     a_binding ||= binding
```
ethn: I should probably know what binding means in the context, but to be honest, I don't....

card/mod/core/set/all/haml.rb
```diff
@@ -1,7 +1,82 @@
...
+  #       %p
+  #         Hi
+  #         = name
+  #     HAML
+  #   # => <p>Hi Joe</p>
+  # @example create a slot in haml code
+  #   - haml_wrap do
+  #     %p
+  #       some haml
+  def render_haml template_or_locals={}, locals_or_binding=nil, a_binding=nil
```
ethn: As much as we overuse "args", I'm wondering if it wouldn't actually be a little clearer in this case.  Something like:

```
def render_haml *args
   return render_haml_template(*args) if args.first.is_a? Symbol
   haml = block_given? ? yield : args.shift
   haml_to_html haml, *args
end
```

card/mod/core/set/all/haml.rb
```diff
@@ -1,7 +1,82 @@
...
+  # This method must be overridden in every set that uses haml templates
+  # so that `__FILE__` points to the right file.
+  # @return [String] path to haml view template
+  # @example
+  #   def view_template_path view
+  #     super(view, __FILE__)
+  #   end
+  def view_template_path view, file=__FILE__
+    first_try = ::File.expand_path("../#{view}.haml", file)
+      .gsub(%r{/tmp/set/mod\d+-([^/]+)/}, '/mod/\1/view/')
```
ethn: perhaps we need a method like `try_view_template_path`?

card/mod/core/set/all/haml.rb
xithan: binding IS the context :smile: 
It defines what variables and methods I can use in haml. If I call binding here then that means I can access in haml everything that I can access here. At some point I had the wrong context in haml. Maybe in specs. I don't see used anywhere at the moment.

card/mod/core/set/all/haml.rb
xithan: the method body is clearer but args doesn't tell you what args are expected

card/mod/core/set/all/haml.rb
xithan: What we really need is a completely different approach. Something like a view to path hash that we generate when we load the set modules. But I was working on research pages ...

PR #771, opened 2017-01-26
introduce native mode for "new" forms

xithan:
If an existing card is used to render a form for a new card with fields then there was no way to attach the field to the new card.
It would create a field on the existing card.

Example: Card "A" renders a form to create a new card with a field "+B". The name for the input tag
has to be "[subcards][+B]" instead of "[subcards][A+B]".
To distinguish those two case we have to distinguish between :new mode and :edit mode. But :new is translated to :edit.
I added a @native_mode attribute to hold on to the original :new mode.


card/mod/standard/set/all/rich_html/form.rb
```diff
@@ -54,7 +54,14 @@ def form_for_multi
...
     return "" if !form_root_format || form_root_format == self
-    "#{@parent.subcard_input_names}[subcards][#{card.contextual_name}]"
+    "#{@parent.subcard_input_names}[subcards][#{name_in_form}]"
+  end
+
+  # If you use subfield cards to render a form for a new card
+  # then the subfield cards should be created on the new card not the existing
+  # card that build the form
+  def name_in_form
+    @native_mode == :new ? card.relative_name : card.contextual_name
```
xithan: that's where I need the @native_mode

ethn:
Can you tell me more about why we want to support using the :new view of "A" to create some other card?  Why isn't the new view of the other card?

xithan:
justifying example 1:  The content of a field is part of the name of the new card. 
The wikirate use case: The form for a new metric answer has a year field. Maybe I could use a new metric answer card with empty name but then I have to juggle with params and args to pass the metric and the company. If I render the form on a metric record card then all the data I need is easily accessible.

justifying example 2: A session card temporarily stores data needed for the form.
The wikirate use case:  The research form needs a list of companies and metrics. This example doesn't really depend on that feature because not the session card renders the form. I assemble the metrics and companies to new record cards which renders the forms. But there might be cases where it makes more sense to use the session card for the form.

card/lib/card/format.rb
```diff
@@ -54,6 +54,7 @@ def initialize card, opts={}
       opts.each { |key, value| instance_variable_set "@#{key}", value }
 
       @mode ||= :normal
+      @native_mode ||= @mode
```
ethn: In general, I'm uncomfortable that such narrow solutions are coming up in such general places.  Here we're introducing @native_mode in the initialization method of, arguably, our second most important class.  (in MoFoS, Card = Mo and Card::Format = Fo.  Maybe it's a tie for second with Card::Set, which = S).

@root and @depth are very natural things to be here, and we could (and I should) explain them with something like: Formats are hierarchical, and any given format can have a parent and/or children.  @root is the Format object at the top of the tree, and the depth refers to the generational depth of this object.

@mode should have a similar explanation.

@native_mode seems so much more complex, no?

card/lib/card/format/nest/subformat.rb
```diff
@@ -5,8 +5,9 @@ module Subformat
         def subformat subcard
           subcard = Card.fetch(subcard, new: {}) if subcard.is_a?(String)
           self.class.new subcard, root: @root, parent: self, depth: @depth + 1,
-                                  form: @form, mode: @mode,
-                                  context_names: @context_names
+                         form: @form,
+                         mode: @mode, native_mode: @native_mode,
+                         context_names: @context_names
```
ethn: More scrooginess from me here, because I was hoping to remove most of these and make them lazy-load from the parent)

card/mod/standard/set/all/rich_html/form.rb
ethn: dumb question: why can't we just always use the relative name?

(somehow I doubt your answer will be "DOH.  never considered that!")

ethn:
OK, a few more higher level comments here:

I'm realizing @mode is already a little hard to define, and that's not helping things.  So let me have a stab here.

A Format's mode (@mode) can be used to override nests views in certain contexts.  There are three built-in modes:

- **normal** mode (default), does not override view
- **edit** mode, used for editing, overrides most nests with edit fields
- **template** mode, used for treating content as an abstract template, overrides nests with views that help wagneers manipulate template

This maps smoothly onto our old REWD framing (readers, editors, wagneers, developers).  normal is for readers, edit is for editors, and template is for wagneers.  Ahh, so clean and clear.

Unfortunately, there's currently also **closed** mode and **layout** mode.

I don't think closed mode should be a mode.  we just need to move that processing to the :closed_content view.  I can't actually find evidence that **layout** mode is in use, but I may be missing something.

Will try to make all this relevant in next comment.

ethn:
Slow to get this, but this is effectively adding a "new" mode (though that's just a native mode).  But the main thing we're trying to get at is whether we're currently within a new view, right?

I think the question of "are we inside X view" is a good one to be able to answer, generally. Can't we traverse the voos to find that out?

card/mod/standard/set/all/rich_html/form.rb
xithan: Yes, I did consider that. I also considered to punch myself in the groin. But after a second of reflection I rejected both ideas.  I don't know why exactly but I have a distant memory that I have done things like these before and their outcome was equally painful.

xithan: More precisely and less sarcastic: I remember that we have touched that before during a frustrating debug session. So my guess is if we change that and we are lucky that our tests will discover the problems caused by that change then it's a few hours of debugging. If we are unlucky we will notice weird stuff in a few months and spend a few days to find and fix it.  Maybe I'm exaggerating. I agree that in principle in an ideal code base the relative name should work but it is sitting at the gate to the hell of unpredictable side effects, the subcards handling :scream:

xithan:
In general, yes, but in the example that motivated that, it's really the mode and not the view (there is no new view involved because of the new form for a different card thing)

xithan:
To get rid of the native_mode we could introduce a voo flag like @name_context = :relative/:absolute. Or maybe `with_context "" do ...`. But there is the issue that `with_context` sets  @context_card but `contextual_name` uses `@contextual_name` and not `@context_card`

ethn:
To go back a couple of comments, the thing that makes something a "mode" is that it changes the interpretation of a nest.  I don't think that happens here, right? 

maybe something like

```
def with_relative_names_in_form
  @relative_names_in_form = true
  yield
  @relative_names_in_form = nil
end

def relative_names_in_form?
  @relative_names_in_form || (parent && parent.relative_names_in_form?)
end
```
?

card/mod/standard/set/all/rich_html/form.rb
ethn: yeah, point taken.

ethn:
It's possible that I'm just being thick and @native_mode is meaningful in a way that I'm not understanding.  I'm interpreting it as a very abstract way to describe something that's really just a narrow flag for narrow functionality, and I'm wary of mixing it in with more central pieces.

xithan:
I'm happy with your narrow solution. I chose the general approach because it was almost already there. I saw that we have a :new mode but forget it immediately and save it as edit. 
So my point was if we support `with_nest_mode :new` anyway, then it should be possible to distinguish it from :edit mode 
Maybe it's just bad naming. When I came across NEST_MODES it wasn't clear to me if the keys or the values are meant to be modes. Because it's only used in `with_nest_mode` and the key variable is called "mode" I considered it as a melting pot of modes not as a view to mode translation which it could also be.

ethn:
Thanks for be willing to go into such length about this.  I know it's obnoxious, but I have felt uncomfortable about our mode handling for a long time, and this makes me feel like we've got a clear path.

I wrote up next steps here: https://www.pivotaltracker.com/story/show/138489591

card/mod/pointer/set/abstract/01_pointer/edit.rb
```diff
@@ -89,7 +89,7 @@ def add_item_button
 
   view :multiselect do |_args|
     select_tag(
-      "pointer_multiselect",
+      "pointer_multiselect-#{unique_id}",
```
ethn: I think this is breaking cukes.  Fwiw, those are awful cuke steps.  If this prompts us to make better high level steps, it will be a double gain!

wagn/features/pointer_inputs.feature
```diff
@@ -21,15 +21,15 @@ Feature: Pointer Inputs
     Given I create Phrase card "User+*type+*structure" with content "{{+friends}}"
     And I create Phrase card "friends+*right+*input" with content "select"
     When I edit "Joe User"
-    And I select "Joe Admin" from "pointer_select"
```
xithan: better?

ethn: much!

wagn/features/pointer_inputs.feature
```diff
@@ -38,7 +38,7 @@ Feature: Pointer Inputs
     Given I create Phrase card "User+*type+*structure" with content "{{+friends}}"
     And I create Phrase card "friends+*right+*input" with content "multiselect"
     When I edit "Joe User"
-    And I select "Joe Admin" from "pointer_multiselect"
+    And I select "Joe Admin" from "friends"
```
ethn: one is +friends.  other is "friends".  difference is...label?

xithan: The correct label is with "+". I forgot the "+" here but it doesn't matter because I remove the "+" in the step definition anyway. I'm not able to find the label with the "+" because it is in a separate span tag.

PR #772, opened 2017-01-30
rake tasks to create script and style cards

xithan:


card/mod/history/set/all/content_history.rb
```diff
@@ -164,5 +164,6 @@ def draft_acts
 def edit_conflict?
   last_action_id_before_edit &&
     last_action_id_before_edit.to_i != last_action_id &&
-    last_action.act.actor_id != Auth.current_id
+    (la = last_action) &&
+    la.act.actor_id != Auth.current_id
 end
```
ethn: so it's possible to have a `last_action_id_before_edit` without having a `last_action`?

fwiw, breaking these multi-test methods up can do wonders for readability.  eg
```
def edit_conflict?
  return unless action_id = potential_conflict_action_id 
  Card::Action[action_id].act.actor_id != Auth.current_id
end
```
that might not solve the problem, but it reads like one or two tests instead of four.

xithan: Maybe a session card gets an action if it is created but not when it is updated? just a guess

PR #775, opened 2017-02-03
multi forms with more than one level

xithan:


card/mod/standard/set/all/rich_html/form.rb
```diff
@@ -1,20 +1,109 @@
...
+
+  view :type_formgroup do
+    wrap_type_formgroup do
+      type_field class: "type-field edit-type-field"
+    end
+  end
+
+  view :edit_in_form, cache: :never, perms: :update, tags: :unknown_ok do
+    @form = form_for_multi
+    multi_edit_slot
```
xithan: the main change: `:edit_in_form` checks `voo.editor`

PR #776, opened 2017-02-08
fix: bug if self is part of nested form

xithan:
make this work
```
  def edit_fields
    [
      [card, { title: "Answer", editor: :standard }],
      [unknown_field_card, { hide: :title }],
      [card.left.fetch(trait: :checked_by, new: {}), { hide: :title }]
    ]
  end
```


ethn:
hypocritical comment: moving forward let's try to reproduce bugs in tests.  (you're better at this than I am, I admit)

PR #782, opened 2017-03-01
Card toolbar

srivig:
Changed the structure of card edit toolbar for responsiveness.
Needs code review.


card/mod/bootstrap/set/all/bootstrap/navbar.rb
```diff
@@ -26,6 +27,14 @@ def navbar id, opts={}
     end
   end
 
+  def navbar_nocollapse id, opts={}, content
+    wrap_with :div, class: "navbar navbar-inverse" do
+      [
+        wrap_with(:div, content, class: "container-fluid")
+      ]
```
xithan: you don't need an array if you have a single element

card/mod/standard/set/all/rich_html/toolbar.rb
```diff
@@ -34,24 +35,29 @@ def default_toolbar_args args
...
+    opts = {}
+    opts[:no_nav] = true
+    close_link "pull-right visible-xs navbar-text", opts
   end
 
   def tool_navbar
     navbar "toolbar-#{card.cardname.safe_key}-#{voo.home_view}",
            toggle_align: :left, class: "slotter toolbar",
            navbar_type: "inverse",
+           no_collapse: true,
```
xithan: General parameters like `:class` and `id` should not be ignored if you set `no_collapse` to true. It's ok that contradictory parameters like `collapsed_content` have no effect but it doesn't make sense to keep them here in the call.

xithan: I also suspect that the buttons don't work properly without the slotter class. Did you check it? If you merge with the latest master you will get proper test results (everything that fails then fails because of your code and not mine :)

xithan:
my PRs were pending when srivig submitted this PR. If he closes this and resubmits it my changes should disappear

srivig:
I just closed that PR and made another PR

PR #785, opened 2017-03-08
bring ?debug=style back to live

xithan:


card/mod/basic_formats/set/all/all_css.rb
```diff
@@ -22,12 +22,14 @@ def show view, args
   end
 
   view :import do |_args|
-    %{\n@import url("#{_render_url items: { view: :import }}");\n}
+    voo.items[:view] = :import
+    %{\n@import url("#{_render_url}");\n}
```
ethn: this change is fine, but I'm surprised that the previous one didn't work with your changes to the url view.

card/mod/machines/set/type/css.rb
```diff
@@ -55,6 +55,12 @@ def default_nest_view
   view :content_changes, mod: Abstract::Script::HtmlFormat
 end
 
+format :css do
+  view :import do
+    %{\n@import url("#{_render_url}");\n}
+  end
+end
```
ethn: I'm surprised that (a) this is needed, and (b) it doesn't have the voo.items handling.  Is there another `import` view somewhere?

card/mod/pointer/set/abstract/01_pointer.rb
```diff
@@ -75,9 +75,14 @@ def default_item_view
   end
 
   view :core do
+    voo.items[:view] = params[:item] if params[:item]
```
ethn: So, the way this is _supposed_ to work is via `main_nest_items_opt`, which processes params[:item].  The intended benefit over a solution like this one is that the item view will only be applied to the main nest and not to any random rendering.  (Eg, if a CSS sheet had lots of nests in it).

But I think the problem is that this is not rendered as part of a nest at all.  That's a problem with our "main_nest" approach, I guess.  Maybe we need to broaden that to "main_content" and have it called in cases like this where the main content is not in a nest?

card/mod/machines/set/type/css.rb
xithan: This is definitely needed and correct. That's were the nesting ends. But the other import views do unnecessary stuff.

PR #786, opened 2017-03-10
pluck in batches

xithan:


card/config/initializers/02_patches/active_record.rb
```diff
@@ -0,0 +1,65 @@
...
+# Arguments
+#   columns      ->  an arbitrary selection of columns found on the table.
+#   batch_size   ->  How many items to pluck at a time
+#   &block       ->  A block that processes an array of returned columns.
+#                    Array is, at most, size batch_size
+#
+# Returns
+#   nothing is returned from the function
+
+module Patches
```
ethn: Patches not Batches?

xithan: I misspelled it "badges" in the branch name but patches is correct. It's a patch for ActiveRecord.

PR #787, opened 2017-03-15
card events

xithan:
Support `notification templates` and `twitter templates` in addition to the existing `email templates`.  They can be added to `*on update`, `*on create` and `*on delete` rules.

Notification template means a configurable message is shown at the top of the slot that is loaded after a successful action. A Twitter template sends a tweet.


card/lib/card/model/save_helper.rb
```diff
@@ -83,6 +83,29 @@ def ensure_card! name_or_args, content_or_args=nil
         end
       end
 
+      def ensure_trait name, codename, args
+        ensure_card name, codename: codename
+        args.each do |k, v|
+          send "ensure_trait_#{k}", name, v
```
GerryG: Might be good to protect this with respond_to? and do something more sensible when the params are wrong. Is this just used in migrations? Maybe throwing an error is fine.

card/config/initializers/02_patches/active_record.rb
```diff
@@ -0,0 +1,66 @@
+# pluck_in_batches:
```
GerryG: Is this actually used by this change?

xithan: no, I needed it in a migration for wikirate and thought it might be useful in wagn, too. Actually,  your comment reminds me that we are using `find_in_batches` somewhere in wagn. `pluck_in_batches` could improve performance there.

card/lib/card/model/save_helper.rb
xithan: That module is only used for migrations and testing but sure, more sensible errors messages never hurt

card/mod/machines/lib/javascript/wagn.js.coffee
```diff
@@ -385,3 +385,7 @@ snakeCase = (str)->
               match[1].toLowerCase()
 
 warn = (stuff) -> console.log stuff if console?
+
+wagn.slotReady (slot) ->
+  slot.find('._disappear').delay(5000).animate(
```
ethn: so `_disappear` is a class?  I may have started the practice of using CSS classes as javascript signifiers, but is that considered good practice?

card/mod/notifications/set/all/observer.rb
```diff
@@ -0,0 +1,22 @@
...
+event :cache_delete_card_events, :store, on: :delete do
+  @card_event_cache = event_cards :on_delete
+end
+
+def execute_card_events args
+  setting = "on_#{args[:on]}".to_sym
+  event_cards(setting).each do |event_card|
+    event_card.deliver(context: self)
+  end
+end
```
ethn: this is lovely.

card/mod/notifications/set/all/show_notifications.rb
```diff
@@ -0,0 +1,7 @@
+format :html do
+  view :flash do
+    flash_notice = Env.success.params[:flash]
```
ethn: given the easy write API, I might expect Env.success.flash (with no args) to return this value.

card/config/initializers/02_patches/active_record.rb
GerryG: Ah, makes sense and good point about potential use. Recently ran into the fact that Rails3 doesn't support multi-column pluck, bummer.

card/lib/card/reference.rb
```diff
@@ -41,10 +41,10 @@ def delete_if_referer_missing
           "LEFT JOIN cards ON card_references.referer_id = cards.id"
         ).where(
           "cards.id IS NULL"
-        ).find_in_batches do |group|
+        ).pluck_in_batches(:id) do |group|
           # used to be .delete_all here, but that was failing on large dbs
           puts "deleting batch of references"
-          where("id in (#{group.map(&:id).join ','})").delete_all
+          where("id in (#{group.join ','})").delete_all
```
GerryG: Cool, would it help to also rename the var to group_ids ?

card/mod/machines/lib/javascript/wagn.js.coffee
xithan: What are the alternatives to CSS classes? @srivig started the practice on wikirate to prefix CSS classes with an underscore if they are javascript signifiers. That seems to me like a reasonable practice.

card/config/initializers/01_core_extensions/array.rb
```diff
@@ -2,7 +2,7 @@ module CoreExtensions
   module Array
     def to_pointer_content
       map do |item|
-        "[[#{item}]]"
+        item =~ /^\[\[.+\]\]$/ ? item : "[[#{item}]]"
       end.join "\n"
     end
```
xithan: depends on what you feed in. Instead of checking the input where I use it I wanted to make it work for any kind of input.

PR #796, opened 2017-04-03
support wql in navbox

xithan:
search expressions that are 
wrapped in curly brackets are interpreted as wql

### DIFF
--- a/card/lib/card/migration.rb
+++ b/card/lib/card/migration.rb
@@ -62,12 +62,10 @@ class Card::Migration < ActiveRecord::Migration
     Card::Cache.reset_all
     Cardio.schema_mode "" do
       Card::Auth.as_bot do
-        ActiveRecord::Base.transaction do
-          begin
-            yield
-          ensure
-            Card::Cache.reset_all
-          end
+        begin
+          yield
+        ensure
+          Card::Cache.reset_all
         end
       end
     end
--- a/card/lib/card/mod/loader.rb
+++ b/card/lib/card/mod/loader.rb
@@ -18,6 +18,7 @@ class Card
     module Loader
       class << self
         def load_mods
+          load_initializers
           load_set_patterns
           load_formats
           load_sets
@@ -87,6 +88,12 @@ class Card
             files << i_card.existing_source_paths
           end
           files.flatten
+        end
+
+        def load_initializers
+          Cardio.config.paths["mod/config/initializers"].existent.sort.each do |initializer|
+            load initializer
+          end
         end
 
         def load_set_patterns
--- a/card/lib/card/mod/loader.rb
+++ b/card/lib/card/mod/loader.rb
@@ -91,7 +91,8 @@ class Card
         end
 
         def load_initializers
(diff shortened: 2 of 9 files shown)


card/lib/card/migration.rb
```diff
@@ -62,12 +62,10 @@ def contentedly
     Card::Cache.reset_all
     Cardio.schema_mode "" do
       Card::Auth.as_bot do
-        ActiveRecord::Base.transaction do
```
ethn: interesting.  don't migrations normally do this?  Was this added in because of the schema change?

card/lib/card/mod/loader.rb
```diff
@@ -89,6 +90,13 @@ def source_files card
           files.flatten
         end
 
+        def load_initializers
+          Cardio.config.paths["mod/config/initializers"].existent
```
ethn: Are we usually using Card.paths for this?

card/lib/cardio.rb
```diff
@@ -93,18 +93,34 @@ def set_paths paths
 
       add_path "mod"
 
+      set_db_paths
+      set_initializer_paths
```
ethn: nice

card/mod/standard/set/abstract/wql_search.rb
```diff
@@ -17,10 +17,13 @@ def wql_hash
   end
 end
 
-def query args={}
+def query_args args={}
```
ethn: I like this structure.  I don't yet love the naming, but we can tweak that.

ethn: (seems weird to override something named "query_args")

card/mod/standard/set/self/search.rb
```diff
@@ -1,5 +1,23 @@
 include_set Abstract::SearchParams
 
+def query_args args={}
+  query_args =
+    if wql_keyword? args
+      parse_wql_keyword args
+    else
+      wql_hash.merge args
```
ethn: should self/search ever really run without a keyword?

xithan: `wql_keyword?` doesn't check if the keyword is present it checks if the keyword is a wql expression

card/mod/standard/set/abstract/wql_search.rb
xithan: We did that for views all the time  (`default_my_view_args`)

card/lib/card/migration.rb
xithan: yes, they do and no, you added it in 2013 when there was no schema change:
https://www.evernote.com/l/ABLb922VeVlAcIh0-QhF6MPci2VUaQ53O8I

xithan: no need for a screenshot. I just figured out that github is smart enough to recognize commit SHAs in comments 5e22747

card/lib/card/mod/loader.rb
xithan: Looks like in this file we are using Card.paths. In other places we are using Cardio.paths. What's the difference?

PR #797, opened 2017-04-05
haml views

xithan:
support haml templates in a Rails like way:
If the view option `template: :haml` is set then wagn expects a haml template in a corresponding template path and renders it

Example: 
```ruby
# mod/core/set/type/basic.rb
view :my_view, template: :haml  # renders mod/core/view/type/basic/my_view.haml 

view :with_instance_variables, template: :haml do 
    @hello = "haml"
end   

# mod/core/type/basic/template/with_instance_variables.haml
Hello 
 = hello

> render :with_instance_variables  # => "Hello haml"
```


card/lib/card/query/join.rb
```diff
@@ -8,13 +8,16 @@ class Join
...
+      # documentation??? -pk
       def initialize opts={}
         from_and_to opts
         opts.each do |key, value|
           send "#{key}=", value if value.present?
         end
         @from_field ||= :id
         @to_field   ||= :id
+
+        # why set to empty? why not allow to initialize with conditions? -pk
```
ethn: I doubt this was explicitly done this way to prevent initializing with subjoins; I don't think it came up.  If using `||=` doesn't break any tests, I would expect it's safe.

card/lib/card/auth/current.rb
```diff
@@ -49,6 +49,21 @@ def current= mark
           end
       end
 
+      def serialize
+        { as_id: as_id, current_id: current_id }
+      end
+
+      def deserialize data
```
ethn: do we have any kind of naming pattern to suggest a method expects a block?   eg `deserializing`?

card/lib/card/query/sorting.rb
```diff
@@ -22,30 +22,34 @@ def sort val
...
-          )
-        else
+        method_name = "sort_by_count_#{item}"
+        unless respond_to?(method_name)
           raise Card::Error::BadQuery, "count with item: #{item} not yet implemented"
         end
+        send method_name, val
+      end
+
+      def sort_by_count_referred_to val
```
ethn: way to isolate the cancer!

card/lib/card/set/event.rb
```diff
@@ -1,21 +1,27 @@
...
     # in a background process then Card::Env.deserialize! decouples the
     # controller's params hash and the Card::Env's params hash with the
     # effect that params changes in the CardController get lost
     # (a crucial example are success params that are processed in
     # CardController#update_params_for_success)
-    return if Wagn.config.active_job.queue_adapter == :inline
-    Card::Env.deserialize! env
-    Card::Auth.current_id = current_id
+    return yield if Wagn.config.active_job.queue_adapter == :inline
+    Card::Auth.deserialize auth do
```
ethn: I like the name of the method defined here and am tempted to suggest just `Card::Auth.with(auth)`

card/lib/card/set/format/haml_views.rb
```diff
@@ -0,0 +1,75 @@
...
+
+        def haml_template_render_block_with_locals
+          proc do |view_args|
+            instance_exec view_args, &block
+            locals = instance_variables.each_with_object({}) do |var, h|
+              h[var.to_s.tr("@", "").to_sym] = instance_variable_get var
+            end
+            voo = View.new(self, view, view_args, @voo)
+            with_voo voo do
+              haml_to_html template, locals
```
ethn: does this work?  where is this template arg coming from?

card/mod/carrierwave/set/abstract/attachment/upload_cache.rb
```diff
@@ -48,6 +48,12 @@ def assign_attachment file, original_filename
   return unless (action = Card::Action.fetch(@action_id_of_cached_upload))
   upload_cache_card.delete_files_for_action action
   action.delete
+end
+
+# at some point uploaded files of canceled file card creation
+# should be deleted. We do this when ever an new file is created.
+event :clear_draft_files, :integrate_with_delay,
+      on: :create do
```
ethn: any chance of race conditions here?  Should we test that the tmp files are more than a minute old, say?

card/mod/core/set/all/assign_attributes.rb
```diff
@@ -35,6 +35,9 @@ def extract_subcard_args! args
 def prepare_assignment_params args
   params = ActionController::Parameters.new(args)
   params.permit!
+  if params[:db_content].is_a? Array
+    params[:db_content] = params[:db_content].join("\n")
+  end
```
ethn: so we allow that?

card/mod/standard/set/self/search.rb
```diff
@@ -1,20 +1,15 @@
...
 end
 
-def parse_wql_keyword args
-  args.merge parse_json_query(args[:vars][:keyword])
+def parse_keyword_wql args
+  parse_json_query(args[:vars][:keyword])
 end
 
-def wql_keyword? hash
+def keyword_contains_wql? hash
```
ethn: better.

card/mod/carrierwave/set/abstract/attachment/upload_cache.rb
xithan: we only delete files that are older than 5 days

card/mod/core/set/all/assign_attributes.rb
xithan: It was already allowed because Rails automatically casts everything to the right type. But now it will be casted to something sensible (`[1, 2]` to `"1\n2"` instead of `"[1, 2]"`). Should have happened a long time ago in my opinion. Initializing pointer cards was always a mess.

card/lib/card/set/format/haml_views.rb
xithan: missing in the method definition

PR #802, opened 2017-04-28
async views

xithan:
- view option `async: true` to load view asynchronously
- tweak autocomplete to find more suggestions

### DIFF
--- a/card/lib/card/query.rb
+++ b/card/lib/card/query.rb
@@ -61,7 +61,8 @@ class Card
                            link_to linked_to_by
                            include included_by                                ),
       conjunction:     %w(and or all any),
-      special:         %w(found_by not sort match complete extension_type),
+      special:         %w(found_by not sort match name_match complete junction_complete
+                          extension_type),
       ignore:          %w(prepend append view params vars size)
     }.each_with_object({}) do |pair, h|
       pair[1].each { |v| h[v.to_sym] = pair[0] }
--- a/card/lib/card/set/format.rb
+++ b/card/lib/card/set/format.rb
@@ -81,14 +81,35 @@ class Card
         include Set::Format::HamlViews
 
         mattr_accessor :views
-        self.views = {}
+        self.views = Hash.new { |h, k| h[k] = {} }
 
         def view view, *args, &block
           view = view.to_viewname.key.to_sym
-          views[self] ||= {}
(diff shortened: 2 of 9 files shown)


card/lib/card/query.rb
```diff
@@ -61,7 +61,8 @@ class Query
                            link_to linked_to_by
                            include included_by                                ),
       conjunction:     %w(and or all any),
-      special:         %w(found_by not sort match complete extension_type),
+      special:         %w(found_by not sort match name_match complete junction_complete
```
ethn: re `name_match`, my memory was that you could do:
```
"match":"X" # meaning name or content
"name":["match", "X"] # meaning name only, or
"content":["match", "X"] # meaning content only.
```
Does that not work?  Or is this different?

card/lib/card/set/format.rb
```diff
@@ -81,14 +81,35 @@ module AbstractFormat
...
           interpret_view_opts view, args[0] if block_given?
-          view_block = views[self][view] = view_block view, args, &block
-          define_method "_view_#{view}", view_block
+          view_method_block = view_block(view, args, &block)
+          if async_view? args
+            # This case makes only sense for HtmlFormat
+            # but I don't see an easy way to override class methods for a specific
+            # format. All formats are extended with this general module. So
+            # a HtmlFormat.view method would be overridden by AbstractFormat.view
+            # We need something like AbstractHtmlFormat for that.
```
ethn: yes, seems valuable.  Good one for a huddle.

ethn: (current solution is fine for now)

card/lib/card/query.rb
xithan: `match` does a lot of weird stuff that I don't understand but the thing that I understand in that method is that it matches only complete words

xithan: I don't see any logic in `Query::Attributes#match` that restricts the match to name or content. It looks to me like it does always both.

xithan: What I looked at was the match attribute method not the match operator. 
Probably the `to_sql` method implements the behaviour you expected and that is confirmed by the tests

wagn: yeah, that would have been a lot easier to find if it were in its own method, eh?

PR #803, opened 2017-05-09
help block below the label instead of below the input element. 

srivig:


card/mod/standard/set/all/rich_html/formgroup.rb
```diff
@@ -10,7 +10,7 @@ def formgroup_body title, opts, &block
     label = formgroup_label opts[:editor], title
     editor_body = editor_wrap opts[:editor], &block
     help_text = formgroup_help_text opts[:help]
-    "#{label}<div>#{editor_body} #{help_text}</div>"
+    "#{label}<div>#{help_text} #{editor_body}</div>"
```
wagn: did you test to make sure this looks ok on normal wagns?

PR #806, opened 2017-05-28
read only buckets

xithan:
A bucket can be defined to be "read only". If a file card of a read only bucket is updated the file will be moved to the default storage location.
(was needed for wikirate: After a db refresh from live for the staging site the machine output updates failed because the staging site doesn't have write permission for the live bucket)


card/mod/carrierwave/set/abstract/attachment/storage_type.rb
```diff
@@ -42,6 +47,11 @@
   @new_storage_type ||= storage_type_from_config
 end
 
+event :change_bucket_if_read_only, :initialize, on: :update, when: :cloud? do
```
ethn: when possible, I like to do things like:
```
  event :change_bucket, when: read_only_cloud?...
```
that way, you can have another event that is `after: :change_bucket`, and neither happens unless the condition is met.

card/mod/carrierwave/spec/set/type/file_spec.rb
```diff
@@ -383,7 +383,7 @@ def source_view card
       after { Cardio.config.file_storage = :local }
       it "copies file to local file system" do
         # not yet supported
-        expect { subject.update_attributes!(storage_type: :local) }
+        expect { Card[subject.name].update_attributes!(storage_type: :local) }
```
xithan: The `subject` call creates the card and the `@file_changed` attrbute doesn't get reset. That's why I need a new object.

ethn: ah, got it.

PR #808, opened 2017-06-06
fixed signup bug

ethn:
my first PR in forever.  And there's a test!


card/mod/account/spec/set/type/signup_spec.rb
```diff
@@ -205,6 +205,10 @@
       expect(@account.token).to be_present
       expect(@account.password).not_to be_present
     end
+
+    it "considers signups created by signed-in users to be invitations" do
+      @signup.format.invitation? == true
```
xithan: This test doesn't test anything. `expect` is missing

xithan (changes requested):

card/mod/account/spec/set/type/signup_spec.rb
ethn: True, though it isn't totally worthless; it still failed with an error before my change ;)

Will fix.

xithan (approved):

PR #809, opened 2017-06-08
More layout options

xithan:
added `thin-sidebar` and `fluid` layout options. 
`fluid` removes the left and right margin (using bootstraps `container-fluid` class).
`thin-sidebar` uses 2 instead of 3 columns of bootstrap's 12 columns grid layout for the sidebar.

I added a "Full Width Layout" layout card as example with the following content:
```html
<!DOCTYPE HTML>
<html>
  <head>
    {{*head|core}}
  </head>
  <body class="left-sidebar thin-sidebar fluid">
    <header>{{*header|core}}</header>
    <article>{{_main|open}}</article>
    <aside>{{*sidebar|content}}</aside>
    <footer>{{*footer|core}}</footer>
  </body>
</html>
```


card/mod/machines/lib/javascript/wagn.js.coffee
```diff
@@ -48,22 +48,27 @@ $.extend wagn,
...
+
+    # if slot-selector doesn't apply to a child, search in all parent slots and finally in the body
+    while target_slot.length == 0 and parent_slot.length > 0
+      target_slot = $(parent_slot).find(selector)
+      parent_slot = $(parent_slot).parent().closest '.card-slot'
+    if target_slot.length == 0
+      $(selector)
+    else
+      target_slot
+
```
ethn: much better.

xithan:
Maybe `class` is the wrong attribute. I think there is no css rule involved. It's all javascript. So maybe something like `data-layout="left-sidebar fluid"`. Another option is to be a little bit more specfic e.g. `fluid-layout` or `full-width-layout`

ethn:
I like the data-layout option.  In general I would like to disentangle CSS and Javascript selectors wherever possible.

PR #810, opened 2017-06-12
input setting for all cards

xithan:
So far there is a *input setting only for pointer cards which let you choose between options like `radio`, `checkbox`, `list` etc. This PR enables the *input setting for every card. New input options are 
   - prosemirror editor (default)
   - tinymce editor
   - ace editor 
   - text area
   - text field
   - calendar


card/lib/card/set_pattern.rb
```diff
@@ -124,12 +125,26 @@ def format_module_list klass
...
             part_id && Card::Codename[part_id.to_i] || (return nil)
           end
         end
 
+        def anchor_parts
+          return [@anchor_name] unless anchor_parts_count > 1
+          parts = @anchor_name.parts
+          if parts.size <= anchor_parts_count
+            parts
+          else
```
ethn: what is the situation that makes this necessary?

ethn: (should probably comment)

card/mod/standard/set/all/rich_html/editor.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def default_editor
+    :rich_text
+  end
+
+  # overriden by mods that provide rich text editors
+  def rich_text_input
+    prosemirror_editor_input
+  end
```
ethn: nice.

xithan: The set_pattern change was necessary because self sets didn't work for junction cards. The issue was the following: 
Self sets have only one anchor according to their set pattern definition, so when the sets are loaded a set in `/self/input/right/options` is considered to be a self set for the card `input` and the set module is stored under the key `Self::Input`. But the key that is generated for `*input+*right+*options` to lookup the set modules is always `Self::Input::Right::Options` even if `*input+*right+*options` has its own codename.
I changed it so that if a junction card has a codename then that codename is used instead of the composite of the codenames of its parts.

xithan:
The set_pattern change was necessary because self sets didn't work for junction cards. The issue was the following: 
Self sets have only one anchor according to their set pattern definition, so when the sets are loaded a set in `/self/input/right/options` is considered to be a self set for the card `input` and the set module is stored under the key `Self::Input`. But the key that is generated for `*input+*right+*options` to lookup the set modules is always `Self::Input::Right::Options` even if `*input+*right+*options` has its own codename.
I changed it so that if a junction card has a codename then that codename is used instead of the composite of the codenames of its parts.

card/lib/card/set_pattern.rb
xithan: For example the example that I described in the other comment.
`A+B+*self` needs to have one anchor part `A+B` and not two parts `[A, B]` in order to get the codename of `A+B` and be able to define code rules for it.

ethn: Got it. Can we add this comment:
"Handles cases where anchor is a compound card, eg `A+B+*self`"

PR #811, opened 2017-06-21
fix paging bugs

xithan:


card/mod/notifications/set/self/follow_defaults.rb
```diff
@@ -31,7 +31,8 @@
 
 format :html do
   view :edit, perms: :update, tags: :unknown_ok do
-    frame_and_form :update do
+    frame_and_form :update, hidden: { success: "_self",
+                                      card: { update_all_users: false } } do
```
ethn: lord, that's much more readable.

Though I don't actually remember what `update_all_users` means...

xithan: It means change the follow rule for all existing users in contrast to only apply it as default rule to new users who will sign up in the future.

PR #812, opened 2017-06-23
fix html escape bug in rules editor

xithan:


ethn:
But thanks for fixing!!

PR #813, opened 2017-06-27
fix Card::Auth.with

xithan:
it has to reset @as_id if no as_id argument is given

### DIFF
--- a/card/lib/card/auth/current.rb
+++ b/card/lib/card/auth/current.rb
@@ -64,14 +64,18 @@ class Card
           auth_data = { current_id: Card.fetch_id(auth_data) }
         end
 
-        tmp_current = current_id
+        tmp_current_id = current_id
         tmp_as_id = as_id
-        @current_id = auth_data[:current_id]
+        tmp_current = @current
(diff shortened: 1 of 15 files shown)


card/lib/card/auth/current.rb
```diff
@@ -64,14 +64,18 @@ def with auth_data
           auth_data = { current_id: Card.fetch_id(auth_data) }
         end
 
-        tmp_current = current_id
+        tmp_current_id = current_id
         tmp_as_id = as_id
-        @current_id = auth_data[:current_id]
+        tmp_current = @current
+        tmp_as_card = @as_card
+        self.current_id = auth_data[:current_id]
```
ethn: any reason why we're mixing `self.current =` and `@as_id =` instead of picking one convention or the other?

ethn: ah, I see. maybe add this comment: `resets @as and @as_card`

ethn (approved):

ethn:
Is there anything here that needs to be merged (or at least considered for merger) into decko?

